PR #358, opened 2014-05-28
increased allowable channel and topic name length

svmehta:


yujinqiu:
@mreiferson 
We use NSQ in production environment and find 64 is not long enough for us,  128 will be good.   
Can we  make  TopicName and ChannelName  length configurable  with default length 64 ?

mreiferson:
@yujinqiu we applied a limit here because after a certain length they get unwieldy to display, etc. We _think_ it promotes good naming.

However, there's no _technical_ reason this needs to be 64, so if you are persuasive in your arguments, I suppose we would consider increasing it.

yujinqiu:
@mreiferson  thanks for you reply. In our use case we monitor different domains  and  domain's path,  to separate  domains we use domain/path as topic name.  sometimes  path will reach 64 bytes limit,  so we want it to be configurable and use 64 bytes as default for downside  compatibility.  please consider this feature, or I have to change this value and recompile it.

PR #323, opened 2014-09-04
nsqadmin: refactoring

mreiferson:
`nsqadmin` is a Go web app, which has some tradeoffs. It's great because it's dead simple to deploy. It's terrible because it's a PITA writing Go based web apps.

I think an ideal middle ground is:

1. `nsqadmin` becomes a JSON API with convenience endpoints to query the cluster and aggregate data
2. `nsqadmin` UI is a client-side app
3. all the static assets for (2) are bundled into the binary so it's still easy to deploy (this is already the case, just stating the obvious)

Simultaneously, it could use a facelift, but this refactoring alone makes things much easier to extend and for others to contribute.

cc @visionmedia


mreiferson:
@jehiah dare I say this is RFR

FYI, one noteworthy item is that it's using hash-based routing so that the urls don't conflict with the JSON API endpoints.

@rolyatmax will you take a look at how all this backbone non-sense came out.  Looking forward to the react re-write next! (should be a lot easier now)

mreiferson:
(rebased on #610)
(fixes #478)

mreiferson:
@jehiah I'm hoping you can give this a test run w/ graphite integration, please?  I don't currently have a handy installation :grin:

nsqadmin/static/js/lib/handlebars_helpers.js
```diff
@@ -0,0 +1,247 @@
...
+        'deferred_count':         'gauge', //eslint-disable-line key-spacing
+        'requeue_count':          'counter', //eslint-disable-line key-spacing
+        'timeout_count':          'counter', //eslint-disable-line key-spacing
+        'message_count':          'counter', //eslint-disable-line key-spacing
+        'clients':                'gauge', //eslint-disable-line key-spacing
+        '*_bytes':                'gauge', //eslint-disable-line key-spacing
+        'gc_pause_*':             'gauge', //eslint-disable-line key-spacing
+        'gc_runs':                'counter', //eslint-disable-line key-spacing
+        'heap_objects':           'gauge', //eslint-disable-line key-spacing
+        'e2e_processing_latency': 'gauge' //eslint-disable-line key-spacing
```
rolyatmax: If it seems easier or cleaner, you could wrap these lines with:
```
/* eslint-disable key-spacing */
...
/* eslint-enable key-spacing */
```

http://eslint.org/docs/user-guide/configuring.html#configuring-rules

nsqadmin/static/js/models/node.js
```diff
@@ -0,0 +1,25 @@
...
+
+var Node = Backbone.Model.extend({
+    idAttribute: 'name',
+
+    constructor: function Node() {
+        Backbone.Model.prototype.constructor.apply(this, arguments);
+    },
+
+    url: function() {
+        return AppState.url('/nodes/' + this.get('name'));
```
rolyatmax: I think you can just define
```
urlRoot: function() { 
    return AppState.url('/nodes');
}
```
to make this work because you have `idAttribute: 'name'` already set.

http://backbonejs.org/#Model-urlRoot
and the code:
http://backbonejs.org/docs/backbone.html#section-89

But maybe that's too much magic/abstraction?

nsqadmin/static/js/models/node.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+
+    tombstoneTopic: function(topic) {
+        return $.ajax(this.url(), {
+            'data': JSON.stringify({'topic': topic}),
+            'method': 'DELETE'
+        });
```
rolyatmax: I think this might also work?
```
this.destroy({
    data: {'topic': topic}
});
``` 

But is also less clear as to what it's actually doing (and might result in having this `node` removed from its collections?

nsqadmin/static/js/models/topic.js
```diff
@@ -0,0 +1,16 @@
...
+var Topic = Backbone.Model.extend({
+    idAttribute: 'name',
+
+    constructor: function Topic() {
+        Backbone.Model.prototype.constructor.apply(this, arguments);
+    },
+
+    url: function() {
+        return AppState.url('/topics/' + this.get('name'));
+    }
```
rolyatmax: See `urlRoot` comment above.

nsqadmin/static/js/lib/handlebars_helpers.js
mreiferson: hah

nsqadmin/static/js/models/node.js
mreiferson: I like it

jehiah:
@mreiferson I'll try this out and should check back in by end of week.

nsqadmin/static/js/views/channel.js
```diff
@@ -0,0 +1,47 @@
...
+    template: require('./spinner.hbs'),
+
+    events: {
+        'click .channel-actions button': 'channelAction'
+    },
+
+    initialize: function() {
+        BaseView.prototype.initialize.apply(this, arguments);
+        this.listenTo(AppState, 'change:graph_interval', this.render);
+        this.model.fetch().done(function() {
```
rolyatmax: Another way to handle this is to listen to `this.model`'s `sync` event and use `render` as the callback.

You also might be able to call `render` when this view is instantiated - placing a loader in the view if the model has no data yet. Then when the fetch returns, the render callback mentioned above will fire - filling the view with the model's newly fetched data.

rolyatmax: From offline convo: might not be worth making this change, as the payoff is mostly in terms of future-proofing the code. But it sounds like we want to do a React rewrite anyway. So let's not worry about this right now.

nsqadmin/static/js/views/counter.js
```diff
@@ -0,0 +1,109 @@
...
+    },
+
+    startLoop: function(i) {
+        this.interval = i;
+        this.poller = setInterval(this.updateStats.bind(this), i);
+    },
+
+    updateStats: function() {
+        $.get('/counter').done(function(data) {
+            var num = _.reduce(data, function(n, v) {
```
rolyatmax: Love me some `reduce`

nsqadmin/static/js/views/counter.js
```diff
@@ -0,0 +1,109 @@
...
+        this.startLoop(100);
+    },
+
+    startLoop: function(i) {
+        this.interval = i;
+        this.poller = setInterval(this.updateStats.bind(this), i);
+    },
+
+    updateStats: function() {
+        $.get('/counter').done(function(data) {
```
rolyatmax: This code should prob live in a model, but again, this is more "future-proofing". If we're gonna scrap and move to React, it might not be a big deal to just leave this as is.

nsqadmin/static/js/views/counter.js
```diff
@@ -0,0 +1,109 @@
...
+        this.targetPollInterval = 30000;
+        this.currentNum = -1;
+        this.tries = 0;
+        this.updateStats();
+        this.startLoop(100);
+    },
+
+    startLoop: function(i) {
+        this.interval = i;
+        this.poller = setInterval(this.updateStats.bind(this), i);
```
rolyatmax: It probably is worth changing this to a `setTimeout`. And then having the callback to the `$.get('/counter')` set another `setTimeout`.

rolyatmax:
My main comments have to do with the ajax/model related code that is finding its way into the views. (Probably why your model files are so thin.) The `counter.js` view is a little hard to follow - seems to be managing too much state in an effort to make the counter animation work. If the thing is functional, it might be fine to just land this, and save refactoring changes for a rewrite PR.

There are a few misc comments (see the `setInterval` comment) which I think could actually cause problems, however, and should probably be fixed in this work.

nsqadmin/static/js/views/counter.js
mreiferson: OK, and related it should probably clear all pending timeouts when this view is removed

jehiah:
Some bugs so far:

* On a channel page, some of the action buttons are misslabeled for topics

screen shot 2015-07-29 at 1 58 16 pm

* On the same page in the client connections list, the `NSQd Host` column is incomplete

* When toggling from the counter page back to other pages, the Counter view is not removed.

mreiferson:
@jehiah give this another spin plz

mreiferson:
@jehiah check the last 4 commits and give it a spin.  Switched to normal URLs, other minor UI fixes, and tested against graphite in a VM.

nsqadmin/static/js/views/channel.js
```diff
@@ -0,0 +1,59 @@
...
+            '/' + this.model.get('name') + '</em>?';
+        bootbox.confirm(txt, function(result) {
+            if (result !== true) {
+                return;
+            }
+            if (action === 'delete') {
+                $.ajax(this.model.url(), {
+                    'method': 'DELETE'
+                }).done(function() {
+                    window.location = '/';
```
jehiah: Since this is a channel deletion it would be nice to go back to `/topic/%s`

nsqadmin/static/js/views/lookup.js
```diff
@@ -0,0 +1,77 @@
...
+        e.stopPropagation();
+        var topic = $(e.target.form.elements['topic']).val();
+        var channel = $(e.target.form.elements['channel']).val();
+        if (topic === '' && channel === '') {
+            return;
+        }
+        $.post(AppState.url('/topics'), JSON.stringify({
+            'topic': topic,
+            'channel': channel
+        })).done(function() {
```
jehiah: It would be great to surface errors from these calls; currently something like an invalid topic/channel name is just dropped here w/o user feedback.

jehiah:
@mreiferson unsure of why yet, but i get this stack trace from the topic endpoint `/api/topics/%s`

```
[nsqadmin] 2015/08/03 18:35:56.399820 200 GET /api/topics/[edited] (127.0.0.1:54645) 101.731ms
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x9dd479)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/runtime/panic.go:491 +0xad
runtime.newstack()
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/runtime/stack.c:784 +0x555
runtime.morestack()
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/runtime/asm_amd64.s:324 +0x7e

goroutine 9 [stack growth]:
reflect.Value.Field(0x7c9800, 0xc2082fa000, 0xd9, 0x0, 0x0, 0x0, 0x0)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/reflect/value.go:708 fp=0xc2288e0208 sp=0xc2288e0200
encoding/json.fieldByIndex(0x7c9800, 0xc2082fa000, 0xd9, 0xc2082a90e0, 0x1, 0x1, 0x0, 0x0, 0x0)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/encoding/json/encode.go:760 +0x199 fp=0xc2288e0280 sp=0xc2288e0208
encoding/json.(*structEncoder).encode(0xc2086d5470, 0xc2080a8630, 0x7c9800, 0xc2082fa000, 0xd9, 0x0)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/encoding/json/encode.go:573 +0x11d fp=0xc2288e0428 sp=0xc2288e0280
encoding/json.*structEncoder.(encoding/json.encode)·fm(0xc2080a8630, 0x7c9800, 0xc2082fa000, 0xd9, 0xc2082fa000)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/encoding/json/encode.go:598 +0x58 fp=0xc2288e0460 sp=0xc2288e0428
encoding/json.(*ptrEncoder).encode(0xc20803a740, 0xc2080a8630, 0x75f120, 0xc2088bd500, 0xd6, 0x77dd100)
	/usr/src/redhat/BUILD/bitly-go-devel-go1.4.2-8/go/src/encoding/json/encode.go:706 +0xeb fp=0xc2288e04b0 sp=0xc2288e0460
```

nsqadmin/static/js/views/lookup.js
mreiferson: Going to do errors as a second pass, i.e. #421

nsqadmin/static/js/views/channel.js
mreiferson: good catch

mreiferson:
@jehiah I wonder if that's an issue when the static assets are built into the binary.  I'll test on my end.

mreiferson:
@jehiah tested locally, not seeing issues - what is the API request that's actually failing? It looks like `/api/topics/:topic` returns successfully and the _next_ request is the one failing?

jehiah:
Yes, that is the request that fails (it 500's with a zero length response). Here is a more complete log. Note that nginx which proxies that request in my configuration recorded a 500 at the 7 second mark.

```
[nsqadmin] 2015/08/03 20:06:32.268167 LOOKUPD: querying http://hostname:4161/lookup?topic=events
[nsqadmin] 2015/08/03 20:06:32.268196 LOOKUPD: querying http://hostname:4161/lookup?topic=events
[nsqadmin] 2015/08/03 20:06:32.268204 LOOKUPD: querying http://hostname:4161/lookup?topic=events
[nsqadmin] 2015/08/03 20:06:32.268212 LOOKUPD: querying http://hostname:4161/lookup?topic=events
[nsqadmin] 2015/08/03 20:06:32.268225 LOOKUPD: querying http://hostname:4161/lookup?topic=events
[nsqadmin] 2015/08/03 20:06:32.286541 NSQD: querying http://hostname:4151/stats?format=json
[nsqadmin] 2015/08/03 20:06:32.286550 NSQD: querying http://hostname:4151/stats?format=json
[nsqadmin] 2015/08/03 20:06:32.286554 NSQD: querying http://hostname:4151/stats?format=json
[nsqadmin] 2015/08/03 20:06:32.286559 NSQD: querying http://hostname:4151/stats?format=json
[nsqadmin] 2015/08/03 20:06:32.286563 NSQD: querying http://hostname:4151/stats?format=json
[nsqadmin] 2015/08/03 20:06:32.384658 200 GET /api/topics/events (127.0.0.1:55898) 116.505ms
runtime: goroutine stack exceeds 1000000000-byte limit
```

Access log. Note the last column w/ response time.
```
remote_ip - - [03/Aug/2015:20:06:40 +0000] "GET /api/topics/events HTTP/1.1" 500 0 "https://nsqadmin/topics/events" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.125 Safari/537.36" "..." ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2 "nsqadmin" 7.921 127.0.0.1:4180 500 7.921
```

jehiah:
@mreiferson I pushed up a few commits here to minify js, add staticmaps, split vendor js into a separate file, and add in gzip compression when serving up responses from nsqadmin. Those all should resolve the previous 1.4M `main.js` content size (which on a not-so-good-connection was taking 30s to load for me).

mreiferson:
@jehiah hah, thanks :+1:

nsqadmin/http.go
```diff
@@ -81,39 +59,42 @@ func NewHTTPServer(ctx *Context) *httpServer {
...
-	router.Handle("POST", "/pause_topic", http_api.Decorate(s.pauseTopicHandler, log, http_api.PlainText))
-	router.Handle("POST", "/unpause_topic", http_api.Decorate(s.pauseTopicHandler, log, http_api.PlainText))
-	router.Handle("POST", "/empty_channel", http_api.Decorate(s.emptyChannelHandler, log, http_api.PlainText))
-	router.Handle("POST", "/delete_channel", http_api.Decorate(s.deleteChannelHandler, log, http_api.PlainText))
-	router.Handle("POST", "/pause_channel", http_api.Decorate(s.pauseChannelHandler, log, http_api.PlainText))
-	router.Handle("POST", "/unpause_channel", http_api.Decorate(s.pauseChannelHandler, log, http_api.PlainText))
-	router.Handle("POST", "/create_topic_channel", http_api.Decorate(s.createTopicChannelHandler, log, http_api.PlainText))
+	router.Handle("GET", "/", http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", "/topics", http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", "/topics/:topic", http_api.Decorate(s.indexHandler, log))
```
jehiah: It's subtle but this plural change from `topic` to `topics` breaks existing links. Is there a strong motivation to change this?

jehiah:
@mreiferson let me know if you have any comments on the api request refactoring. I have one more validation pass to do here (i'm not positive i caught all the template/js fixes from some key changes), but we should be close to done.

nsqadmin/http.go
mreiferson: just consistency with the API endpoint names, I'm fine reverting this

mreiferson: EDIT: reverting _just_ this so that links work

internal/clusterinfo/data.go
```diff
@@ -0,0 +1,399 @@
+package clusterinfo
```
mreiferson: all this simplification came out nice and now that we see the result it looks like there's clear opportunity to DRY it up in some future iteration (a lot of the methods looks suspiciously identical structurally)

internal/clusterinfo/types.go
```diff
@@ -0,0 +1,286 @@
...
+	CipherSuite                   string `json:"tls_cipher_suite"`
+	TLSVersion                    string `json:"tls_version"`
+	TLSNegotiatedProtocol         string `json:"tls_negotiated_protocol"`
+	TLSNegotiatedProtocolIsMutual bool   `json:"tls_negotiated_protocol_is_mutual"`
+}
+
+// UnmarshalJSON does post-processing of ConnectedDuration
+func (s *ClientStats) UnmarshalJSON(b []byte) (err error) {
+	type c ClientStats // used to prevent recursion
+	var i c
```
mreiferson: I'm not sure I understand what's happening here

internal/http_api/api_request.go
```diff
@@ -37,48 +35,58 @@ func NewDeadlineTransport(timeout time.Duration) *http.Transport {
 			}
 			return &deadlinedConn{timeout, c}, nil
 		},
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
mreiferson: I used this to test locally, we should drop it

internal/http_api/api_request.go
```diff
@@ -37,48 +35,58 @@ func NewDeadlineTransport(timeout time.Duration) *http.Transport {
...
 	}
 	return transport
 }
 
 // NegotiateV1 is a helper function to perform a v1 HTTP request
 // and fallback to parsing the old backwards-compatible response format
-func NegotiateV1(method string, endpoint string, body io.Reader) (*simplejson.Json, error) {
+// storing the result in the value pointed to by v.
+func NegotiateV1(method string, endpoint string, v interface{}) error {
```
mreiferson: do we use this for anything other than `GET`, now?  if not, let's drop the `method` param entirely?

internal/http_api/api_request.go
```diff
@@ -112,3 +120,35 @@ func GETV1(endpoint string, v interface{}) error {
...
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("got response %s %q", resp.Status, body)
+	}
+	// err = json.Unmarshal(body, &v)
+	// if err != nil {
+	// 	return err
+	// }
```
mreiferson: drop debugging

internal/http_api/api_response.go
```diff
@@ -43,11 +42,9 @@ func PlainText(f APIHandler) APIHandler {
 		}
 		switch d := data.(type) {
 		case string:
-			w.Header().Set("Content-Length", strconv.Itoa(len(d)))
```
mreiferson: How is it possible for it to be able to calculate the length to provide this automatically for us _before_ we've written the body - it doesn't know the size of the body to expect?

mreiferson: Wait, I think I see what's going on - it's this compression wrapper?

internal/http_api/compress.go
```diff
@@ -0,0 +1,89 @@
...
+					h = nil
+				}
+
+				w = &compressResponseWriter{
+					Writer:         fw,
+					ResponseWriter: w,
+					Hijacker:       h,
+				}
+
+				break L
```
mreiferson: I'm not a fan of the labeled breaks, IMO it's much more clear to `goto label`

nsqadmin/gulpfile.js
```diff
@@ -0,0 +1,153 @@
...
+
+
+function cleanTask(){
+    var paths = Array.prototype.slice.apply(arguments);
+    return function () {
+        gulp.src(paths).pipe(clean());
+    };
+}
+
+gulp.task('vendor-build-js', function() {
```
mreiferson: I feel like we shouldn't have to "one off" this vendored JS builds since we already have a `browserifyTask` helper, but I don't really care _that_ much, thoughts @rolyatmax

internal/http_api/compress.go
```diff
@@ -0,0 +1,89 @@
...
+				w.Header().Set("Content-Encoding", "gzip")
+				w.Header().Add("Vary", "Accept-Encoding")
+
+				gw := gzip.NewWriter(w)
+				defer gw.Close()
+
+				h, hok := w.(http.Hijacker)
+				if !hok { /* w is not Hijacker... oh well... */
+					h = nil
+				}
```
mreiferson: can't this just be simplified to:

```go
h, _ := w.(http.Hijacker)
```

nsqadmin/nsqadmin.go
```diff
@@ -115,7 +115,7 @@ func (n *NSQAdmin) Main() {
 	n.Unlock()
 	httpServer := NewHTTPServer(&Context{n})
 	n.waitGroup.Wrap(func() {
-		http_api.Serve(n.httpListener, httpServer, "HTTP", n.opts.Logger)
+		http_api.Serve(n.httpListener, http_api.CompressHandler(httpServer), "HTTP", n.opts.Logger)
```
mreiferson: In an ideal world we'd have _one_ type of middleware decorator.  I realize you probably did this so that we didn't have to specify it exhaustively per-endpoint, but really we only need it for the two javascript payloads, all the other ones are negligible.

Thoughts?

nsqadmin/static/html/index.html
```diff
@@ -0,0 +1,27 @@
...
+<html>
+<head>
+    <title>nsqadmin</title>
+    <link rel="stylesheet" href="/static/bootstrap.min.css">
+    <link rel="stylesheet" href="/static/base.css">
+    <meta http-equiv="Content-type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+</head>
+<body>
+    <img class="image-preview img-polaroid">
```
mreiferson: I don't think I ever implemented this, maybe we should drop it?

nsqd/nsqd_test.go
```diff
@@ -353,17 +353,22 @@ func TestCluster(t *testing.T) {
...
 
 	// allow some time for nsqd to push info to nsqlookupd
 	time.Sleep(350 * time.Millisecond)
 
-	data, err := http_api.NegotiateV1("GET", fmt.Sprintf("http://%s/debug", lookupd.RealHTTPAddr()), nil)
+	endpoint := fmt.Sprintf("http://%s/debug", lookupd.RealHTTPAddr())
+	d := make(map[string]interface{})
+	err = http_api.NegotiateV1("GET", endpoint, &d)
+	data := simplejson.New()
+	data.SetPath(nil, d)
```
mreiferson: seems like we could use a little helper for this dance we're doing here a thousand times?

internal/http_api/compress.go
jehiah: I had some local changes that are now upstream, but this is just directly imported from `https://github.com/gorilla/handlers`. Would you rather we add that package as a dependency?

internal/clusterinfo/types.go
jehiah: We can't unmarshal directly to something that's a `type` of `ClientStats` because this function **is** the unmarshal code for that (recursion here would be bad); We just have to alias the type to work around that. That's your question right?

internal/http_api/api_response.go
jehiah: `net/http` will actually set a content-length for us in most cases (depending on how content is written out). This code here became useless because of the compression middleware.

nsqadmin/nsqadmin.go
jehiah: I prefer this globally because the responses from some of the stats api endpoints could be pretty big and we want to compress as much as possible back to the client, (and json api responses will compress well). (and also as i recently realized I had a 250k response from nsqd `/stats` we may want to use this elsewhere too). Really, the only thing we wouldn't particularly want to compress is proxied graphite requests, but i think i'd rather address that with a future PR that adds mime type opt-out on the compression middleware.

nsqadmin/gulpfile.js
jehiah: I'd love any cleanup/refactoring of this, but I think the two general benefits are:

* Less large blob changes in git history presuming the main.js will change much more frequently than vendor'd items. (at worst it's no better than in a single js).
* vendored items aren't staticMap'd (resulting in a much smaller staticMap)

internal/http_api/api_request.go
jehiah: good catch.

nsqadmin/gulpfile.js
mreiferson: I meant specifically re-using some code in this gulpfile to produce that vendor build, not that we shouldn't produce a vendor JS :grin:

nsqadmin/nsqadmin.go
mreiferson: ok, maybe leave a comment at the top of the file pointing out where it's copied from then?

internal/http_api/compress.go
mreiferson: :+1:, see comment about leaving comment

internal/http_api/api_response.go
mreiferson: :+1:since we're keeping the global wrapper it doesn't matter

internal/clusterinfo/types.go
mreiferson: we can't just decode into `s` (the receiver) and then override fields?

jehiah: nope, that's the recursion. `json.Unmarshal` would see that `s` implements the `Unmarshaler` interface with this function, so it would call us again. When we alias with another type, `type c` does not implement the `Unmarshaler` so `json.Unmarshal` can do it's job.

My variables admittedly are pretty terrible here. Let me try to improve them some.

nsqd/nsqd_test.go
jehiah: haha clearly i was just in brute force make-it-work mode. fixed

mreiferson:
cool, LGTM - let's land this :airplane: 

Want to try to squash 091eabc on?

jehiah:
Close. two things i found in validation

1) Bugs getting per-nsqd rate on the topics page (this should be easy; i think it's from variable renames)
2) some bugs on the Counter page (it's not working correctly for me). Not sure what's happening there yet.

jehiah:
I spoke too soon. A few more items:

* It seems the default timeframe for graphite charts is defaulting to 'off' I expect that to default to 2h when it's not set via a cookie.
* The 'attributes' section which should display TLS and auth state isn't displaying TLS details, or auth state. (see image below)

screen shot 2015-08-06 at 12 04 20 pm

mreiferson:
> It seems the default timeframe for graphite charts is defaulting to 'off' I expect that to default to 2h when it's not set via a cookie.

I think it's now in local storage, should be easy to default it to `2h`

mreiferson:
> some bugs on the Counter page (it's not working correctly for me). Not sure what's happening there yet.

I didn't translate the JS 1:1 here, it's just broken or?

jehiah:
For the counter page, i'm getting Zero's back from the /api/counter endpoint, but also the graphite URLS have unexpected `&amp` instead of just `&`.

jehiah:
I think i fixed all the previously mentioned bugs (please take a look before i squash those); glyphicon fonts are not being embedded properly and i think that's the only outstanding thing left.

mreiferson:
> For the counter page, i'm getting Zero's back from the /api/counter endpoint, but also the graphite URLS have unexpected &amp instead of just &.

Did you figure this out?

jehiah:
>> For the counter page, i'm getting Zero's back from the /api/counter endpoint, but also the graphite URLS have unexpected &amp instead of just &.

> Did you figure this out?

Yes, it was broken handling of `message_count` also affecting the Message Count columns (i just didn't notice on my dev instance that it was always Zero). Fixed by f9cdd4f3a06906f7d51d9b4e41892a2fc726c6db

mreiferson:
cool :hammer:

jehiah:
@mreiferson font issues addressed; take a look at last two commits?

mreiferson:
@jehiah don't you think the console.log might be useful when people inevitably encounter bugs?

mreiferson:
also, do we need _another_ HTTP route for the same handler?  Is it because bootstrap expects them at that URL?

jehiah:
yeah, bootstrap requests them as `fonts/....` and that felt like the easiest path to resolving since the gulp step currently flattens directories. I'd rather one line of Go to more gulp =)

I have "no console.log in production" burned into my brain since it's not present in all browsers and thus causes JS errors (Thanks IE 9!)

mreiferson:
very well, squash time?

mreiferson:
:scream:

klucar:
Has anyone started work on refactoring the UI for #443? We get i/o timeout errors when nsq is under heavy load. I'd like to get rid of this timeout, make it configurable or ?

I have experience with doing ReactJS interfaces and this is kind of a perfect use-case for it. If nobody is already working this, I'll give it a go.

mreiferson:
@klucar we made a bunch of improvements recently in https://github.com/nsqio/nsq/pull/701 that should significantly improve this situation, if you're interested in deploying an `alpha`

klucar:
we're just evaluating if NSQ can work/scale for us, so I'll gladly give an alpha a shot :)

klucar:
@mreiferson any idea when a new release will be cut?

mreiferson:
@klucar I've been lazy, we should get one out (I've been running it in production without issue for a few weeks)

PR #520, opened 2015-01-07
Adding support for gossip as an alternative to lookupd

stephensearles:


mreiferson:
@stephensearles thanks for pushing this forward, I have a few comments I'll put inline

util/registrationdb/registrationdb.go
```diff
@@ -0,0 +1,356 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// RegistrationDB provides storage and lookup mechanisms for NSQ topic producers.
+type RegistrationDB struct {
+	mtx     sync.RWMutex
+	data    map[RegistryKey]Registry
+	clients map[string]*Producer
```
mreiferson: I'm not a big fan of building explicit knowledge of clients into `RegistrationDB`.  It was intentionally designed to be agnostic to the data that we put in, for the most part.

Granted, it is only used inside `nsqlookupd` (and now `nsqd`) but I think that approach is still relevant.

Basically, I don't see the benefit of conflating these sets of changes in this PR.  Ideally the semantics and API of `RegistrationDB` would remain the same (save for obvious cleanup/bugs/etc.) and this PR would focus on the actual core changes to introduce gossip.  I think this will reduce the noise and make it easier to actually land this PR confidently.

What do you think?

mreiferson:
Having dug in a little more now I see there are quite a few changes to `RegistrationDB` in this PR.  I'll admit that its API might benefit from improvement but I can't help but feel like some of this is busy work.

I guess, more generally, why don't we outline and discuss the changes you're proposing to `RegistrationDB` and settle on things that make sense and are necessary (for whatever reason) for these gossip changes.

Perhaps we can defer some of the naming/cleanup changes to a follow-up PR.

stephensearles:
Yeah, I think part of the problem was that I misunderstood the design of `RegistrationDB` while I did most of that work. I do think it could stand some improvement, but I agree that it may work better in a separate PR.

mreiferson:
gotta re-add the serf dependency to Godeps

mreiferson:
this diff fixed things up for me (serf must have added `AdvertisePort` at some point):

```diff
diff --git a/Godeps b/Godeps
index 36f5b21..43ed0a5 100644
--- a/Godeps
+++ b/Godeps
@@ -6,3 +6,4 @@ github.com/bitly/go-simplejson          fc395a5db941cf38922b1ccbc083640cd76fe4bc
 github.com/bmizerany/perks/quantile     6cb9d9d729303ee2628580d9aec5db968da3a607
 github.com/mreiferson/go-options        2cf7eb1fdd83e2bb3375fef6fdadb04c3ad564da
 github.com/mreiferson/go-snappystream   307a466b220aaf34bcee2d19c605ed9e96b4bcdb # v0.2.0
+github.com/hashicorp/serf               5e0771b8d61bee28986087a246f7611d6bd4a87a # v0.6.3
diff --git a/nsqd/gossip.go b/nsqd/gossip.go
index 0c26c74..df6ca46 100644
--- a/nsqd/gossip.go
+++ b/nsqd/gossip.go
@@ -82,8 +82,8 @@ func initSerf(opts *nsqdOptions,
        serfConfig.Tags["h"] = hostname
        serfConfig.Tags["v"] = util.BINARY_VERSION
        serfConfig.NodeName = net.JoinHostPort(opts.BroadcastAddress, strconv.Itoa(tcpAddr.Port))
-
        serfConfig.MemberlistConfig.AdvertiseAddr = broadcastAddr.String()
+       serfConfig.MemberlistConfig.AdvertisePort = gossipAddr.Port
        serfConfig.MemberlistConfig.BindAddr = gossipAddr.IP.String()
        serfConfig.MemberlistConfig.BindPort = gossipAddr.Port
        serfConfig.MemberlistConfig.GossipInterval = 100 * time.Millisecond
```

nsqd/gossip_test.go
```diff
@@ -0,0 +1,141 @@
...
+			gossipNotifyCh <- struct{}{}
+		}
+	}
+
+	num := 3
+	for i := 0; i < num; i++ {
+		var nsqd *NSQD
+
+		// find an open port
+		tmpl, err := net.Listen("tcp", hostname+":0")
```
mreiferson: I don't understand why we need to use hostname here?

mreiferson: basically none of these hostname and other ([recent](https://github.com/bitly/nsq/compare/91a2119...2082caf)) address related changes make sense (for the test specifically)

If we had been explicitly telling it to listen on localhost (setting `GossipAddress`) why is this any different?

nsqd/gossip_test.go
```diff
@@ -0,0 +1,141 @@
...
+		opts := NewNSQDOptions()
+		opts.ID = int64(i)
+		opts.Logger = newTestLogger(t)
+		opts.GossipAddress = addr.String()
+		if seedNode != nil {
+			sn := net.JoinHostPort(seedNodeAddr, fmt.Sprint(seedNode.serf.Memberlist().LocalNode().Port))
+			opts.SeedNodeAddresses = []string{sn}
+		}
+		tcpAddr, _, nsqd := mustStartNSQD(opts)
+		defer func(n *NSQD) {
```
mreiferson: this new defer shouldn't be necessary, I don't think it was the cause of your "test never ending" issue

nsqd/gossip_test.go
```diff
@@ -0,0 +1,141 @@
...
+		producers = nsqd.rdb.FindProducers("channel", topicName, "ch")
+		equal(t, len(producers), 1)
+		equal(t, producers[0].TCPPort, firstPort)
+	}
+}
+
+func converge(timeout time.Duration, nsqds []*NSQD, convergence func() bool) {
+	convergingLock.Lock()
+	converging = true
+	convergingLock.Unlock()
```
mreiferson: these locks are really hard to follow, there's got to be a way to do this without locking on global state

nsqd/gossip.go
```diff
@@ -0,0 +1,361 @@
...
+			case serf.EventUser:
+				n.serfUserEvent(ev)
+			case serf.EventQuery:
+				// nothing
+			case serf.EventMemberUpdate:
+				// nothing
+			default:
+				n.logf("WARNING: un-handled Serf event: %#v", ev)
+			}
+			gossipNotify()
```
mreiferson: We're going to need to improve this API and how it's configured :grin:

nsqd/nsqd.go
```diff
@@ -504,6 +541,11 @@ func (n *NSQD) Notify(v interface{}) {
 			n.Unlock()
 		}
 	})
+
+	select {
```
mreiferson: you were correct, this does need a waitgroup after all (after seeing it in context when I played around with the code it made sense)

nsqd/nsqd.go
```diff
@@ -365,6 +399,9 @@ func (n *NSQD) Exit() {
 		n.httpsListener.Close()
 	}
 
+	n.serf.Leave()
```
mreiferson: I don't think we want to `Leave` before shutting down - specifically because we want to be able to rejoin as the exact same node if the process was just restarting.

I don't recall the exact nature of how this works in `serf`, so I should probably re-educate myself.

nsqd/gossip_test.go
stephensearles: yeah, fixed

nsqd/gossip_test.go
```diff
@@ -0,0 +1,141 @@
...
+		equal(t, tcpPorts, actTCPPorts)
+	}
+
+	// create a topic/channel on the first node
+	topicName := "topic1"
+	topic := nsqds[0].GetTopic(topicName)
+	topic.GetChannel("ch")
+	firstPort := nsqds[0].tcpListener.Addr().(*net.TCPAddr).Port
+
+	converge(10*time.Second, nsqds, func() bool {
```
mreiferson: conceptually I like the approach you're taking to make the tests not depend on time, nice work.

aside from my other comments relating to how we actually implement that callback within `nsqd` and some of the global state locking, what do you think of this for the body of the function?

```go
for _, nsqd := range nsqds {
    if len(nsqd.rdb.FindProducers("topic", topicName, "")) != 1 ||
        len(nsqd.rdb.FindProducers("channel", topicName, "ch")) != 1 {
        return false
    }
    return true
}
```

nsqd/nsqd.go
stephensearles: Yeah, this is another remnant of fixing that never-ending-test issue

nsqd/gossip.go
stephensearles: the "gossipNotify" thing? Yeah, that's just an unexposed mechanism to support the convergence checking in the tests. I'd be happy to improve it if you have suggestions

nsqd/stats.go
```diff
@@ -52,7 +53,7 @@ func NewChannelStats(c *Channel, clients []ClientStats) ChannelStats {
 		BackendDepth:  c.backend.Depth(),
 		InFlightCount: len(c.inFlightMessages),
 		DeferredCount: len(c.deferredMessages),
-		MessageCount:  c.messageCount,
+		MessageCount:  atomic.LoadUint64(&c.messageCount),
```
mreiferson: If this is truly necessary (and I think you're right, it might be) then it's probably necessary for *all* of these fields that are modified atomically by `Topic` or `Channel`.

We should keep this change as a separate commit, too.

p.s. did you encounter some race condition?  what prompted the change?

Godeps
```diff
@@ -6,3 +6,6 @@ github.com/bitly/go-simplejson          fc395a5db941cf38922b1ccbc083640cd76fe4bc
 github.com/bmizerany/perks/quantile     6cb9d9d729303ee2628580d9aec5db968da3a607
 github.com/mreiferson/go-options        2cf7eb1fdd83e2bb3375fef6fdadb04c3ad564da
 github.com/mreiferson/go-snappystream   307a466b220aaf34bcee2d19c605ed9e96b4bcdb # v0.2.0
+github.com/hashicorp/serf               5e0771b8d61bee28986087a246f7611d6bd4a87a # v0.6.3
+github.com/hashicorp/memberlist         9a1e242e454d2443df330bdd51a436d5a9058fc4
+github.com/ugorji/go/codec              006e1534301cb75b848ee452ab5d3ba8c6a70784
```
mreiferson: this is an unfortunate dependency, that repository is a mess

nsqd/stats.go
stephensearles: Yeah, the race detector yelled at me.

Tangentially related, I think the test.sh script may not be actually failing if race conditions are found. It looks like `go test -race` still returns an exit code of `0` if the tests pass, but a race is found. I just noticed this race is indeed happening on master.

nsqd/gossip_test.go
stephensearles: Small mistake in what you suggested, but other than that it seems good :)

nsqd/stats.go
mreiferson: strange, I've never seen it report this as a race - although it clearly could be...

nsqd/gossip_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestGossip(t *testing.T) {
+	var nsqds []*NSQD
+	var seedNode *NSQD
+	var tcpPorts []int
+
+	gossipNotify = func() {
+		convergingLock.Lock()
+		defer convergingLock.Unlock()
+		if converging {
+			gossipNotifyCh <- struct{}{}
```
mreiferson: ok, I think you added this lock so that you could add logic around this channel send - if we made this channel send non-blocking, ie:

```go
select {
case gossipNotifyCh <- struct{}{}
default:
}
```

then we wouldn't need any of the locks... which would simplify all of this, particularly `converge()`

nsqd/gossip_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// create a topic/channel on the first node
+	topicName := "topic1"
+	topic := nsqds[0].GetTopic(topicName)
+	topic.GetChannel("ch")
+	firstPort := nsqds[0].tcpListener.Addr().(*net.TCPAddr).Port
+
+	converge(10*time.Second, nsqds, func() bool {
+		for _, nsqd := range nsqds {
+			if len(nsqd.rdb.FindProducers("topic", topicName, "")) != 1 || len(nsqd.rdb.FindProducers("channel", topicName, "ch")) != 1 {
```
mreiferson: let's newline the 2nd logic check?

nsqd/gossip_test.go
```diff
@@ -0,0 +1,128 @@
...
+	sort.Ints(tcpPorts)
+
+	// wait for convergence
+	converge(5*time.Second, nsqds, func() bool {
+		converged := true
+		for _, nsqd := range nsqds {
+			producers := nsqd.rdb.FindProducers("client", "", "")
+			converged = converged && len(producers) == num
+		}
+		return converged
```
mreiferson: same aesthetic changes should be applied to this func, too, please

mreiferson:
OK, so after pulling down these changes and testing locally, I can apply the following diff and the tests pass (the diff is *really* just removing all the networking related changes, leaving only the improvements to the test related to convergence).

This is inline with what I expected - I feel like the hostnames on your local machine aren't resolving correctly and I'm not convinced this implies that the implementation needs to change.  Also, the code *reads* correct to me with this diff applied (as we briefly discussed in `#nsq`).

In a production deployment these knobs would all be configurable, so I don't think we're limiting things in any way.

```diff
diff --git a/nsqd/gossip.go b/nsqd/gossip.go
index 09df7c7..4e374ce 100644
--- a/nsqd/gossip.go
+++ b/nsqd/gossip.go
@@ -3,7 +3,6 @@ package nsqd
 import (
        "bytes"
        "encoding/json"
-       "fmt"
        "net"
        "os"
        "strconv"
@@ -62,18 +61,17 @@ func memberToProducer(member serf.Member) *registrationdb.Producer {
 
 func initSerf(opts *nsqdOptions,
        serfEventChan chan serf.Event,
-       tcpAddr *net.TCPAddr, httpAddr *net.TCPAddr, httpsAddr *net.TCPAddr, broadcastAddr *net.TCPAddr) (*serf.Serf, error) {
-       hostname, err := os.Hostname()
-       if err != nil {
-               return nil, err
-       }
+       tcpAddr *net.TCPAddr,
+       httpAddr *net.TCPAddr,
+       httpsAddr *net.TCPAddr,
+       broadcastAddr *net.TCPAddr) (*serf.Serf, error) {
 
-       gossipHost, gossipPortStr, err := net.SplitHostPort(opts.GossipAddress)
+       hostname, err := os.Hostname()
        if err != nil {
                return nil, err
        }
 
-       gossipPort, err := strconv.Atoi(gossipPortStr)
+       gossipAddr, err := net.ResolveTCPAddr("tcp", opts.GossipAddress)
        if err != nil {
                return nil, err
        }
@@ -90,15 +88,10 @@ func initSerf(opts *nsqdOptions,
        serfConfig.Tags["h"] = hostname
        serfConfig.Tags["v"] = util.BINARY_VERSION
        serfConfig.NodeName = net.JoinHostPort(opts.BroadcastAddress, strconv.Itoa(tcpAddr.Port))
-
-       broadcastAddr.Port = gossipPort
-       opts.Logger.Output(0, fmt.Sprintf("serf: advertising: %s", broadcastAddr))
-       opts.Logger.Output(0, fmt.Sprintf("serf: binding: %s:%d", gossipHost, gossipPort))
-
        serfConfig.MemberlistConfig.AdvertiseAddr = broadcastAddr.IP.String()
-       serfConfig.MemberlistConfig.AdvertisePort = gossipPort
-       serfConfig.MemberlistConfig.BindAddr = gossipHost
-       serfConfig.MemberlistConfig.BindPort = gossipPort
+       serfConfig.MemberlistConfig.AdvertisePort = gossipAddr.Port
+       serfConfig.MemberlistConfig.BindAddr = gossipAddr.IP.String()
+       serfConfig.MemberlistConfig.BindPort = gossipAddr.Port
        serfConfig.MemberlistConfig.GossipInterval = 100 * time.Millisecond
        serfConfig.MemberlistConfig.GossipNodes = 5
        serfConfig.MemberlistConfig.LogOutput = logWriter{opts.Logger, []byte("memberlist:")}
diff --git a/nsqd/gossip_test.go b/nsqd/gossip_test.go
index 8b23139..b504ae4 100644
--- a/nsqd/gossip_test.go
+++ b/nsqd/gossip_test.go
@@ -33,9 +33,8 @@ func TestGossip(t *testing.T) {
                opts.ID = int64(i)
                opts.Logger = newTestLogger(t)
                opts.GossipAddress = addr.String()
-               opts.BroadcastAddress = addr.IP.String()
                if seedNode != nil {
-                       opts.SeedNodeAddresses = []string{seedNode.broadcastAddr.String()}
+                       opts.SeedNodeAddresses = []string{seedNode.opts.GossipAddress}
                }
                tcpAddr, _, nsqd := mustStartNSQD(opts)
                defer nsqd.Exit()
```

mreiferson:
re: your question about how this PR should handle `nsqlookupd`...

As implemented `nsqlookupd` is still fully supported.  I don't see any reason to change that. It means that `nsqd` can still register like it always has with an existing `nsqlookupd` cluster and opens up the door to a variety of operational choices teams can make on how to transition.

A future release can add deprecation warnings, etc. before finally removing it when we're happy & ready.

PR #523, opened 2015-01-12
nsq_stat: add non loop mode support

nordicdyno:
with negative value of -status-every shows only one stats line

N.B.
based on our OPs request, it's easy way to monitor capacity of channels without any overhead for them (just standard toolbox) 
also sometimes basic stats it's all we need

### DIFF
--- a/apps/nsq_stat/nsq_stat.go
+++ b/apps/nsq_stat/nsq_stat.go
@@ -20,12 +20,14 @@ import (
 )
 
 var (
-	showVersion      = flag.Bool("version", false, "print version")
-	topic            = flag.String("topic", "", "NSQ topic")
-	channel          = flag.String("channel", "", "NSQ channel")
-	statusEvery      = flag.Duration("status-every", 2*time.Second, "duration of time between polling/printing output")
+	showVersion = flag.Bool("version", false, "print version")
+	topic       = flag.String("topic", "", "NSQ topic")
+	channel     = flag.String("channel", "", "NSQ channel")
+	statusEvery = flag.Duration("status-every", 2*time.Second,
+		"duration of time between polling/printing output (show stats only once if value is negative, i.e. -1s)")
 	nsqdHTTPAddrs    = util.StringArray{}
 	lookupdHTTPAddrs = util.StringArray{}
+	onceRun          = false
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


nordicdyno:
Is there any chance to merge? I would like see it in next 3.x release if it's ok. :beers:  Is it possible? :flushed:

mreiferson:
hmmmm, I think the code is fine - I'm just unsure about the usage semantics...

Ideally it would mimic the behavior of `iostat` and friends - perhaps it makes sense to have the default spit out 1 line and specify the interval for repeating?

/cc @jehiah

nordicdyno:
Probably it would be fine. But this new behaviour would break backward compatibility. Maybe it would be better to implement nsq_stat with new semantic and behaviour in 0.4.x branch.

Also I have idea to add support for printing cummulative topic allChannelStats (but this topic for other issue/feature/pull request)

jehiah:
What do you think about a `-count` flag to output a certain number of iterations instead of overload the interval flag.

nordicdyno:
if PR #524 is fine, it's would be ok to close current PR and continue discussion there :four_leaf_clover:

mreiferson:
I think `--count` would provide a better backwards compatible option but at the expense of an even more confusing interface.

I think the options are:

1. break backwards compatibility and make it work like `iostat` ("best" familiar interface, worst upgrade)
2. use this implementation here (or some other proposal that is backwards compatible but more user friendly -- maybe a `--once` flag?)
3. do nothing and suggest forking (we are talking about 100 lines of code) :dancers: 

???

jehiah:
I like the ability to have `--count=6 --status-every=10s` which is similar to `iostat 10 6` just with named arguments (I'm not really sure what you mean by "make it work like `iostat`). It makes me wish `--status-every` was named `--interval` but i like it.

mreiferson:
OK, that makes more sense actually, good pt.

What I specifically meant was that, by default, `iostat` doesn't iterate (so by your new proposal that would mean `--status-every=0 --count=1` default arguments).  This would actually be the backwards incompatible change.

re: naming, I like `--interval` much better, we could easily transition via deprecation :+1:

jehiah:
ok so let's use `--count` in #524 and we can separately transition from `--status-every` to `--interval`. 

We can then take steps to change our defaults but i'm not yet convinced we should.

mreiferson:
SGTM, closing

PR #524, opened 2015-01-14
nsq_stat: add -count flag

nordicdyno:
based on #523 discussion

### DIFF
--- a/apps/nsq_stat/nsq_stat.go
+++ b/apps/nsq_stat/nsq_stat.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


apps/nsq_stat/nsq_stat.go
```diff
@@ -80,6 +81,10 @@ func statLoop(interval time.Duration, topic string, channel string,
 			continue
 		}
 
+		if *countNum > 0 && (i > *countNum) {
```
mreiferson: can't we just add this as the conditional in the `for` loop we're in?

nordicdyno:
as you wish :)
also i've added fix&checks (in separated commit)

apps/nsq_stat/nsq_stat.go
```diff
@@ -36,7 +37,7 @@ func init() {
 func statLoop(interval time.Duration, topic string, channel string,
 	nsqdTCPAddrs []string, lookupdHTTPAddrs []string) {
 	var o *lookupd.ChannelStats
-	for i := 0; ; i++ {
+	for i := 0; *countNum == 0 || *countNum > i-1; i++ {
```
mreiferson: ` >= i` :grin:

apps/nsq_stat/nsq_stat.go
```diff
@@ -120,6 +122,14 @@ func main() {
 		log.Fatal("--topic and --channel are required")
 	}
 
+	if int64(*statusEvery) <= 0 {
+		log.Fatal("--status-every should be positive")
+	}
+
+	if int64(*countNum) < 0 {
+		log.Fatal("--count should be positive or equals zero")
```
mreiferson: which means I think this should also assert positive, `count = 0` doesn't make sense, right?

apps/nsq_stat/nsq_stat.go
nordicdyno: agreed

apps/nsq_stat/nsq_stat.go
nordicdyno: Yes, it sounds logical. But `-count` is optional flag and must to have default value because of flag API.
So I don't know how to solve it in elegant way.

mreiferson: ahhh, you are correct - it needs to accept `0` as "don't check count", sorry!

nordicdyno:
I've done this patch, probably. :sunglasses:

mreiferson:
LGTM

Would you mind squashing down the commits?

Thanks!

nordicdyno:
:sleeping:

mreiferson:
😍

PR #379, opened 2014-06-20
top secret

mreiferson:
:smiling_imp:

PR #526, opened 2015-01-21
nsqd: gossip

mreiferson:
supercedes #520 and #379 

TODO:

- [x] some sort of stats/health endpoint integration
- [x] add new config vars to example configs in `contrib/`
- [x] serf encryption / security questions

cc @stephensearles

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -92,7 +92,7 @@ func nsqFlagset() *flag.FlagSet {
 	lookupdTCPAddrs := util.StringArray{}
 	flagSet.Var(&lookupdTCPAddrs, "lookupd-tcp-address", "lookupd TCP address (may be given multiple times)")
 
-	flagSet.String("gossip-address", "0.0.0.0:7946", "<addr>:<port> to listen on for gossip")
+	flagSet.String("gossip-address", "", "<addr>:<port> to listen on for gossip. reasonable default: 0.0.0.0:7946")
 	seedNodeAddrs := util.StringArray{}
 	flagSet.Var(&seedNodeAddrs, "seed-node-address", "TCP address of an nsqd serving as a seed node to bootstrap gossip protocol (may be given multiple times)")
 
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -92,6 +92,10 @@ func nsqFlagset() *flag.FlagSet {
 	lookupdTCPAddrs := util.StringArray{}
 	flagSet.Var(&lookupdTCPAddrs, "lookupd-tcp-address", "lookupd TCP address (may be given multiple times)")
 
+	flagSet.String("gossip-address", "0.0.0.0:7946", "<addr>:<port> to listen on for gossip")
+	seedNodeAddrs := util.StringArray{}
+	flagSet.Var(&seedNodeAddrs, "seed-node-address", "TCP address of an nsqd serving as a seed node to bootstrap gossip protocol (may be given multiple times)")
+
 	// diskqueue options
 	flagSet.String("data-path", "", "path to store disk-backed messages")
 	flagSet.Int64("mem-queue-size", 10000, "number of messages to keep in memory (per topic/channel)")
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -94,7 +94,8 @@ func nsqFlagset() *flag.FlagSet {
 
 	flagSet.String("gossip-address", "", "<addr>:<port> to listen on for gossip. reasonable default: 0.0.0.0:7946")
 	seedNodeAddrs := util.StringArray{}
-	flagSet.Var(&seedNodeAddrs, "seed-node-address", "TCP address of an nsqd serving as a seed node to bootstrap gossip protocol (may be given multiple times)")
+	flagSet.Var(&seedNodeAddrs, "gossip-seed-address", "TCP address of an nsqd serving as a seed node to bootstrap gossip protocol (may be given multiple times)")
+	flagSet.Duration("gossip-regossip-interval", 60*time.Second, "how often node will re-gossip topic and channel information with known peers")
 
 	// diskqueue options
 	flagSet.String("data-path", "", "path to store disk-backed messages")
--- a/contrib/nsqd.cfg.example
+++ b/contrib/nsqd.cfg.example
@@ -21,6 +21,16 @@ nsqlookupd_tcp_addresses = [
     "127.0.0.1:4160"
 ]
 
+## publishing a port for participating in gossip cluster
+gossip_address = "0.0.0.0:7946"
+
+## how often to resend gossiped topic/channel information
+gossip_regossip_interval = "1m0s"
+
+## seeding gossip
+# gossip_seed_address = [
+#       "192.168.0.10:7946"
+# ]
 
 ## path to store disk-backed messages
 # data_path = "/var/lib/nsq"
--- a/nsqd/gossip.go
+++ b/nsqd/gossip.go
@@ -282,10 +282,6 @@ func (n *NSQD) gossip(evName string, topicName string, channelName string) error
 }
 
 func (n *NSQD) gossipLoop() {
-	if n.serf == nil {
-		return
-	}
-
 	var evName string
 	var topicName string
 	var channelName string
--- a/nsqd/gossip.go
+++ b/nsqd/gossip.go
@@ -265,6 +265,9 @@ exit:
 }
 
 func (n *NSQD) gossip(evName string, topicName string, channelName string) error {
+	if n.serf == nil {
+		return nil
+	}
 	gev := gossipEvent{
 		Name:    n.serf.LocalMember().Name,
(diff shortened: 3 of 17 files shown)


apps/nsqd/nsqd.go
```diff
@@ -92,6 +92,10 @@ func nsqFlagset() *flag.FlagSet {
 	lookupdTCPAddrs := util.StringArray{}
 	flagSet.Var(&lookupdTCPAddrs, "lookupd-tcp-address", "lookupd TCP address (may be given multiple times)")
 
+	flagSet.String("gossip-address", "", "<addr>:<port> to listen on for gossip. reasonable default: 0.0.0.0:7946")
+	seedNodeAddrs := util.StringArray{}
+	flagSet.Var(&seedNodeAddrs, "seed-node-address", "TCP address of an nsqd serving as a seed node to bootstrap gossip protocol (may be given multiple times)")
```
jehiah: Naming: `gossip-seed-address`? It would be nice for this to sort with the flag above

nsqd/nsqd.go
```diff
@@ -504,6 +545,13 @@ func (n *NSQD) Notify(v interface{}) {
 			n.Unlock()
 		}
 	})
+
+	n.waitGroup.Wrap(func() {
+		select {
+		case <-n.exitChan:
+		case n.gossipChan <- v:
```
jehiah: why this? If needed can we comment this block of code?

nsqd/gossip.go
```diff
@@ -0,0 +1,371 @@
...
+				}
+				// otherwise only send a channel, implying the existence of the topic
+				for _, channelStat := range topicStat.Channels {
+					err := n.gossip("channel=", topicStat.TopicName, channelStat.ChannelName)
+					if err != nil {
+						n.logf("ERROR: failed to send Serf user event - %s", err)
+					}
+				}
+			}
+		case v := <-n.gossipChan:
```
jehiah: where do items on this channel come from? i'm having trouble tracking it down

nsqd/http.go
```diff
@@ -226,6 +231,32 @@ func (s *httpServer) getTopicFromQuery(req *http.Request) (url.Values, *Topic, e
 	return reqParams, s.ctx.nsqd.GetTopic(topicName), nil
 }
 
+func (s *httpServer) doLookup(req *http.Request) (interface{}, error) {
```
jehiah: this should error `GOSSIP_NOT_ENABLED` when serf/gossip isn't enabled?

jehiah: We should be able to share this code (and the implementation for `/topics`, `/channels`, and `/nodes`) across `nsqd` and `nsqlookupd` if we write this as a function that takes in the registration db:

```go
func (s *httpServer) doLookup(req *http.Request, rdb *registrationdb.RegistrationDB) (interface{}, error) {
...
}
```

Adding the other two makes this close to sufficient to back nsqadmin (though we should probably namespace these endpoints under `/lookup/...`).

nsqd/nsqd.go
```diff
@@ -223,6 +239,27 @@ func (n *NSQD) Main() {
...
+			n.broadcastAddr)
+		if err != nil {
+			n.logf("FATAL: failed to initialize Serf - %s", err)
+			os.Exit(1)
+		}
+		n.serf = serf
+	}
+
+	n.waitGroup.Wrap(func() { n.serfEventLoop() })
+	n.waitGroup.Wrap(func() { n.gossipLoop() })
```
jehiah: i'm not intimately familiar enough to know if it's better to put these inside the `GossipAddress` check above. I see `serfEventLoop` does a short circuit check on `self.serf != nil` but don't see anything similar for `gossipLoop`. Feels like they should both have short circuit checks if it doesn't complicate cleanup logic.

jehiah:
nice work @mreiferson @stephensearles general feedback is that this looks great; I'd love to see more docstrings on some of the functions (++ for adding them to registrationdb) but i understand how that goes.

nsqd/gossip.go
stephensearles: here: https://github.com/bitly/nsq/pull/526/files#diff-13e146c62dc03f79a6c5963c9c13d4cdR552

stephensearles:
Basically just a typo:

```
diff --git a/nsqd/gossip_test.go b/nsqd/gossip_test.go
index 3fb0b2c..c1a6cd4 100644
--- a/nsqd/gossip_test.go
+++ b/nsqd/gossip_test.go
@@ -106,7 +106,7 @@ func TestGossip(t *testing.T) {
 	}
 }
 
-func converge(timeout time.Duration, nsqds []*NSQD, notifyChan chan struct, isConverged func() bool) {
+func converge(timeout time.Duration, nsqds []*NSQD, notifyChan chan struct{}, isConverged func() bool) {
 	// wait for convergence
 	converged := false
 	t := time.NewTimer(timeout)
```

nsqd/nsqd.go
jehiah: oh ok this is in Nofify(); I misread this as a different block of code.

nsqd/gossip.go
jehiah: thanks.

apps/nsqd/nsqd.go
mreiferson: yea, good idea

nsqd/nsqd.go
mreiferson: still think it needs documenting? (I don't, given it's context within the `Notify` func)

nsqd/http.go
mreiferson: :+1: to that error - related I think /health or /stats should report gossip status... thoughts?

I'll look into code re-use...

nsqd/nsqd.go
mreiferson: yes, they should be inside that block and remove the escape logic within the funcs (I prefer doing it externally)

nsqd/nsqd.go
jehiah: nah i'm ok; the larger the block of code, the less i like single letter variables (`v` in this case), but it's fine.

nsqd/http.go
jehiah: **something** should expose gossip status `/gossip/status` ?

I'm not sure how much of the `lookupd` endpoints you wanted to land here vs follow on PR's; Some translate easily, and some have some things i hadn't thought about (tombstoning, pre-creating in registry db, deleting, etc) for making this fully back nsqadmin.

mreiferson: Yes, I just want it to be easy and obvious to both programmatically or manually inspect/confirm the state of a node WRT to the gossip cluster (this seems like an obvious thing to alert on).

mreiferson: > I'm not sure how much of the lookupd endpoints you wanted to land here vs follow on PR's; Some translate easily, and some have some things i hadn't thought about (tombstoning, pre-creating in registry db, deleting, etc) for making this fully back nsqadmin.

Let's defer this conversation to a future PR because there are other alternatives.  For one, I have #323 inching closer to completion.  And this work here paves the way for nsqadmin to *participate* in the gossip cluster as a means of having data pushed to it + querying :flushed:.

stephensearles:
Another thing to tackle before merging: extending encryption/security options to serf

mreiferson:
FYI added a todo for config vars and the security questions

contrib/nsqd.cfg.example
```diff
@@ -21,6 +21,13 @@ nsqlookupd_tcp_addresses = [
     "127.0.0.1:4160"
 ]
 
+## publishing a port for participating in gossip cluster
+gossip_address = "0.0.0.0:7946"
+
+## seeding gossip
+# gossip-seed-address = [
+#       "192.168.0.10:7946"
+# ]
```
mreiferson: need `gossip_regossip_interval`

contrib/nsqd.cfg.example
```diff
@@ -21,6 +21,13 @@ nsqlookupd_tcp_addresses = [
     "127.0.0.1:4160"
 ]
 
+## publishing a port for participating in gossip cluster
+gossip_address = "0.0.0.0:7946"
+
+## seeding gossip
+# gossip-seed-address = [
```
mreiferson: `s/-/_`

nsqd/gossip.go
```diff
@@ -0,0 +1,402 @@
...
+	strKey := base64.StdEncoding.EncodeToString(key)
+	_, err = n.serf.KeyManager().InstallKey(strKey)
+	if err != nil {
+		return err
+	}
+	_, err = n.serf.KeyManager().UseKey(strKey)
+	if err != nil {
+		return err
+	}
+	_, err = n.serf.KeyManager().RemoveKey(string(n.initialGossipKey()))
```
mreiferson: does this always need to be the *last* key - it seems like after the first loop, it won't remove anything, right?

mreiferson:
wonder if we've got :green_heart:

mreiferson:
seems like a pretty reliable race condition eh

mreiferson:
we're going to re-open this PR against the `dev` branch

PR #527, opened 2015-01-21
nsq_stat: deprecate --status-every in favor of --interval

mreiferson:
see #524 for original discussion


mreiferson:
RFR @jehiah

PR #522, opened 2015-01-12
nsqadmin: failed to delete ephemeral topic

mreiferson:
So I have created three ephemeral topics :

screenshot from 2015-01-08 15 11 55

Then just clicked in one of them: 

screenshot from 2015-01-08 15 12 27

Then press to delete that topic:

screenshot from 2015-01-08 15 12 59

but surprisingly the topic is still there:

screenshot from 2015-01-08 15 17 01

Also one more thing I saw that is even after shutdown completely all nsq services and restarting it, I could still see those three topics.

PR #529, opened 2015-01-23
nsqadmin: more ephemeral topic deletion issues

mreiferson:
another missed URL encoding...

ping #522 cc @PolyviosT


mreiferson:
RFR @jehiah

PR #530, opened 2015-01-24
nsqd.cfg.example: to change the worker_id it should just be id

jnewmano:
The configuration file option for worker_id should just be id


mreiferson:
Thanks!

Issue #528, opened 2015-01-22
dist: Docker image tag for v0.3.1 is v0.3.0 binary


MDrollette:
```
$ docker run --rm nsqio/nsqd:v0.3.1 -version
nsqd v0.3.0 (built w/go1.3.1)
```

mreiferson:
Wonderful, I'll try and attempt to figure out how on earth that happened...

Thanks for the report @MDrollette!

mreiferson:
hmmm - I think this has to do with the fact that we have this base docker image `nsqio/nsq` that actually pulls in the repo.

Since the app-specific docker images include `latest` [here](https://github.com/bitly/nsq/blob/master/apps/nsqd/Dockerfile#L1), they pickup whatever the base image last built, not the tag they were given.

This means we need to rethink and restructure how these are setup.

I'm honestly not sure what the best approach is here.

/cc @paddyforan and @rexposadas, they might have some ideas

MDrollette:
On a related note, it would be great if your build/release process created a more streamlined image for the apps. For example:

```Dockerfile
FROM progrium/busybox

ADD ./gopath/src/github.com/bitly/nsq/apps/nsqd/nsqd nsqd

VOLUME ["/data"]

EXPOSE 4150
EXPOSE 4151

ENTRYPOINT ["/nsqd", "--data-path=/data"]
```

And building with these steps:
```
$ docker run --rm --entrypoint /bin/sh nsqio/nsqd:latest -c 'tar cf - /gopath/src/github.com/bitly/nsq/apps/nsqd/nsqd' | tar xf -
$ docker build --rm -t nsqio/nsqd .
```

The resulting nsqio/nsqd image is 14mb instead of the current 726mb.

paddyforan:
> On a related note, it would be great if your build/release process created a more streamlined image for the apps.

It's something I've been looking into, but this (and the versioning of the files) is a problem I kind of just delayed. Looks like it's time to deal with it now.

@mreiferson you mind if I jump on IRC and hash out some ideas with you at some point? Most of my questions revolve around cutting a release more than docker-specific items.

mreiferson:
@paddyforan yes please :)

paddyforan:
From IRC:

```
paddyforan: basically, it revolves around ditching the auto-building that we’re doing right now, and instead doing manual builds
[2:54pm] paddyforan: the Dockerfiles are all hardcoded to use the “:unstable” tag, so that running `docker build` in the git source works as expected
[...]

PR #534, opened 2015-01-28
dist: single versioned Dockerfile approach

paddyforan:
Now we're trying to build the smallest container possible, and only ship
one container, instead of one container per app.

The strategy is to create a `docker` target in `dist`, and add all the
binaries from the linux target to that folder. The Dockerfile is based
off of `scratch`, and adds the `docker` target in `dist` at the root of
the filesystem. Finally, we add a `ca-certificates.crt` file (pulled
from CoreOS) so that SSL will continue to work.

This gives us a Docker image with about 5MB of overhead over the
binaries, compared to our previous strategy's ~700MB. It also means that
the versions will be easier to maintain and control.

The resulting images can be used as follows:

`docker run nsqio/nsq /nsqd --version`

Where `nsqd` is the binary you want to run. Unfortunately, I believe `/`
is required.

This is the result of our discussion in #528.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,5 @@
-FROM google/golang:latest
+FROM scratch
 
-RUN curl -s https://raw.githubusercontent.com/pote/gpm/v1.2.3/bin/gpm > /usr/local/bin/gpm
-RUN chmod +x /usr/local/bin/gpm
+ADD dist/docker/bin/ /
 
-ADD . $GOPATH/src/github.com/bitly/nsq
-RUN cd $GOPATH/src/github.com/bitly/nsq && gpm install
-RUN go get github.com/bitly/nsq/...
-
-RUN cd $GOPATH/src/github.com/bitly/nsq && ./test.sh
+VOLUME /etc/ssl/certs
--- a/dist.sh
+++ b/dist.sh
@@ -7,6 +7,8 @@ set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 mkdir -p $DIR/dist
+rm -rf $DIR/dist/docker
+mkdir -p $DIR/dist/docker
 
 mkdir -p $DIR/.godeps
 export GOPATH=$DIR/.godeps:$GOPATH
@@ -27,8 +29,10 @@ for os in linux darwin; do
     GOOS=$os GOARCH=$arch CGO_ENABLED=0 make
     make DESTDIR=$BUILD/$TARGET PREFIX= install
     pushd $BUILD
+    if [ "$os" == 'linux' ]; then cp -r $BUILD/$TARGET/bin $DIR/dist/docker/; fi
     tar czvf $TARGET.tar.gz $TARGET
     mv $TARGET.tar.gz $DIR/dist
     popd
     make clean
 done
+docker build -t nsqio/nsq:v$version .
--- a/apps/nsqd/Dockerfile
+++ b/apps/nsqd/Dockerfile
@@ -1,8 +1,0 @@
-FROM nsqio/deprecated_nsq_base:latest
-
-RUN cd /gopath/src/github.com/bitly/nsq/apps/nsqd && go build .
-VOLUME ["/data"]
-EXPOSE 4150
-EXPOSE 4151
-
-ENTRYPOINT ["/gopath/src/github.com/bitly/nsq/apps/nsqd/nsqd", "--data-path=/data"]
--- a/apps/nsqd/README.md
+++ b/apps/nsqd/README.md
@@ -3,20 +3,3 @@
 `nsqd` is the daemon that receives, queues, and delivers messages to clients.
 
 Read the [docs](http://nsq.io/components/nsqd.html).
-
-### Docker
-
-1. Pull down this image:
-
-        docker pull nsqio/nsqd
-
-2. Run the image:
-
-        docker run --name nsqd -p 4150:4150 -p 4151:4151
-            nsqio/nsqd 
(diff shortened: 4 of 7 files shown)


MDrollette:
I think this is great for packaging a release. The only downside I see is that it's not leveraging Docker for the build process. It makes assumptions about the builder's environment (go installed, cross-compile enabled, gpm installed). That may be fine, but sticking those dependancies in a Dockerfile and doing the build there, and then exporting just the binaries to this `scratch` image might simplify that. Just a thought.

mreiferson:
@MDrollette - it's something we briefly touched on in the IRC transcript pasted in #528...

*I* don't have an issue with that, and it sounds like something we can easily improve over time.

Now that this has materialized, my biggest concern is communicating all of these changes to the all the folks relying on the official `nsqd`, `nsqlookupd`, and `nsqadmin` trusted builds, now that everything is in a single image.

dist.sh
```diff
@@ -7,6 +7,7 @@ set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 mkdir -p $DIR/dist
+mkdir -p $DIR/dist/docker
```
mreiferson: probably need to clear this directory first, too (for example if we ever removed a binary)

dist.sh
```diff
@@ -27,8 +28,10 @@ for os in linux darwin; do
     GOOS=$os GOARCH=$arch CGO_ENABLED=0 make
     make DESTDIR=$BUILD/$TARGET PREFIX= install
     pushd $BUILD
+    if [ "${os}" == 'linux' ]; then cp -r $BUILD/$TARGET/bin $DIR/dist/docker/; fi
```
mreiferson: don't strictly need the `{` and `}`

Dockerfile
```diff
@@ -1,10 +1,4 @@
...
-RUN curl -s https://raw.githubusercontent.com/pote/gpm/v1.2.3/bin/gpm > /usr/local/bin/gpm
-RUN chmod +x /usr/local/bin/gpm
-
-ADD . $GOPATH/src/github.com/bitly/nsq
-RUN cd $GOPATH/src/github.com/bitly/nsq && gpm install
-RUN go get github.com/bitly/nsq/...
-
-RUN cd $GOPATH/src/github.com/bitly/nsq && ./test.sh
+ADD dist/docker/bin/ /
+ADD ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
mreiferson: lets put this file in `/contrib`

paddyforan:
> That may be fine, but sticking those dependancies in a Dockerfile and doing the build there, and then exporting just the binaries to this scratch image might simplify that. Just a thought.

I'll defer to @mreiferson on whether that's how he wants to handle the build or not. I'd vote that's a separate issue, however, as where we build the binaries is really immaterial as far as this PR is concerned.

The only person it really affects (AFAIK) is the person making the releases, so I have absolutely 0 opinion on it.

> Now that this has materialized, my biggest concern is communicating all of these changes to the all the folks relying on the official nsqd, nsqlookupd, and nsqadmin trusted builds, now that everything is in a single image.

Perhaps a mailing list announcement? A deprecation timeline? I'm not sure how to handle this. Theoretically, the `nsqio/nsqlookupd`, `nsqio/nsqd`, and `nsqio/nsqadmin` builds could all coexist with this... it's the `nsqio/nsq` image that is problematic. We could rename that to `nsqio/legacy` or something and update `nsqio/nsqd` et al, but anyone that's pulling `nsqio/nsq` by name will still get bitten.

paddyforan:
(I'll address the code updates, as well.)

mreiferson:
> Perhaps a mailing list announcement? A deprecation timeline? I'm not sure how to handle this. Theoretically, the nsqio/nsqlookupd, nsqio/nsqd, and nsqio/nsqadmin builds could all coexist with this... it's the nsqio/nsq image that is problematic. We could rename that to nsqio/legacy or something and update nsqio/nsqd et al, but anyone that's pulling nsqio/nsq by name will still get bitten.

I think these are all good ideas, particularly the deprecation timeline.  If we can keep the old images up for a bit, ideally with some sort of notification that you're using something that's going away, that would be great.

Related, this reminded me that we need a separate PR to update the docker docs page in the `gh-pages` branch for nsq.io.

paddyforan:
> I think these are all good ideas, particularly the deprecation timeline. If we can keep the old images up for a bit, ideally with some sort of notification that you're using something that's going away, that would be great.

Let me do a quick test of this to confirm. I'm at work at the moment, but if I get time tonight, I'll do it then. Or tomorrow. _Soon_.

A notification that you're using something deprecated is trickier. I have a couple ideas, with varying degrees of visibility. I'll work on testing them out, too. They may involve wrapping the binaries in shell scripts that log to stdout the deprecation warning.

> Related, this reminded me that we need a separate PR to update the docker docs page in the gh-pages branch for nsq.io.

Awww, crapskies. I had forgotten about that. I'll get another PR in for that as well shortly.

mreiferson:
@paddyforan what's actually left to do here beyond docs?  I'd like to stamp a new release this weekend and I'm curious if this new process can be a part of it?

paddyforan:
@mreiferson let's make it happen. I'm going to run a couple quick tests and make the minor changes you requested above. I'll be on IRC while I do this, if you want to hit me up and see how things are going.

Sorry, I'm the absolute worst at following up here.

paddyforan:
Woo, hit the wrong button there, sorry.

mreiferson:
:+1: :heart_eyes:

PR #538, opened 2015-02-06
nsqd: orphaned ephemeral channels

mreiferson:
If an `nsqd` cleanly exits while there are ephemeral channels with clients connected it will persist the existence of those channels to its metadata file on disk.

Upon startup, it will re-create those ephemeral channels.

If the names used for the ephemeral channels are random and not re-used upon client reconnection they are now effectively orphaned and will never be reconnected to and thus never cleaned up (something handled explicitly by disconnection of the last client).

Proposed fix is to not persist ephemeral topics and channels in the metadata file.

cc @jehiah @adamsathailo

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -413,6 +413,11 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 		if len(n.lookupPeers) > 0 {
 			channelNames, _ := lookupd.GetLookupdTopicChannels(t.name, n.lookupHttpAddrs())
 			for _, channelName := range channelNames {
+				if strings.HasSuffix(channelName, "#ephemeral") {
+					// we don't want to pre-create ephemeral channels
+					// because there isn't a client connected
+					continue
+				}
 				t.getOrCreateChannel(channelName)
 			}
 		}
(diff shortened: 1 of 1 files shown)


mreiferson:
RFR @jehiah 

It ended up being slightly different than I originally anticipated...

We were already skipping ephemeral channels (not topics) in our metadata file BUT when an `nsqd` starts up and creates topics, it queries lookupd and could receive ephemeral channels to *pre-create*, which isn't necessary and should be based entirely on a client connecting.

nsqd/nsqd.go
```diff
@@ -310,12 +313,13 @@ func (n *NSQD) PersistMetadata() error {
 		topic.Lock()
 		for _, channel := range topic.channelMap {
 			channel.Lock()
-			if !channel.ephemeral {
-				channelData := make(map[string]interface{})
-				channelData["name"] = channel.name
-				channelData["paused"] = channel.IsPaused()
-				channels = append(channels, channelData)
+			if channel.ephemeral {
+				continue
```
jehiah: `channel.Unlock()`?

jehiah:
LGTM aside from that comment

nsqd/nsqd.go
mreiferson: good catch, fixed and squashed in

twmb:
What happens now for a non-ephemeral topic on an ephemeral channel?

PR #539, opened 2015-02-07
Set apps to use deprecated base Docker image.

paddyforan:
In the pursuit of taking another swing at the NSQ Docker strategy, we're
redoing the nsqio/nsq image. Trying to keep the app builds working,
however, required us to create a deprecated_nsq_base image that they
could use as a source layer.


mreiferson:
:+1:

PR #540, opened 2015-02-07
dockerfile: create /data dir

paddyforan:
nsqd uses a /data dir, so let's make that available as a volume.


mreiferson:
thanks :+1:

PR #544, opened 2015-02-19
nsq_to_http: accept any 2xx response

mikedewar:
previously, nsq_to_http would fail on a 201, which is the response one receives when posting to elasticsearch.


mreiferson:
hahaha, works for me - want to squash this down @mikedewar ?

mikedewar:
squashing commits! that brings me back to the old days...

mreiferson:
thanks!

Issue #543, opened 2015-02-18
docker: scratch vs. busybox


saidimu:
Unlike the [deprecated images] (https://registry.hub.docker.com/u/nsqio/nsqd/), the official image [`nsqio/nsq`] (https://registry.hub.docker.com/u/nsqio/nsq/) isn't an Automated Build.

mreiferson:
Hi @saidimu - can you elaborate?  This was an intentional design decision because the old way we were doing things was broken.

The other benefit with the new approach is that the images are an order of magnitude smaller.

See https://github.com/bitly/nsq/pull/534 and relates issues it links to for more context.

mreiferson:
cc @paddyforan

saidimu:
The dramatically smaller images are super awesome, even with the loss of auto-discovery of `nsqd` host ips (previously I relied on `hostname --ip` to auto set the `--broadcast-address` flag). For now I'll stick to the deprecated images until I figure out another approach.

The deprecated images were linked to the Github repo so users could easily inspect the [Dockerfile](https://registry.hub.docker.com/u/nsqio/nsqd/dockerfile/) and be certain of the source of the image. Additional [build details](https://registry.hub.docker.com/u/nsqio/nsqd/builds_history/41603/) were also displayed; this was useful when deciding whether to pull updated images.

All that meta-data is lost with the new images because they aren't setup to automatically build on Github repo pushes.

AFAIK setting up an [automated build] (https://docs.docker.com/docker-hub/builds/) won't interfere with the new images.

paddyforan:
@saidimu couple of issues here:

1. No auto-discovery. This is a really annoying loss, and something I'm very interested in fixing. The complication is that all the binaries are in the same image. An obvious solution is to just expose the ports for all the binaries on the image, which would (I think) fix the auto discovery problem.
2. The loss of metadata in Docker Hub can be offset by updating the description on Docker Hub, can it not? Or am I not understanding? The build details you link to are just the timestamps and resulting image IDs of the builds. I'm struggling to find a use case for this information, considering (as far as I know) @mreiferson does not make a practice of updating old versions after they're released, opting instead to increment the version number, which would result in a new timestamped release, with a changelog, and links to the commits that changed between that release and the previous. That seems much more useful, right?
[...]

PR #545, opened 2015-02-20
Enable shell expansion.

paddyforan:
We're changing from scratch to busybox as a base image, after discussion
in #543. This gives us some free shell expansion.

I also added a step at the end of dist.sh that would take care of
tagging the image as the "latest" build, as long as there's not a hyphen
in the version number (so our "latest" build will never be an alpha or
beta, as I assume we want stability by default).

I also listed the ports used in the Dockerfile, so we can take advantage
of container linking.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
-FROM scratch
+FROM busybox
 
 ADD dist/docker/bin/ /
+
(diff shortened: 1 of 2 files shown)


mreiferson:
thanks @paddyforan

saidimu:
Good stuff @paddyforan

PR #548, opened 2015-02-24
exhaust request body

davidbirdsong:
Hi, we're using nsq_to_http as a source for dark traffic testing--it's great!

Would you guys be open to consuming the request body? It adds the overhead of the round trips incurred by body bytes, but otherwise causes all types of transport errors on the other side should the endpoint expect to return a body.

This doesn't change functionality at all (I think.)

### DIFF
--- a/apps/nsq_to_http/nsq_to_http.go
+++ b/apps/nsq_to_http/nsq_to_http.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"log"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


wtolson:
Might consider using [`ioutil.Discard`](http://golang.org/pkg/io/ioutil/#Discard). Not sure opening `/dev/null` would work on some OS's plus its a little cleaner and guaranteed not to return an error.

davidbirdsong:
@wtolson good call. I was searching around in ioutil, but didn't see ```iotul.Discard```.

Patched

mreiferson:
Makes sense, thanks!

PR #552, opened 2015-02-25
Lint style changes

cespare:
These are many simple style changes to match standard Go code style. All of these were flagged by golint.

There are still many warnings, but most are related to comments on exported types.

### DIFF
--- a/util/binary_version.go
+++ b/util/binary_version.go
@@ -5,8 +5,8 @@ import (
 	"runtime"
 )
 
-const BINARY_VERSION = "0.3.3-alpha"
+const BinaryVersion = "0.3.3-alpha"
 
 func Version(app string) string {
-	return fmt.Sprintf("%s v%s (built w/%s)", app, BINARY_VERSION, runtime.Version())
+	return fmt.Sprintf("%s v%s (built w/%s)", app, BinaryVersion, runtime.Version())
 }
--- a/util/lookupd/lookupd.go
+++ b/util/lookupd/lookupd.go
@@ -172,8 +172,8 @@ func GetLookupdProducers(lookupdHTTPAddrs []string) ([]*Producer, error) {
 					p = &Producer{
 						Hostname:         hostname,
 						BroadcastAddress: broadcastAddress,
-						TcpPort:          tcpPort,
-						HttpPort:         httpPort,
+						TCPPort:          tcpPort,
+						HTTPPort:         httpPort,
 						Version:          version,
 						VersionObj:       versionObj,
 						Topics:           topics,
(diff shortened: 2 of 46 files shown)


util/lookupd/lookupd.go
```diff
@@ -243,9 +243,11 @@ func GetLookupdTopicProducers(topic string, lookupdHTTPAddrs []string) ([]string
 // GetNSQDTopics returns a []string containing all the topics
 // produced by the given nsqd
 func GetNSQDTopics(nsqdHTTPAddrs []string) ([]string, error) {
-	topics := make([]string, 0)
-	var lock sync.Mutex
-	var wg sync.WaitGroup
+	var (
```
mreiferson: Does the linter really complain about this one or is this just personal preference?

Outside the global scope I just find this style distracting.

util/lookupd/lookupd.go
```diff
@@ -173,7 +173,7 @@ func GetLookupdProducers(lookupdHTTPAddrs []string) ([]*Producer, error) {
 						Hostname:         hostname,
 						BroadcastAddress: broadcastAddress,
 						TcpPort:          tcpPort,
```
mreiferson: `TCPPort`?

util/lookupd/lookupd.go
cespare: It's just the `make([]string, 0)`. We could just leave as

```
var topics []string
var lock sync.Mutex
var wg sync.WaitGroup
```

I changed to the var block because I prefer it, but I'm happy to leave the repeated `var` declarations if you like.

mreiferson: yes, please, thanks!

mreiferson:
this all LGTM after the minor comments above - thanks for doing this @cespare!

util/lookupd/lookupd.go
cespare: Not sure why it's distracting, though. Having less `var` keywords and alignment makes it clearer to my eyes.

mreiferson: Certainly not objective, also feels like too much code to change in order to be consistent with this, `go-nsq`, etc.

nsqadmin/templates/bindata.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
```
mreiferson: Maybe we should exclude this file since it's auto-generated by `go-bindata`?

util/lookupd/lookupd.go
cespare: Good point. Lint doesn't have that baked in. See golang/lint#110

nsqadmin/templates/bindata.go
cespare: Yes, my mistake. Good catch.

cespare:
OK done.

mreiferson:
Awesome, waiting for the green light - I can separately take a pass on the noise related to lack of comments.

util/binary_version.go
```diff
@@ -5,8 +5,8 @@ import (
 	"runtime"
 )
 
-const BINARY_VERSION = "0.3.3-alpha"
+const BinaryVersion = "0.3.3-alpha"
```
mreiferson: `dist.sh` needs to be patched [here](https://github.com/bitly/nsq/blob/master/dist.sh#L31)

cespare: Fixed.

PR #554, opened 2015-02-26
move utility packages under internal

mreiferson:
to clearly communicate that these are internal packages
and abide by the convention laid out in https://golang.org/doc/go1.4#internalpackages


mreiferson:
RFR @jehiah 

I figured that since we just landed some other hard-to-rebase-against changes might as well bite the bullet on this one.

I'd like to do some more re-organization of the random things we now have in `internal/util`, but we can do that separately.

cc @cespare @stephens2424

jehiah:
agreed. we need to kill the `util` namespace.

PR #555, opened 2015-02-26
internal pkg reorg

mreiferson:
This just unravels the `util` cruft into some reasonably sane internal packages


mreiferson:
ready @jehiah

jehiah:
nice. want to squash here?

It occurred to me, if you managed to do these renames with the [gorename](https://godoc.org/golang.org/x/tools/cmd/gorename) and [gomvpkg](https://godoc.org/golang.org/x/tools/cmd/gomvpkg) tools you might have been able to apply them easier to the outstanding dev branch. (disclaimer: i've never used either of those tools before)

mreiferson:
yea, I kept thinking "there's got to be a better way", shrug

:hammer:

PR #556, opened 2015-03-11
test.sh: fix find flag -depth to -maxdepth

twmb:
-depth specifies depth first search for processing directories.
-maxdepth specifies how deep in directories to go.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -33,7 +33,7 @@ go test -timeout 60s ./...
 GOMAXPROCS=4 go test -timeout 60s -race ./...
 
 # no tests, but a build is something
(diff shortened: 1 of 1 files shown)


ploxiln:
It's probably worth mentioning that this did the right thing for the stock mac os x `find` utility. For `-depth`, it evaluates to true for only that depth, and not for lesser depths (depth 0 in this case) as well, so it doesn't print the "apps" dir itself.

Anyway, the "grep package main" test limits the results to only the correct directories, and `-maxdepth` works for both gnu findutils and os x (probably freebsd) find, so :+1:

mreiferson:
Cool thanks

PR #557, opened 2015-03-11
nsqd: fix -tls-required=tcp-https with -tls-client-auth-policy

twmb:
Previously, specifying -tls-client-auth-policy would set -tls-required to
TLSRequired, overriding TLSRequiredExceptHTTP if it was specified.

Now, specifying the policy only sets TLSRequired if the current option is
TLSNotRequired.


mreiferson:
Nice, thanks!

Issue #521, opened 2015-01-08
nsqd: client error logged even when cleanly closed


detailyang:
nsqd always print log as follow on our production environment:
```
[nsqd] 2015/01/08 22:49:51.130538 TCP: new client(10.6.28.53:22505)
[nsqd] 2015/01/08 22:49:51.130847 CLIENT(10.6.28.53:22505): desired protocol magic '  V2'
[nsqd] 2015/01/08 22:49:51.137175 PROTOCOL(V2): [10.6.28.53:22505] exiting ioloop
[nsqd] 2015/01/08 22:49:51.137273 ERROR: client(10.6.28.53:22505) - failed to read command - EOF
[nsqd] 2015/01/08 22:49:51.137310 PROTOCOL(V2): [10.6.28.53:22505] exiting messagePump
```
my environment like this:
```
CentOS release 6.6 (Final)                                                                                                 
Kernel \r on an \m                                                                                                
2.6.32-504.3.3.el6.x86_64
nsqd v0.3.0 (built w/go1.3.3)
```
and my nsq client used [nsqphp](https://github.com/davegardnerisme/nsqphp)

so, Any idea?

mreiferson:
Can you paste code?

You can turn `nsqd` debug mode on to see what the client is sending, too.

P.S. This issue *probably* belongs on the `nsqphp` repo, though.

detailyang:
hi, i found this error reported by func (*Reader) ReadSlice
```
func (*Reader) ReadSlice

func (b *Reader) ReadSlice(delim byte) (line []byte, err error)
ReadSlice reads until the first occurrence of delim in the input, returning a slice pointing at the bytes in the buffer. The bytes stop being valid at the next read. If ReadSlice encounters an error before finding a delimiter, it returns all the data in the buffer and the error itself (often io.EOF). ReadSlice fails with error ErrBufferFull if the buffer fills without a delim. Because the data returned from ReadSlice will be overwritten by the next I/O operation, most clients should use ReadBytes or ReadString instead. ReadSlice returns err != nil if and only if line does not end in delim.
```

```
So it means there are not '\n' in Reader
line, err = client.Reader.ReadSlice('\n')
		if err != nil {
			err = fmt.Errorf("failed to read command - %s", err)
			break
```

I'm not familiar with Go:(
And finally,  how to turn nsqd debug mode? i cannot found nsqd --help with debug, only verbose=true.

mreiferson:
[...]

PR #560, opened 2015-03-23
nsqd: do not print EOF when closing cleanly

twmb:
When nsqd shuts down, protocol_v2 responds to CLS with CLOSE_WAIT and then the
IOLoop waits for another command with ReadSlice on the buffered client
connection. ReadSlice returns EOF because the client closes the connection;
this error is returned from IOLoop to the tcp handler and printed.

Because the presence of EOF is expected when the client is in stateClosing, the
error should be ignored.

### DIFF
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -62,7 +62,11 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 		// ie. the returned slice is only valid until the next call to it
 		line, err = client.Reader.ReadSlice('\n')
 		if err != nil {
-			err = fmt.Errorf("failed to read command - %s", err)
+			if atomic.LoadInt32(&client.State) == stateClosing {
+				err = nil
+			} else {
+				err = fmt.Errorf("failed to read command - %s", err)
+			}
 			break
 		}
 


nsqd/protocol_v2.go
```diff
@@ -113,6 +113,9 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 	if client.Channel != nil {
 		client.Channel.RemoveClient(client.ID)
 	}
+	if err == io.EOF && atomic.LoadInt32(&client.State) == stateClosing {
```
mreiferson: This is exactly what I had in mind, but I'm not sure `io.EOF` is the _only_ error possible in this state - if we're in `stateClosing` I think all errors should be ignored, right?

mreiferson:
fixes #521

nsqd/protocol_v2.go
twmb: Ignoring all errors would mask if a connection issued CLS followed by FIN or other messages (which error with `cannot XYZ in current state`). Also, it looks like PUB/MPUB can still be issued while closing.

mreiferson: If we move this code into `readLoop` it can more explicitly ignore all errors related to the _socket_ after a `CLS` exchange.

Thoughts?

twmb: Good point, changed.

nsqd/protocol_v2.go
```diff
@@ -62,7 +62,11 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 		// ie. the returned slice is only valid until the next call to it
 		line, err = client.Reader.ReadSlice('\n')
 		if err != nil {
-			err = fmt.Errorf("failed to read command - %s", err)
+			if err == io.EOF && atomic.LoadInt32(&client.State) == stateClosing {
```
mreiferson: cool, so _now_ we can just do this for any error, right?

twmb: I'm not so sure that ErrBufFull should be ignored, but since we're closing I guess it really *doesn't* matter.

mreiferson: right, I think the idea is to clean up logging here - we've already done the `CLS` exchange at this point and I think any read-based error is just noise.

mreiferson:
:+1:

mreiferson:
Thanks @twmb

PR #561, opened 2015-03-23
quantile: nil deref

mreiferson:
See commit comments.  `nsqadmin` and `nsq_stat` are the only bundled
apps that end up using this code path and currently panic.

cc @cespare

RFR @jehiah

### DIFF
--- a/internal/quantile/aggregate.go
+++ b/internal/quantile/aggregate.go
@@ -74,36 +74,32 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
 	return e.Percentiles[i]["percentile"] > e.Percentiles[j]["percentile"]
 }
 
-func (e *E2eProcessingLatencyAggregate) Add(e2 *E2eProcessingLatencyAggregate, N int) *E2eProcessingLatencyAggregate {
-	if e == nil {
-		*e = *e2
-	} else {
-		p := e.Percentiles
-		e.Count += e2.Count
-		for _, value := range e2.Percentiles {
-			i := -1
-			for j, v := range p {
-				if value["quantile"] == v["quantile"] {
-					i = j
-					break
-				}
+// Add merges e2 into e by averaging the percentiles
+func (e *E2eProcessingLatencyAggregate) Add(e2 *E2eProcessingLatencyAggregate) {
+	e.Addr = "*"
+	p := e.Percentiles
+	e.Count += e2.Count
+	for _, value := range e2.Percentiles {
+		i := -1
+		for j, v := range p {
+			if value["quantile"] == v["quantile"] {
+				i = j
+				break
 			}
-			if i == -1 {
-				i = len(p)
-				e.Percentiles = append(p, make(map[string]float64))
-				p = e.Percentiles
-				p[i]["quantile"] = value["quantile"]
-			}
-			p[i]["max"] = math.Max(value["max"], p[i]["max"])
-			p[i]["min"] = math.Min(value["max"], p[i]["max"])
-
-			p[i]["count"] += value["count"]
-			delta := value["average"] - p[i]["average"]
-			R := delta * value["count"] / p[i]["count"]
-			p[i]["average"] = p[i]["average"] + R
 		}
+		if i == -1 {
+			i = len(p)
+			e.Percentiles = append(p, make(map[string]float64))
+			p = e.Percentiles
+			p[i]["quantile"] = value["quantile"]
+		}
+		p[i]["max"] = math.Max(value["max"], p[i]["max"])
+		p[i]["min"] = math.Min(value["max"], p[i]["max"])
+
+		p[i]["count"] += value["count"]
+		delta := value["average"] - p[i]["average"]
+		R := delta * value["count"] / p[i]["count"]
+		p[i]["average"] = p[i]["average"] + R
 	}
 	sort.Sort(e)
-	e.Addr = "*"
-	return e
 }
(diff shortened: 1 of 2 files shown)


internal/quantile/aggregate.go
```diff
@@ -76,7 +76,7 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
 
 func (e *E2eProcessingLatencyAggregate) Add(e2 *E2eProcessingLatencyAggregate, N int) *E2eProcessingLatencyAggregate {
 	if e == nil {
-		*e = *e2
+		e = &(*e2)
```
mreiferson: This copies and then assigns the (new) address of `e2`, which is eventually returned by `Add` and then assigned over the original caller.

Gimme a bit to document that this method is side-effect free (and actually make the other branch in this method act that way instead of overwriting the receiver).

mreiferson:
recommend reviewing via https://github.com/bitly/nsq/pull/561/files?w=1

internal/quantile/aggregate.go
```diff
@@ -74,36 +74,43 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
 	return e.Percentiles[i]["percentile"] > e.Percentiles[j]["percentile"]
 }
 
+// Add merges two E2eProcessingLatencyAggregate by averaging the percentiles
+//
+// It is side-effect free and does not modify the receiver or the parameter
```
cespare: As written this function does modify the receiver.

internal/quantile/aggregate.go
```diff
@@ -74,36 +74,43 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
...
-			}
-			if i == -1 {
-				i = len(p)
-				e.Percentiles = append(p, make(map[string]float64))
-				p = e.Percentiles
-				p[i]["quantile"] = value["quantile"]
-			}
-			p[i]["max"] = math.Max(value["max"], p[i]["max"])
-			p[i]["min"] = math.Min(value["max"], p[i]["max"])
+		e = &(*e)
```
cespare: This is a no-op.

mreiferson: do I need:

```go
e3 := *e
e = &e3
```

internal/quantile/aggregate.go
cespare: And actually, why is this desirable? Why not just make this `func (e *E2eProcessingLatencyAggregate) Add(e2 *E2eProcessingLatencyAggregate, N int)` and have it do an in-place update? From looking at the callers it seems like they're all doing `x = x.Add(y)` anyway.

internal/quantile/aggregate.go
```diff
@@ -74,36 +74,43 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
 	return e.Percentiles[i]["percentile"] > e.Percentiles[j]["percentile"]
 }
 
+// Add merges two E2eProcessingLatencyAggregate by averaging the percentiles
+//
+// It is side-effect free and does not modify the receiver or the parameter
 func (e *E2eProcessingLatencyAggregate) Add(e2 *E2eProcessingLatencyAggregate, N int) *E2eProcessingLatencyAggregate {
 	if e == nil {
-		*e = *e2
```
cespare: Sorry for this bug. Probably what I meant to do is

``` go
if e == nil {
  e = &E2eProcessingLatencyAggregate{}
  *e = *e2
} else {
  ...
```

But I also didn't realize that this function was not supposed to modify the receiver. It looked like a `math/big`-style API.

internal/quantile/aggregate.go
cespare: Yes

internal/quantile/aggregate.go
mreiferson: I didn't either, to be honest.

I'm just trying to be consistent, now.  Before, depending on the branch it took it would either modify the receiver or not.

internal/quantile/aggregate.go
mreiferson: I would be fine with this approach, too (as it would also now be internally consistent).

... and I thought this would be possible too, even if `x` is initially `nil`, and I might be missing something simple and obvious, but I couldn't seem to get it to work the way I would have expected.

cespare: No, Go is pass-by-value always so that's not possible.

My suggestion is to change the calling code so that the receiver is never nil. I'm happy to send over a counter-PR to fix this bug, if you like.

mreiferson: I'll update - thanks for the help!

mreiferson:
@cespare check that latest commit

internal/quantile/aggregate.go
```diff
@@ -74,36 +74,33 @@ func (e *E2eProcessingLatencyAggregate) Less(i, j int) bool {
...
-					break
-				}
-			}
-			if i == -1 {
-				i = len(p)
-				e.Percentiles = append(p, make(map[string]float64))
-				p = e.Percentiles
-				p[i]["quantile"] = value["quantile"]
+// Add merges the passed E2eProcessingLatencyAggregate into the receiver
+// by averaging the percentiles
```
cespare: How about

```
// Add merges e2 into e by averaging the percentiles.
```

cespare:
LGTM

internal/quantile/aggregate.go
mreiferson: sure

mreiferson:
ready @jehiah

PR #562, opened 2015-03-24
nsq_to_{nsq,http}: add epsilon-greedy hostpool mode

twmb:
go-hostpool's EpsilonGreedy chooses hosts to send to based on which host has
been most performant in the past, which is a good behavior to add.

The current hostpool option round-robins with backoff on failing hosts.

### DIFF
--- a/apps/nsq_to_http/nsq_to_http.go
+++ b/apps/nsq_to_http/nsq_to_http.go
@@ -40,7 +40,7 @@ var (
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 
 	numPublishers = flag.Int("n", 100, "number of concurrent publishers")
-	mode          = flag.String("mode", "round-robin", "the upstream request mode options: multicast, round-robin, hostpool")
+	mode          = flag.String("mode", "hostpool", "the upstream request mode options: multicast, round-robin, hostpool (default), epsilon-greedy")
 	sample        = flag.Float64("sample", 1.0, "% of messages to publish (float b/w 0 -> 1)")
 	httpTimeout   = flag.Duration("http-timeout", 20*time.Second, "timeout for HTTP connect/read/write (each)")
 	statusEvery   = flag.Int("status-every", 250, "the # of requests between logging status (per handler), 0 disables")
@@ -222,7 +222,7 @@ func main() {
 		selectedMode = ModeAll
 	case "round-robin":
 		selectedMode = ModeRoundRobin
-	case "hostpool":
+	case "hostpool", "epsilon-greedy":
 		selectedMode = ModeHostPool
 	}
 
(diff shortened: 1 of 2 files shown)


apps/nsq_to_http/nsq_to_http.go
```diff
@@ -40,7 +40,7 @@ var (
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 
 	numPublishers = flag.Int("n", 100, "number of concurrent publishers")
-	mode          = flag.String("mode", "round-robin", "the upstream request mode options: multicast, round-robin, hostpool")
+	mode          = flag.String("mode", "round-robin", "the upstream request mode options: multicast, round-robin, hostpool, epsilon-greedy")
```
mreiferson: switch default to `hostpool`?

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -40,7 +40,7 @@ var (
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 
 	statusEvery = flag.Int("status-every", 250, "the # of requests between logging status (per destination), 0 disables")
-	mode        = flag.String("mode", "round-robin", "the upstream request mode options: round-robin (default), hostpool")
+	mode        = flag.String("mode", "round-robin", "the upstream request mode options: round-robin (default), hostpool, epsilon-greedy")
```
mreiferson: same here?

mreiferson:
LGTM otherwise

apps/nsq_to_http/nsq_to_http.go
twmb: But that would be behavior change?

mreiferson: Yea, I mentioned this in IRC, but I think it's the _right_ behavior change longer term.  I think in the majority of cases this is what people want.

twmb:
I've switched the default for both to hostpool.

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -380,6 +380,9 @@ func main() {
 		perAddressStatus: perAddressStatus,
 		timermetrics:     timer_metrics.NewTimerMetrics(*statusEvery, "[aggregate]:"),
 	}
+	if *mode == "epsilon-greedy" {
+		handler.hostPool = hostpool.NewEpsilonGreedy(destNsqdTCPAddrs, 0, &hostpool.LinearEpsilonValueCalculator{})
+	}
```
mreiferson: actually, now that I'm looking at the surrounding code - I think this could be cleaner if we move this block above the instantiation of `PublishHandler`, pivot on `*mode`, assign a local `hostPool` variable, and reference that when instantiating.

twmb: always be pivoting

twmb: Changed.

mreiferson:
thanks

PR #565, opened 2015-03-25
Add dotabuff logo (gh-pages)

jcoene:
Billions of messages later, we're still in love with NSQ and would love to show it!


mreiferson:
THANKS :heart_eyes:

PR #575, opened 2015-04-23
nsqd: skip persist metadata while loading

xiaost:
:laughing: :laughing: :laughing: 
reasons:
1. the in-memory metadata is incomplete;
2. nsqd triggers N (where N >= # of topics + channels)
    times `PersistMetadata` call which is unnecessary;
3. nsqd will persist metadata after loading.

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -32,6 +32,11 @@ const (
 	TLSRequired
 )
 
+const (
+	flagHealthy = 1 << iota
+	flagLoading
+)
+
 type NSQD struct {
 	// 64bit atomic vars need to be first for proper alignment on 32bit platforms
 	clientIDSequence int64
@@ -40,8 +45,8 @@ type NSQD struct {
 
 	opts *nsqdOptions
 
-	healthMtx sync.RWMutex
-	healthy   int32
+	flag      int32
+	errMtx    sync.RWMutex
 	err       error
 	startTime time.Time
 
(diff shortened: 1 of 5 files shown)


nsqd/nsqd.go
```diff
@@ -502,12 +534,16 @@ exit:
...
 
 func (n *NSQD) Notify(v interface{}) {
+	persist := !n.getFlag(flagLoading)
 	n.waitGroup.Wrap(func() {
 		// by selecting on exitChan we guarantee that
 		// we do not block exit, see issue #123
 		select {
 		case <-n.exitChan:
 		case n.notifyChan <- v:
+			if !persist {
```
mreiferson: do we need this check here?

nsqd/nsqd.go
```diff
@@ -298,6 +324,12 @@ func (n *NSQD) LoadMetadata() {
 }
 
 func (n *NSQD) PersistMetadata() error {
+	if n.getFlag(flagLoading) {
```
mreiferson: since notification (and persisting) happens asynchronously, I think we need to ensure that we don't even _try_ to notify so there isn't any race - i.e. do this logic where the notification is sent rather than received.

mreiferson:
@xiaost thanks for this patch, really nice find!

Just a few comments above.

nsqd/nsqd.go
xiaost: yes, since `Notify` may called by `NewTopic` /  `NewChannel`  in the loading stage

nsqd/nsqd.go
xiaost: agree. I'm going to update the patch.

xiaost:
@mreiferson 

C1. `Load`move `PersistMetadata` out of topic / channel `doPause`, public pause api takes the responsibility of it
WHY:  I think `PersistMetadata` should be called in high level logic, so we can better control persistence

C2. remove flag checking from `LoadMetadata`
WHY:  with change of `doPause` and `Notify`, `LoadMetadata` is persistence-free

mreiferson:
thanks!

LGTM

PR #577, opened 2015-04-30
nsqd: questions for large number of topics/channels

mreiferson:
Background: We need to dispatch user messages to a specific topic, where every consumer only subscribe the message for a fixed range of user. Since nsqd doesn't support partitions by key, we created thousands of topics/channels, i.e. user_topic_[0-999], one topic for a range of users.

Let's suppose there will be 10K topics and channels, we want to know that whether nsqd will work well or not in this situations.

1. file number limit of process. This is easy to increase it.
2. memory consumption? After some trying, seems not so much memory consumption.
3. High cpu usage even if EMPTY workload. About 40% - 60% for one nsqd process, while no producer or consumer connect to it. `strace` tells there are lots of futex syscall.
    @xiaost helps to troubleshoot, it may be caused by the 100ms ticker for every channel, and maybe we need to make it configurable?
4. slow startup. @xiaost already submited a PR https://github.com/bitly/nsq/pull/575
5. Any other potential problems?

Thanks

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -59,6 +59,8 @@ type NSQD struct {
 	httpsListener net.Listener
 	tlsConfig     *tls.Config
 
+	poolSize int
+
 	idChan     chan MessageID
 	notifyChan chan interface{}
 	exitChan   chan int
@@ -245,6 +247,7 @@ func (n *NSQD) Main() {
 		http_api.Serve(n.httpListener, httpServer, n.opts.Logger, "HTTP")
 	})
 
+	n.waitGroup.Wrap(func() { n.queueScanLoop() })
 	n.waitGroup.Wrap(func() { n.idPump() })
 	n.waitGroup.Wrap(func() { n.lookupLoop() })
 	if n.opts.StatsdAddress != "" {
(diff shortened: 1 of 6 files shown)


mreiferson:
opened a PR here w/ the pool of workers for concurrency

RFR @allenlz @xiaost @jehiah

nsqd/nsqd.go
```diff
@@ -551,6 +554,139 @@ func (n *NSQD) Notify(v interface{}) {
...
+		}
+
+		numDirty := 0
+		for i := 0; i < num; i++ {
+			if <-responseCh {
+				numDirty++
+			}
+		}
+
+		if float64(numDirty)/float64(num) > n.opts.QueueGCDirtyPercent {
```
jehiah: What does this mean for a case where you have 1 hot channel (with a large deferred queue or something) and lots of quiet/idle channels. Should we loop on the same channels to drain them instead of picking different ones each loop?

That would mean that we will be actively draining priority queues (25% of them a time up to max QueueGCPoolMax) and only go back to the random selection if the one we were draining is empty. This would keep `QueueGCInterval` from being an accidental cap of how fast we can drain priority queues of QueueGCPoolMax per QueueGCInterval in that edge case.

nsqd/nsqd.go
```diff
@@ -551,6 +554,139 @@ func (n *NSQD) Notify(v interface{}) {
...
+			num = len(channels)
+		}
+
+		for _, i := range uniqRands(num, len(channels)) {
+			workCh <- channels[i]
+		}
+
+		numDirty := 0
+		for i := 0; i < num; i++ {
+			if <-responseCh {
```
mreiferson: making a note for myself:

This is insufficient in the case where we asynchronously get a signal from `n.exitChan` to close... there's definitely a case where this would deadlock.

We should probably select on `n.exitChan` here, too.

nsqd/nsqd.go
```diff
@@ -551,6 +554,139 @@ func (n *NSQD) Notify(v interface{}) {
...
+			continue
+		case <-n.exitChan:
+			goto exit
+		}
+
+	loop:
+		num := n.opts.QueueGCSelectionCount
+		if num > len(channels) {
+			num = len(channels)
+		}
```
mreiferson: another note for myself:

this doesn't have to be in the `loop` block.

nsqd/nsqd.go
mreiferson: Each iteration should "drain" the deferred queue in the sense that there was no more work to do _now_.  There might still be more in the queue that have not timed out.  Does this clarify anything? Do you think we still need to somehow prioritize busier queues?

jehiah: ahh. the loop inside `processInFlightQueue` is what i was missing. It works as i was suggesting then. (I mistakenly thought it only pop'd one message each time that channel got pushed into `workCh`)

√

nsqd/options.go
```diff
@@ -28,6 +28,12 @@ type nsqdOptions struct {
 	SyncEvery       int64         `flag:"sync-every"`
 	SyncTimeout     time.Duration `flag:"sync-timeout"`
 
+	QueueGCInterval        time.Duration
```
jehiah: I'm not a fan of `GC` here, but i don't have much of a better suggestion. Naming here feels complicated because this controls both the in-flight timeouts, and the deferred queue timeouts. If we were to only describe this in terms of Deferred or in-flight it might be easier?

`DeferredQueueCheckInterval`, `DeferredQueueScalingInterval` (i'm not sure we should even expose this one), `DeferredQueueMaxWorkers` (or max concurrent?), `DeferredQueueDirtyPercent`. 

The other naming pattern that comes to mind would be `RequeueCheckInterval`, ....

mreiferson: Agreed on coming up with better names.  Note that I explicitly chose not to expose them as flags or config file vars (I assumed we might want to iterate and improve these knobs).

jehiah: fair point; i'm ok w/ exposure in the config obj but not flags.

mreiferson:
@jehiah updated names, bit of re-org, added comments and fixed up the exit paths PTAL

jehiah:
LGTM. you ready for this to land?

mreiferson:
Yes, I'll :hammer:

nsqd/nsqd.go
mreiferson: this was resolved

mreiferson:
:fire: :facepunch:

allenlz:
Today, we met another problem about large number topics/channels. We found nslookupd consumes a lot of cpu. Please refer to https://github.com/bitly/nsq/pull/584

PR #553, opened 2015-05-01
Tests/benchmarks should clean up after themselves

mreiferson:
There are many tests that drop tempfiles directly into `os.TempDir()` and don't delete them. Instead, these tests should do this:

```
tempDir, err := ioutil.TempDir("", "nsq-some-desc-here")
if err != nil {
  t.Fatal(err)
}
defer os.RemoveAll(tempDir)
```

I discovered this by finding that `/tmp` has 21GB of nsq data data in it.

I'll try to send over a PR for this.

### DIFF
--- a/nsqd/channel_test.go
+++ b/nsqd/channel_test.go
@@ -1,6 +1,7 @@
 package nsqd
 
 import (
+	"os"
 	"strconv"
 	"testing"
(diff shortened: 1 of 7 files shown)


mreiferson:
RFR @jehiah @cespare

Issue #34, opened 2012-09-23
nsqd: disk-backed deferred queue


mreiferson:
the deferred queue is the only spot where there are no bounds to memory footprint.

we could write deferred messages to disk and keep the index in memory which would for most use cases completely eliminate this problem.

mreiferson:
closing this in favor of #510

PR #293, opened 2015-05-01
nsqd: support deferred PUB

mreiferson:
This came up on the mailing list: https://groups.google.com/forum/#!topic/nsq-users/HkoOMygaHK0 - just opening the issue here for others to weigh in and gauge interest.

This would allow you to `PUB` a message that goes directly to the deferred priority queue with the specified delay.

Also, It's complicated a bit by #34.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -1,6 +1,6 @@
 github.com/BurntSushi/toml              2dff11163ee667d51dcc066660925a92ce138deb
 github.com/bitly/go-hostpool            58b95b10d6ca26723a7f46017b348653b825a8d6
-github.com/bitly/go-nsq                 22a8bd48c443ec23bb559675b6df8284bbbdab29 # v1.0.4
+github.com/bitly/go-nsq                 4271cd1529a78175e327570894988cc2cb21228f # v1.0.5-alpha
 github.com/bitly/go-simplejson          fc395a5db941cf38922b1ccbc083640cd76fe4bc # v0.5.0-alpha
 github.com/bmizerany/perks/quantile     6cb9d9d729303ee2628580d9aec5db968da3a607
 github.com/mreiferson/go-options        2cf7eb1fdd83e2bb3375fef6fdadb04c3ad564da
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -243,19 +243,31 @@ func (s *httpServer) doPUB(req *http.Request) (interface{}, error) {
 		return nil, http_api.Err{500, "INTERNAL_ERROR"}
 	}
 	if int64(len(body)) == readMax {
-		s.ctx.nsqd.logf("ERROR: /put hit max message size")
 		return nil, http_api.Err{413, "MSG_TOO_BIG"}
 	}
 	if len(body) == 0 {
 		return nil, http_api.Err{400, "MSG_EMPTY"}
 	}
 
-	_, topic, err := s.getTopicFromQuery(req)
+	reqParams, topic, err := s.getTopicFromQuery(req)
 	if err != nil {
 		return nil, err
 	}
 
+	var deferred time.Duration
+	if ds, ok := reqParams["defer"]; ok {
+		di, err := strconv.ParseInt(ds[0], 10, 64)
+		if err != nil {
+			return nil, http_api.Err{400, "INVALID_DEFER"}
+		}
+		deferred = time.Duration(di) * time.Millisecond
+		if deferred < 0 || deferred > s.ctx.nsqd.opts.MaxReqTimeout {
+			return nil, http_api.Err{400, "INVALID_DEFER"}
+		}
+	}
+
 	msg := NewMessage(<-s.ctx.nsqd.idChan, body)
+	msg.deferred = deferred
 	err = topic.PutMessage(msg)
 	if err != nil {
 		return nil, http_api.Err{503, "EXITING"}
--- a/nsqd/http_test.go
+++ b/nsqd/http_test.go
@@ -154,6 +154,33 @@ func TestHTTPmputBinary(t *testing.T) {
 	equal(t, topic.Depth(), int64(5))
 }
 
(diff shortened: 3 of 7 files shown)


mreiferson:
Here goes nothing... thoughts on this approach?

mreiferson:
(the tests are failing because of a dependency on `go-nsq` changes, which is unfortunate)

jehiah:
actually comes out pretty cleanly. :+1:

mreiferson:
@jehiah added HTTP support.  Note, we're explicitly not exposing deferred pub for MPUB varieties.  These tests should pass.

jehiah:
:+1: thoughts on if/how this should be exposed in nsq_to_nsq?

mreiferson:
> thoughts on if/how this should be exposed in nsq_to_nsq?

Do we need to support it there?  Arguably, it's already got too many knobs.  Supporting this _at all_ was already a question, having to write your own consumer to use it when copying streams around is too much to ask? :grin:

jehiah:
Yeah i hear ya. I mean the obvious answer is it's the one spot where you would want to have a `--defer-msg=60s` type option.

Since this isn't a "reader option" it would need to be exposed directly, and we don't (yet?) support config files there so ... I'm in favor of adding there.

jayridge:
its like christmas boys

mreiferson:
we've been had

zygis:
How about MDPUB or DMPUB (Multi Deferred Publish)?

mreiferson:
@zygis I alluded to this in my comment above, but I intentionally didn't want to expose a means to defer en masse a large set of messages.  In the current implementation, `nsqd` keeps deferred messages _entirely_ in memory (i.e. they _do not_ overflow to disk).

zygis:
Well... 
1000xDPUB - **same result** with ~1000 network calls
1xMDPUB - **same result** with ~1 network call

When deferred duration is few seconds maybe this limitation makes sence, but if duration is minutes or even hours I think this limitation solves nothing.

mreiferson:
@zygis I realize that.  I'm concerned with the behaviors that we encourage here.  I can be convinced, it just didn't seem like the obvious use case.

jehiah:
@zygis if the client connection buffers the socket there is little network difference between those two different protocol messages. You can pipeline a thousand DPUB just like a MDPUB would effectively provide. (there is minimal difference in command overhead between the two approaches)

zygis:
@mreiferson I just do lot of writes to DB and via nsq notifying rest part of system do some computations. The main problem is that reads are done from slaves in other continent, and some times written data still not available, because of cross-continent replication lag. So DPUB is acceptable solution. But few weeks ago switched from PUB(client nsq.js) to MPUB(client go-nsq) because of networking problems on a servers... MDPUB could be perfect solution.

@jayridge not sure how it can be done with go-nsq client

jehiah:
@zygis What go-nsq needs to expose for you is the equivlant of [`Producer.DeferredPublishAsync`](https://github.com/bitly/go-nsq/blob/master/producer.go#L161). The `*Async` commands allow you to pipeline multiple requests for throughput.

zygis:
@jehiah Thanks, I will try this.

jonathannorris:
I've been interested in using nsq for a while, we have a potential use case coming up that requires queuing of millions of deferred messages for up to a day or so. I was wondering if there are any potential performance pitfalls with this, other then nsq storing millions of messages in memory (and not backing them up on disk). It's probably not the right solution for the problem, but I'm interested to hear from @jehiah @mreiferson on this.

ploxiln:
just a general purpose PSA: For scheduling functionality, you could use a database, with an "expires" field/column, with an index on that field/column. Have a process query the database once a minute or so for "expires < $now", process those results, then delete those rows/documents. You can limit the query to return, say, 100 results, and you can re-run the query immediately after processing&deleting if there were any results.

That's roughly what's done for a few parts of @jayridge's current project, and he grudgingly uses nsq as well anyway ;)

mreiferson:
@jonathannorris there aren't any other performance implications, there's just the durability question.  Whether it's the right solution to the problem depends on your requirements... i.e. can you afford to lose messages that are in memory?

mreiferson:
> That's roughly what's done for a few parts of @jayridge's current project, and he grudgingly uses nsq as well anyway ;)

I think that's a valid statement for _any_ technology @jayridge uses.

jayridge:
true that

captainblue2013:
So ? Can anyone tell me how to use DPUB ?
There is no concept of DPUB in  http://nsq.io/components/nsqd.html#post-pub

mreiferson:
@captainblue2013 we should obviously document this - just pass the `?defer=x` param to `/pub`, see https://github.com/nsqio/nsq/blob/master/nsqd/http.go#L226-L237

captainblue2013:
@mreiferson  I did it after read the source code

kgdev:
I found a potential bug, I switched our code from using PUB to DPUB recently, and the message got handled by the client within the specified period successfully, which was good. 
However, when I check the status in localhost:4171, the "Requeued" count for that channel keeps increasing (which is not the case for PUB).

Can someone take a look into it?

image

sakop:
@kgdev It is very easy to reproduce this bug, just exec this

``` curl -d "aDa" 'http://127.0.0.1:4151/pub?defer=1000&topic=aa'```
and you will see **Requeued** keep increasing while **Messages** remains 0

ploxiln:
This makes sense because nsqd treats it just like a message on which the consumer called "requeue(delay=...)". It could be improved - but for that you should definitely open a new separate issue.

(Also, nsqd was not designed as a scheduling service, and thus has a few other problems with this use case.)

kgdev:
Agree. I prefer not using "deferred" enqueue.

mreiferson:
@kgdev I believe this accounting issue was resolved in #805

PR #584, opened 2015-05-07
nsqlookupd: improved performance of registration db.

xiaost:
without '*' expression of topic & channel, we can fetch results from map directly.

it is another issue we found in #577, nsqlookupd eat up my CPUs :cry:

### DIFF
--- a/nsqlookupd/registration_db.go
+++ b/nsqlookupd/registration_db.go
@@ -113,9 +113,20 @@ func (r *RegistrationDB) RemoveRegistration(k Registration) {
 	delete(r.registrationMap, k)
 }
 
+func (r *RegistrationDB) needFilter(key string, subkey string) bool {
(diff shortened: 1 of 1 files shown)


nsqlookupd/registration_db.go
```diff
@@ -113,9 +113,20 @@ func (r *RegistrationDB) RemoveRegistration(k Registration) {
 	delete(r.registrationMap, k)
 }
 
+func (r *RegistrationDB) needFilter(category string, key string, subkey string) bool {
```
jehiah: you can drop `category` from this function parameter list as it's not used.

jehiah:
Nice improvements. Squash in that one small fix and we'll merge this in.

xiaost:
@jehiah :+1: fixed

PR #586, opened 2015-05-08
nsqd: fix valid range for --worker-id

mreiferson:
I did the math wrong in #429, it's actually 10-bit width not 12-bit :(

RFR @jehiah

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -83,8 +83,8 @@ func NewNSQD(opts *nsqdOptions) *NSQD {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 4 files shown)


twmb:
lgtm based off

https://github.com/bitly/nsq/blob/master/nsqd/guid.go#L19 and https://github.com/bitly/nsq/blob/master/nsqd/nsqd.go#L511

jehiah:
@mreiferson integration test [here](https://travis-ci.org/bitly/nsq/jobs/61839950) hit a data race; can you take a look at that first? (ok if we defer i just want to make sure we capture and understand if it's test specific or not).

mreiferson:
@jehiah should be fixed

PR #600, opened 2015-07-10
nsqd: enforce minimum msg size when reading from the diskqueue

twmb:
See bitly/nsq#598

### DIFF
--- a/nsqd/diskqueue.go
+++ b/nsqd/diskqueue.go
@@ -32,7 +32,9 @@ type diskQueue struct {
 	// instantiation time metadata
 	name            string
 	dataPath        string
-	maxBytesPerFile int64         // currently this cannot change once created
+	maxBytesPerFile int64 // currently this cannot change once created
+	minMsgSize      int32
+	maxMsgSize      int32
 	syncEvery       int64         // number of writes per fsync
 	syncTimeout     time.Duration // duration of time per fsync
 	exitFlag        int32
@@ -65,12 +67,15 @@ type diskQueue struct {
 // newDiskQueue instantiates a new instance of diskQueue, retrieving metadata
 // from the filesystem and starting the read ahead goroutine
 func newDiskQueue(name string, dataPath string, maxBytesPerFile int64,
+	minMsgSize int32, maxMsgSize int32,
 	syncEvery int64, syncTimeout time.Duration,
 	logger logger) BackendQueue {
 	d := diskQueue{
 		name:              name,
 		dataPath:          dataPath,
 		maxBytesPerFile:   maxBytesPerFile,
+		minMsgSize:        minMsgSize,
+		maxMsgSize:        maxMsgSize,
 		readChan:          make(chan []byte),
 		writeChan:         make(chan []byte),
 		writeResponseChan: make(chan error),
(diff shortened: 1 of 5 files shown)


nsqd/diskqueue.go
```diff
@@ -341,6 +353,10 @@ func (d *diskQueue) writeOne(data []byte) error {
 	d.writePos += totalBytes
 	atomic.AddInt64(&d.depth, 1)
 
+	// TODO: this is off by one, but we cannot fix this without also fixing
+	// readOne above. We cannot fix readOne above until we guarantee that
+	// every old diskqueue format has been fully processed, otherwise we
+	// will miss one message from old formatted files.
```
mreiferson: what do you mean by this comment?

nsqd/diskqueue.go
```diff
@@ -261,6 +261,14 @@ func (d *diskQueue) readOne() ([]byte, error) {
 		return nil, err
 	}
 
+	if msgSize < minMsgSize {
```
mreiferson: `DiskQueue` doesn't (and arguably shouldn't) know about "messages".  I think that `DiskQueue` should test for `msgSize > 0`, not `minMsgSize`.  The receiver of the read channel should enforce that the size is appropriate for a message, i.e. at least `minMsgSize`, so we can create an additional check there.

mreiferson:
nice work on this @twmb - thanks for adding all those comments to the tests!

nsqd/diskqueue.go
twmb: This is meant to rotate files when we reach maxBytesPerFile, but will only
rotate when we exceed that number. If max bytes is 1024 and we are at bytes
1024, we will write one more message before rotating files.

nsqd/diskqueue.go
twmb: If the receiver reads an invalid message, there is no way for the receiver
to invalidate the current backing file. This also means that any new read
from the current file cannot be trusted, as if the reader trusts the file
to only contain messages, and it doesn't, how can the reader know any more
of the current file is valid?

nsqd/diskqueue.go
mreiferson: Fair enough, however for all practical purposes I don't think this is something that needs to be addressed other than perhaps the documentation around the config param.

Can we remove this comment?

nsqd/diskqueue.go
mreiferson: I only think `DiskQueue` needs to protect itself from an undesirable `panic`, but perhaps we're just settling on the fact that we _do_ actually need to pass in a `min` and `max` in the `DiskQueue` constructor (so it can internally apply both).

That seems like the right happy medium to both our points and would be objectively better than the current implementation.  It would still not protect from actual message body corruption, however that would require more invasive changes (CRC).

nsqd/diskqueue.go
twmb: Sure thing

nsqd/diskqueue.go
```diff
@@ -261,6 +266,14 @@ func (d *diskQueue) readOne() ([]byte, error) {
 		return nil, err
 	}
 
+	if msgSize < d.minMsgSize || msgSize > d.maxMsgSize {
```
mreiferson: I don't know if `msgSize > d.maxMsgSize` means it's _corrupt_?  It should _skip_ it, but does it need to discard the whole segment?

twmb: msgSize could be > d.maxMsgSize because of either

a) a corrupt message, or
b) a person changed the flag of their maxMsgSize between nsqd restarts

If it's a corrupt message, then we'd skip forward in the file and then have no guarantee of what we're reading. If a person changed their maxMsgSize, they'd miss this message and be fine reading the next. I don't think it's too common to reduce MaxMsgSize between nsqd restarts other than when a person is trying to figure out parameters, and if they're fine with losing any large messages when tuning, I can't see why they wouldn't be fine losing an entire segment.

I favor discarding the file but I could go either way.

mreiferson: Alright, you've sold me - all of this falls under "reasonable behavior" without _actually_ knowing if it's corrupt.

PR #601, opened 2015-07-14
nsqd: ability to dynamically reconfigure nsqlookupd addresses

mreiferson:
I'd love it if NSQ could support DNS-based discovery for nsqlookupd instances. This would be a light layer on top of the already-proven architecture that would allow ops people to abstract away some of the configuration required to get NSQ working.

Basically, rather than hardcoding nsqlookupd addresses, I'd much rather just say "use all the A records returned by this DNS name". There's also an [RFC](http://www.ietf.org/rfc/rfc6763.txt) for using DNS for service discovery. This supports things like [Prometheus](http://prometheus.io) and makes deployment a snap on [Kubernetes](http://kubernetes.io). It basically makes it dead simple to use NSQ in a cluster environment, where you don't _know_ what IP address or machine any of your services will be available at, but know you can reach them through a DNS lookup.

[More docs](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/services.md#headless-services) on the relevant portion of Kubernetes.

Here's the desired user experience:

1. I set up a service (logical grouping) of pods (group of containers, think "application cookie cutter") and say I want 4 of them. Kubernetes automatically _does the right thing_ for me, setting them up, tearing them down, restarting them, moving them, whatever.
2. I configure nsqd to say "do a DNS lookup on nsqlookupd.cluster, and connect to every IP address returned as an A record. Also, every minute/five minutes/ten minutes, check again, and update your connections"
3. I configure my consumers to say "do a DNS lookup on nsqlookupd.cluster, and connect to every IP address returned as an A record. Also, every minute/five minutes/ten minutes, check again and update your connections"
4. Everything just works. Containers get restarted, machines get removed, machines get added, containers get moved, and everything just points to the right thing.

Most of this (the DNS querying part) can be achieved simply with wrappers/helpers. The big missing piece is some sort of runtime live-reloading/editing of configuration. Any of the following would work beautifully:

* A way to easily wrap nsqd inside another Go binary, and use a channel/function/whatever to say "reload configuration/here are new configuration values"
* A way to say "Here's a configuration file, read the nsqlookupd addresses out of it, and re-check whenever I HUP you" (I'm unsure whether one container can HUP a process in another. If not, this is less ideal, but still probably workable. It'll just take some noodling on how to make it happen. The unix socket approach would be more desirable, honestly)
* A TCP/unix socket/http/whatever interface that we can use to pass new configuration values into.

Or I'm sure there are myriad better ways to accomplish this. Basically, the important bit is "here are the new nsqlookupd addresses, use them instead, but don't stop running/serving requests, and please don't catch on fire".

I'm making the following assumptions when imagining how this would be run:

* A fresh nsqlookupd will be able to gather shared state from the registering consumers/nsqd instances quickly-ish (maybe a few seconds, at most?)
* Assuming each nsqd/consumer connects to at least two nsqlookupd instances, one going away doesn't actually affect anything except your redundancy level.


paddyforan:
> ```
    GET /config
         returns JSON body of current config
    POST /config
        body is JSON dictionary where keys are the config fields you want to update
```

My only complaint is the POST. I'd, personally, prefer PATCH. But I'm nit-picking here. I'd be _ecstatic_ to get this, even with POST.

File: nsqd/http.go [hunk 1]
```diff
@@ -596,3 +600,41 @@ func (s *httpServer) printStats(stats []TopicStats, health string, startTime tim
 	}
 	return buf.Bytes()
 }
+
+type allowedOpts struct {
+	NSQLookupdTCPAddresses []string `json:"nsqlookupd_tcp_addresses"`
```
paddyforan: It may be nice to support these as atomic add/remove _or_ overwrite. The way I've done this in the past, which isn't _too_ onerous, is

```go

type allowedOpts struct {
    NSQLookupdTCPAddresses []string `json:"nsqlookupd_tcp_addresses"`
    NSQLookupdTCPAddressesAdd []string `json:"nsqlookupd_tcp_addresses_add"`
    NSQLookupdTCPAddressesRemove []string `json:"nsqlookupd_tcp_addresses_remove"`
}
```

That way I have the option of overwriting the values _completely_, or I can just add a new address and remove a different one.

nsqd/http.go
```diff
@@ -596,3 +600,41 @@ func (s *httpServer) printStats(stats []TopicStats, health string, startTime tim
...
+	Verbose                bool     `json:"verbose"`
```
paddyforan: It'd be cool if this were a pointer, so I don't accidentally set Verbose false by default when updating the lookupd addresses, especially if I want it to be true.

mreiferson: :+1:

nsqd/http.go
mreiferson: I understand why that would be valuable but the API feels wrong.  The requirements really argue for _specific_ `nsqlookupd` config endpoints, which I was hoping to avoid.

paddyforan: I mean, where I'm using this elsewhere, I just continue adding pointers to allowedOpts. The only difference is nsqlookupdtcpaddresses gets 3 properties, instead of 1, and a use specifies at most 2 in any request.

_However_, I'm nitpicking. This is a Useful Thing because it avoids weird race conditions and provides atomicity, which is Always Useful once you start distributing things. (Look at me, explaining distributing things to _you_. :trollface:) In practice, I feel like it would be an edge case where this would be an issue. If you feel it's not worth the API bloat, that's totally cool. I thought I'd bring it up while I had the chance :P

mreiferson: Yes, I'm happy to discuss alternative APIs because now is the time to do it.  It's not that I disagree with the benefits, it's that I think we should try to come up with _an even better API_ if we want those semantics.

What about HTTP verbs?  What about individual endpoints?  If being more specific helps improve the API then I'm all for it...

Implementation wise, the HTTP side of this PR is the trivial stuff, most of the noise here is cleanup and proper handling of (now dynamic) state.

paddyforan: Verbs:

* PATCH uses the add/remove properties
* PUT uses the replace properties
* GET returns, as it does now
* DELETE is unused, because that would be silly

I'm :-/ about the semantics because it's not _really_ REST and is more RPC in spirit. However, the REST way to do it would be to do

* `POST /config/nsqlookupdtcpaddresses/` to add an address
* `PUT /config/nsqlookupdtcpaddresses/` to replace the entire list
* `DELETE /config/nsqlookupdtcpaddresses/{address}` to remove an address

But that seems excessive, and as I'm thinking about it, I'm probably overcomplicating matters. Let's step back for a second:

Can we think of a single service discovery mechanism that would give us an atomic view of the world? If we know about a change, we're almost certainly going to want to update the state of the world in DNS and etcd, at the very least.

As long as nsqd is intelligent about diffing updates (e.g., if it's connected to A & B, and you update the addresses to be A, B, and C, it just connects to C. it doesn't disconnect from A & B) then I'm having trouble thinking of a scenario where I'd actually use this.

Sorry for the knee jerk here, I may be off base about whether this is actually useful.

mreiferson: After a discussion offline, what if we just implemented `PUT /config/<option>` for now, which leaves room for some of these future decisions (the body is the JSON encoded value).

The `<option>` would match the equivalent key in the config file?

paddyforan: That would fit my use case. :+1:

mreiferson:
RFR @jehiah @paddyforan 

I did a ton of cleanup and the overall diff is noisy.  The individual commits are useful...

nsqd/http.go
```diff
@@ -596,3 +533,80 @@ func (s *httpServer) printStats(stats []TopicStats, health string, startTime tim
...
+		var v interface{}
+		err = json.Unmarshal(body, &v)
+		if err != nil {
+			return nil, http_api.Err{400, "INVALID_VALUE"}
+		}
+
+		opts := *s.ctx.nsqd.getOpts()
+		switch opt {
+		case "nsqlookupd_tcp_addresses":
+			var addrs []string
```
jehiah: I feel like this would be cleaner to Unmarshal  directly to the expected type inside this switch (ie: all the type checks go away)

```go
var addrs []string
if err := json.Unmarshal(body, &addrs); err != nil {
    return nil, http_api.Err{400, "INVALID_VALUE"}
}
opts.NSQLookupdTCPAddresses = addrs
```

any reason not to?

jehiah: actually i think we can Unmarshal directly to `json.Unmarshal(body, &opts.NSQLookupdTCPAddresses)` ??

jehiah:
LGTM aside from comments.

paddyforan:
Looks super great to me.

nsqd/lookup.go
```diff
@@ -129,6 +140,13 @@ func (n *NSQD) lookupLoop() {
 					break
 				}
 			}
+		case <-n.optsNotificationChan:
```
paddyforan: I may be reading this wrong, but it looks like on _every_ config change, the lookupds will be disconnected. So if I want to change the verbosity, that triggers a lookupd reconnection?

Would it be possible to diff the current lookupPeers and the list in opts to only add/remove the peers necessary? That would at least keep us from triggering a total reconnect each and every time any config value changes, which seems like (if this use case gets expanded upon) something that may happen more and more frequently over time.

nsqd/http.go
mreiferson: hah, that is _much_ simpler :+1:

nsqd/lookup.go
mreiferson: you're right, good catch, I'll fix

mreiferson:
PTAL @jehiah @paddyforan

paddyforan:
My only concern was addressed :+1:

jehiah:
:hammer: ?

mreiferson:
ready

ekristen:
The only problem I see with having an API only update approach and not supporting DNS is that you still have to have an outside force reconfigure each nsqd instance vs nsqd being able to reconfigure itself for multiple lookupd instances based on DNS.

Just my two cents.

PR #610, opened 2015-07-25
nsq*: log HTTP requests

mreiferson:
This adds a log decorator as middleware for the HTTP handlers.

Notably, this does not log /pub /mpub (should it?).

This also adds 404, 405, and "panic" handlers so that those responses
are consistently formatted.

RFR @jehiah

### DIFF
--- a/nsqadmin/http.go
+++ b/nsqadmin/http.go
@@ -3,7 +3,6 @@ package nsqadmin
 import (
 	"fmt"
 	"html/template"
-	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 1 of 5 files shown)


jehiah:
LGTM; there are probably a dozen ways to approach this, but this one feels fine.

nsqadmin/http.go
```diff
@@ -83,29 +87,29 @@ func NewHTTPServer(ctx *Context) *httpServer {
...
-	router.Handle("POST", "/unpause_channel", s.pauseChannelHandler)
-	router.Handle("POST", "/create_topic_channel", s.createTopicChannelHandler)
+	router.Handle("GET", "/ping", http_api.Decorate(s.pingHandler, log, http_api.PlainText))
+
+	router.Handle("GET", "/", http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", "/nodes", http_api.Decorate(s.nodesHandler, log))
+	router.Handle("GET", "/node/:node", http_api.Decorate(s.nodeHandler, log))
+	router.Handle("GET", "/topic/:topic", http_api.Decorate(s.topicHandler, log))
+	router.Handle("GET", "/topic/:topic/:channel", http_api.Decorate(s.channelHandler, log))
+	router.Handle("GET", "/static/:asset", http_api.Decorate(s.embeddedAssetHandler, log))
```
jehiah: @mreiferson turns out this doesn't work because log doesn't handle the response leaving nothing to render the response from the handler (ie: `s.embededAssetHandler`) into the response.

The effect is that requests for static assets return a 0 length content.

PR #612, opened 2015-07-31
nsqadmin: format responses

jehiah:
This fixes a regresion from #610 where some responses from nsqadmin were not handled properly (notably errors and /static/... assets)

RFR @mreiferson

### DIFF
--- a/nsqadmin/http.go
+++ b/nsqadmin/http.go
@@ -94,7 +94,7 @@ func NewHTTPServer(ctx *Context) *httpServer {
 	router.Handle("GET", "/node/:node", http_api.Decorate(s.nodeHandler, log, http_api.PlainText))
 	router.Handle("GET", "/topic/:topic", http_api.Decorate(s.topicHandler, log, http_api.PlainText))
 	router.Handle("GET", "/topic/:topic/:channel", http_api.Decorate(s.channelHandler, log, http_api.PlainText))
-	router.Handle("GET", "/static/:asset", http_api.Decorate(s.embeddedAssetHandler, log, http_api.PlainText))
+	router.Handle("GET", "/static/*asset", http_api.Decorate(s.embeddedAssetHandler, log, http_api.PlainText))
 	router.Handle("GET", "/counter", http_api.Decorate(s.counterHandler, log, http_api.PlainText))
 	router.Handle("GET", "/counter/data", http_api.Decorate(s.counterDataHandler, log, http_api.PlainText))
 	router.Handle("GET", "/lookup", http_api.Decorate(s.lookupHandler, log, http_api.PlainText))
@@ -128,6 +128,9 @@ func (s *httpServer) pingHandler(w http.ResponseWriter, req *http.Request, ps ht
 
 func (s *httpServer) embeddedAssetHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) (interface{}, error) {
 	assetName := ps.ByName("asset")
+	if strings.HasPrefix(assetName, "/") {
+		assetName = assetName[1:]
+	}
 
 	asset, error := templates.Asset(assetName)
 	if error != nil {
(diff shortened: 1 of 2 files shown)


nsqadmin/http.go
```diff
@@ -89,27 +89,27 @@ func NewHTTPServer(ctx *Context) *httpServer {
...
+	router.Handle("POST", "/tombstone_topic_producer", http_api.Decorate(s.tombstoneTopicProducerHandler, log, http_api.PlainText))
+	router.Handle("POST", "/empty_topic", http_api.Decorate(s.emptyTopicHandler, log, http_api.PlainText))
+	router.Handle("POST", "/delete_topic", http_api.Decorate(s.deleteTopicHandler, log, http_api.PlainText))
+	router.Handle("POST", "/pause_topic", http_api.Decorate(s.pauseTopicHandler, log, http_api.PlainText))
+	router.Handle("POST", "/unpause_topic", http_api.Decorate(s.pauseTopicHandler, log, http_api.PlainText))
+	router.Handle("POST", "/empty_channel", http_api.Decorate(s.emptyChannelHandler, log, http_api.PlainText))
+	router.Handle("POST", "/delete_channel", http_api.Decorate(s.deleteChannelHandler, log, http_api.PlainText))
+	router.Handle("POST", "/pause_channel", http_api.Decorate(s.pauseChannelHandler, log, http_api.PlainText))
+	router.Handle("POST", "/unpause_channel", http_api.Decorate(s.pauseChannelHandler, log, http_api.PlainText))
+	router.Handle("POST", "/create_topic_channel", http_api.Decorate(s.createTopicChannelHandler, log, http_api.PlainText))
```
ploxiln: maybe it makes sense to use a little function-local helper function at this point, maybe something like

```go
func router_reg(method, path string, handler http.Handler) {
    router.Handle(method, path, http_api.Decorate(handler, log), http_api.PlainText)
}
```

mreiferson:
whoooops, thanks :grin:

nsqadmin/http.go
mreiferson: not a bad idea but also not necessary

nsqadmin/http.go
```diff
@@ -89,27 +89,27 @@ func NewHTTPServer(ctx *Context) *httpServer {
...
-	router.Handle("POST", "/delete_channel", http_api.Decorate(s.deleteChannelHandler, log))
-	router.Handle("POST", "/pause_channel", http_api.Decorate(s.pauseChannelHandler, log))
-	router.Handle("POST", "/unpause_channel", http_api.Decorate(s.pauseChannelHandler, log))
-	router.Handle("POST", "/create_topic_channel", http_api.Decorate(s.createTopicChannelHandler, log))
+	router.Handle("GET", "/", http_api.Decorate(s.indexHandler, log, http_api.PlainText))
+	router.Handle("GET", "/nodes", http_api.Decorate(s.nodesHandler, log, http_api.PlainText))
+	router.Handle("GET", "/node/:node", http_api.Decorate(s.nodeHandler, log, http_api.PlainText))
+	router.Handle("GET", "/topic/:topic", http_api.Decorate(s.topicHandler, log, http_api.PlainText))
+	router.Handle("GET", "/topic/:topic/:channel", http_api.Decorate(s.channelHandler, log, http_api.PlainText))
+	router.Handle("GET", "/static/*asset", http_api.Decorate(s.embeddedAssetHandler, log, http_api.PlainText))
```
mreiferson: Meh, I should have dug into this a bit more - only this static route needed the output formatter, all the other template based endpoints write directly to the `http.ResponseWriter`.

It's fine though, it won't cause any problems and it's all going away in #323

jehiah: I *think* they still needed it for early error returns where the handler didn't write a response?

mreiferson: aha!

PR #582, opened 2015-08-08
nsqd: io.EOF when client-conn is closed

mreiferson:
NSQd does not distinguish io.EOF.
nsq/nsqd/protocol_v2.go

```Golang
        line, err = client.Reader.ReadSlice('\n')
        if err != nil {
            if atomic.LoadInt32(&client.State) == stateClosing {
                err = nil 
            } else {
                err = fmt.Errorf("failed to read command - %s", err)
            }   
            break
        }
```

### DIFF
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -62,7 +62,7 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 		// ie. the returned slice is only valid until the next call to it
 		line, err = client.Reader.ReadSlice('\n')
 		if err != nil {
-			if atomic.LoadInt32(&client.State) == stateClosing {
+			if err == io.EOF {
 				err = nil
 			} else {
 				err = fmt.Errorf("failed to read command - %s", err)


mreiferson:
according to the golang docs:

> EOF is the error returned by Read when no more input is available. Functions should return EOF only to signal a graceful end of input. If the EOF occurs unexpectedly in a structured data stream, the appropriate error is either ErrUnexpectedEOF or some other error giving more detail.

RFR @jehiah

mreiferson:
no errors in my testing:

```
[nsqd] 2015/08/08 16:01:27.656116 PROTOCOL(V2): [127.0.0.1:63472] exiting ioloop
[nsqd] 2015/08/08 16:01:27.656275 TOPIC(test): deleting channel tail433788#ephemeral
[nsqd] 2015/08/08 16:01:27.656290 CHANNEL(tail433788#ephemeral): deleting
[nsqd] 2015/08/08 16:01:27.656392 PROTOCOL(V2): [127.0.0.1:63472] exiting messagePump
```

fjjiaboming:
@mreiferson  Thanks.

PR #558, opened 2015-08-08
docker: root directory is /dist/docker/bin

mreiferson:
Is there a reason that / is set to /dist/docker/bin for the docker image?

I would like to use nsq_to_file and output messages to /dev/null to avoid the queues filling up my server disks while testing, which is not possible because of this.


mreiferson:
RFR @jehiah 

/cc @ploxiln

Issue #510, opened 2014-12-05
nsqd: resilience


stephensearles:
Here's a [proposal](https://docs.google.com/document/d/1PkaINjB35Z-El5pvjg0T7rewAg0KzTAJ7lJmuc27ENw/edit?usp=sharing) we're looking at to provide resilience on top of NSQ. This idea requires a small change to the nsqd library in that it would need to be notified when a message has been FIN'd.

EDIT: latest proposal [link](https://docs.google.com/document/d/1vxPKbLYckD80OiG9CxTWZufB4lv0b3hGZy4C9yBmI_k)

stephensearles:
So in light of the gossip work, I ended up writing a re-implementation of raft that uses serf as its mode of communication and source of node-liveness information. Perhaps its not accurate to call it raft, but that's definitely what it's based on. Other than the obvious advantage of a resilient communication layer, we can adapt to use the same serf instance for the lookup and resilience tasks. While deduplicating work, that will also deduplicate the need for bootstrapping. 

The documentation is here: http://godoc.org/github.com/shipwire/ansqd/internal/polity

For the WAL question, yes, I agree, there is likely value in making sure there's a persistent storage element to it. The current implementation for message storage is to group them by an "expiration time," removing messages as they get confirmed as handled by the cluster. The expiration time can be looked up periodically for any outstanding messages that need to be recovered. The implementation of that is here: https://github.com/shipwire/ansqd/blob/master/buckets.go

mreiferson:
>So in light of the gossip work, I ended up writing a re-implementation of raft that uses serf as its mode of communication and source of node-liveness information. Perhaps its not accurate to call it raft, but that's definitely what it's based on. Other than the obvious advantage of a resilient communication layer, we can adapt to use the same serf instance for the lookup and resilience tasks. While deduplicating work, that will also deduplicate the need for bootstrapping.
>
>The documentation is here: http://godoc.org/github.com/shipwire/ansqd/internal/polity

I want to make sure I'm on the same page with the direction you're proposing.  Polity would have to responsibility, on top of our ongoing gossip work, for "self-organizing" nodes in the cluster into various roles.  We would then use these "roles" to form replication groups (of sizes 3-5, etc.)?

From there, we can layer on a replication strategy between those role-based groups, strongly consistent or otherwise?

mreiferson:
> The documentation is here: http://godoc.org/github.com/shipwire/ansqd/internal/polity

[...]

PR #625, opened 2015-08-09
nsqd: per-topic WAL

mreiferson:
first part of #510


mreiferson:
This is a good sign so far:

screen shot 2015-08-09 at 2 21 02 pm

mreiferson:
Updated description, let me know if any of this sounds insane @jehiah.

/cc @stephensearles @judwhite

mreiferson:
**depth, `FIN` accounting, retention windows**

The current plan is for each topic to maintain a `RangeSet` (thanks @rolyatmax).  A `RangeSet` maintains a list of contiguous ranges of message IDs.  For the topic, it will store the ranges of IDs of messages that have been published.  For each channel, it will store the ranges of IDs of messages that have been `FIN`d (for that channel).  Thus, the difference in "count" between the two sets _should_ be the current depth of a given channel.  Topics will no longer have non-zero depth unless there are no channels present, in which case it is the count of its `RangeSet`.

I don't really know what to do about `BackendDepth` - what do you think @jehiah?

This is also the mechanism we can use to determine what the retention window for a topic should be.  By scanning each channel's `RangeSet`s, we can determine the lowest non-contiguous index.  The lowest index _across_ channels is the retention point for the topic.  Since we'll probably provide some configurable options around how much data we keep around, the largest of these values will be chosen as the _actual_ retention point.

Dieterbe:
any thoughts on how the performance of this compares for single channel, 100% of msg going into diskqueue vs into WAL, and 100% of reads going out of diskqueue vs WAL ?

(i presume once multiple channels come into play, the shared WAL vs individual diskqueues makes the WAL significicantly better)

mreiferson:
@Dieterbe working on correctness right now :grin:.  Performance should be better.

Dieterbe:
@mreiferson is ack-on-sync something you think should make it into nsqd as well?
i think it would be great if my producers could know a message is safely stored in the diskqueue/WAL and synced to disk. (i'm looking into nsqd for a new project and kinda need this feature.  if you think it sounds good i would happily work on this and PR it)

mreiferson:
@Dieterbe that already is the case in both the current production implementation (IF the message overflowed to disk) and this PR.  If it were not the case there would be no back-pressure!

Dieterbe:
@mreiferson I doesn't look like it.. or there is a misunderstanding.
`(d *diskQueue) Put(data []byte)` writes to d.writeChan and returns when it reads from d.writeResponseChan. and those ops only have a `count++` and a `writeOne()` (which only syncs when we open a new file) in between, the sync is only enforced at the next iteration of the loop. so `Put()` returns *before* the sync to disk happens (unless a new file is opened), so message loss can still happen. which is what i'm trying to avoid.  am i missing something? (also, sorry to distract slightly from your WAL work)

mreiferson:
Sorry, I misunderstood what you meant by "sync" - no, there is no mechanism to deliver fsync acks to producers.

The complexity doesn't feel like something that would be worth it.  The edge cases where you would lose messages, although they certainly exist, are probably better handled via replication.

You could also run with `--sync-every=1` but that would also be quite aggressive.

mreiferson:
NOTE: there is still back-pressure from the periodic syncing because `DiskQueue` cannot make progress during a sync and individual writes are synchronized with the `PUB`.

stephensearles:
Nice work! I'll dig into it a bit more later this week

Dieterbe:
@mreiferson: does this mean that
1) if this is enabled, we have guaranteed FIFO message ordering semantics?
2) will consumers be able to seek to arbitrary positions? for example if i have a log of the last 24h, and my consumer hasn't consumed anything in the last 4 hours, but it always prefers data 5 minutes old and more recent, and i know that i have a message per second will by consumer be able to say "seek to 300 messages before the end of the queue , and consume from there until the end" ? (and then later that consumer, or a different one, will make sure to also read the entire 4h range) something like this would be really useful to me.

mreiferson:
1. No, not exactly.  On a per-node basis, despite the WAL itself being FIFO, there are still requeues that will be redelivered based on their expiration.  And, across a cluster in aggregate, there is no coordination and no guaranteed ordering.

2. That will not be in this first pass, but is a feature we could consider providing down the road based on this work.

Dieterbe:
1. oh yes sure, i just meant single nsqd and assuming no requeues. a bit of a bold assumption i know.. let's call it "best case scenario" ;-)

2. cool thanks. i thought it wouldn't be possible due to messages having varying sizes so you wouldn't know where to seek to unless you start reading from the beginning. would this require adding an extra index datastructure or something? or is that already included in the design?

mreiferson:
2. It's already required to be able to seek to a certain index

icanfly:
@mreiferson  Great, this is what improve I want to get, at what time we can get the version available ? I think it is very helpful in case that the channels are much more , to reduce the copy between topic and channels

AshishKumarGoel:
The proposal looks very interesting. It will significantly improve the disk queue sizes and copy overheads. When can we expect this feature?

Dieterbe:
@mreiferson new years resolution to pick this back up ? :)

mreiferson:
@Dieterbe trying to land `1.0` first (this would _definitely_ not be in that release and would likely be a "2.0" kind of thing)

mreiferson:
I rebased this god forsaken branch yesterday and also wanted to drop this in here as a likely superior data structure for managing FINs/offsets:

http://web.engr.oregonstate.edu/~erwig/papers/Diet_JFP98.pdf

foxmailed:
any updates?

TomKaltz:
+1

PR #631, opened 2015-08-14
nsqadmin: drop display of protocol version

jehiah:
Since the protocol verison is always V2, let's drop it from nsqadmin and change the column name from protocol->user-agent.

RFR @mreiferson


mreiferson:
Yea, I'm fine with this.

PR #632, opened 2015-08-14
nsqadmin: clean up client-host display

jehiah:
Prior to #323 the "Client Host" column in nsqadmin displayed the hostname appended by the port as taken from the remote-address. That PR switched to always show client_id and remote_address. It was unclear though the ideal relationship between Hostname, Name, Short_id and Client ID (see #298). 

This clarifies the display to always show hostname + port (from remote_address), along with a hover state that shows the remote_address. Client ID will also be displayed when it's not equal to the hostname or a short version of the hostname.

RFR @mreiferson @rolyatmax


nsqadmin/static/js/models/channel.js
```diff
@@ -12,6 +12,25 @@ var Channel = Backbone.Model.extend({
...
+        var remote_address = v["remote_address"];
+      
+        // ignore client_id if it's duplicative
+        if (client_id === hostname.split(".")[0] || client_id === hostname) {
+          v["client_id"] = "";
+        }
+
+        var port = remote_address.split(":").pop()
+        v['hostname_port'] =  hostname + ":" + port;
+      })
```
rolyatmax: I don't know enough about when the `clients` attribute would change and what values we might expect `client_id`, `hostname`, and `remote_address` to be to review the logic of this code. I might tweak the naming a bit to make what this method is doing a little more explicit. For example, maybe something like `setHostnamePort` is better?

Besides that, it's mostly just being consistent with style (e.g. indentation, single quotes, missing semicolons, spaces between `function()` and `{`).

nsqadmin/static/js/models/channel.js
```diff
@@ -12,6 +12,25 @@ var Channel = Backbone.Model.extend({
         return AppState.url('/topics/' +
             encodeURIComponent(this.get('topic')) + '/' +
             encodeURIComponent(this.get('name')));
+    },
+    initialize: function() {
+      this.on('change:clients', this.calculateHostnamePort);
+    },
+    calculateHostnamePort: function(m, v) {
```
rolyatmax: I'd remove these arguments as they aren't used in the body of this method and can cause confusion with the `v` variable in the `forEach` method's iterator.

nsqadmin/static/js/models/channel.js
```diff
@@ -12,6 +12,25 @@ var Channel = Backbone.Model.extend({
...
+      
+      this.get("clients").forEach(function(v){
```
rolyatmax: Maybe `client` is better than `v`?

nsqadmin/static/js/models/channel.js
jehiah: oops forgot about those. good catch

nsqadmin/static/js/views/channel.hbs
```diff
@@ -178,7 +178,7 @@
             </tr>
             {{#each clients}}
             <tr>
-                <td title="{{remote_address}}">{{client_id}}</td>
+                <td title="{{remote_address}}">{{hostname_port}}{{#if client_id.length}} ({{client_id}}){{/if}}</td>
```
rolyatmax: I believe `{{#if client_id}}` should be sufficient if you're just trying to catch empty strings. Either way is fine though.

nsqadmin/static/js/models/channel.js
jehiah: :+1:

jehiah:
Hmm i realized i probably have some work i should also add in here to cleanup the display of the same fields on the nodes page as well.

jehiah:
nodes page updated. Any feedback @mreiferson?

nsqadmin/static/js/models/channel.js
```diff
@@ -12,6 +12,25 @@ var Channel = Backbone.Model.extend({
         return AppState.url('/topics/' +
             encodeURIComponent(this.get('topic')) + '/' +
             encodeURIComponent(this.get('name')));
+    },
+    initialize: function() {
+      this.on('change:clients', this.calculateHostnamePort);
+    },
+    calculateHostnamePort: function() {
```
mreiferson: vertical whitespace :cop:  (and between functions)

nsqadmin/static/js/models/channel.js
```diff
@@ -12,6 +12,25 @@ var Channel = Backbone.Model.extend({
...
+        var hostname = client["hostname"];
+        var remote_address = client["remote_address"];
+      
+        // ignore client_id if it's duplicative
+        if (client_id === hostname.split(".")[0] || client_id === hostname) {
+          client["client_id"] = "";
+        }
+
+        var port = remote_address.split(":").pop()
+        client['hostname_port'] =  hostname + ":" + port;
```
mreiferson: can we just keep these consistent with 4 spaces?

mreiferson:
got a screenshot with how this ends up looking by any chance?

mreiferson:
:+1:

PR #641, opened 2015-08-25
nsqd: HTTP pprof endpoints 404

mreiferson:
probably broke this in #601


mreiferson:
RFR @jehiah

jehiah:
so close, yet so far.

PR #396, opened 2015-08-27
nsqadmin: doesn't handle nsqd --tls-required flag

mreiferson:
From https://groups.google.com/d/msg/nsq-users/PkkPN9z7gPc/6EgTyHmuvg0J

`nsqadmin` cannot correctly negotiate TLS when `nsqd` in the cluster are configured with `--tls-required`.

Not sure yet how to fix this, some thoughts:

1. `nsqadmin` needs to be able to specify scheme in the `--nsqd-http-address` flag
2. `nsqd` needs to propagate TLS status to `nsqlookupd`, which needs to return this in its API responses
3. `nsqadmin` will pivot on (1) and (2) to negotiate which scheme to use

Thoughts @jehiah?

### DIFF
--- a/apps/nsq_stat/nsq_stat.go
+++ b/apps/nsq_stat/nsq_stat.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/bitly/nsq/internal/app"
 	"github.com/bitly/nsq/internal/clusterinfo"
+	"github.com/bitly/nsq/internal/http_api"
 	"github.com/bitly/nsq/internal/version"
 )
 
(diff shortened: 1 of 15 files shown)


mreiferson:
RFR @jehiah 

This updates `nsqd`'s `403` TLS response format to be JSON:

```json
{
    "message": "TLS_REQUIRED",
    "https_port": 4152
}
```

It updates `internal/http_api` to retry HTTP requests that receive a `403` and exposes new `nsqadmin` flags to specify TLS related HTTPS client config.

jehiah:
LGTM

jnicholls:
When do you all expect to make another stable release, with this functionality included? Thanks in advance.

mreiferson:
@jnicholls soon - we plan on addressing #421 and then stamping a stable release.

jnicholls:
That's great news, thanks @mreiferson!

PR #421, opened 2015-08-29
nsqadmin: display errors

mreiferson:
nsqadmin doesn't display when it's unable to query all the appropriate hosts to return stats. It should list them as errors at the top of the page.

(it can be missleading when the query to one host timesout but you don't realize it)


mreiferson:
work in progress:

screen shot 2015-08-29 at 2 57 54 pm

jehiah:
@mreiferson this approach looks good

mreiferson:
@rolyatmax would you mind taking a look at the JS error handling approaches here? :heart:

mreiferson:
@jehiah want to give this a spin?

nsqadmin/static/js/collections/nodes.js
```diff
@@ -2,7 +2,7 @@ var Backbone = require('backbone');
 
 var AppState = require('../app_state');
 
-var Node = require('../models/node');
+var Node = require('../models/node'); //eslint-disable-line no-undef
```
rolyatmax: Why do you need this?:
```
//eslint-disable-line no-undef
```

rolyatmax: You should add:
```
  "globals": {
    "module": true,
    "require": true
  }
```
to your `.eslintrc`.

nsqadmin/static/js/models/node.js
```diff
@@ -1,9 +1,7 @@
-var $ = require('jquery');
-
 var AppState = require('../app_state');
 var Backbone = require('backbone');
 
-var Node = Backbone.Model.extend({
+var Node = Backbone.Model.extend({ //eslint-disable-line no-undef
```
rolyatmax: Why do you need this? `//eslint-disable-line no-undef`

nsqadmin/static/js/views/app.js
```diff
@@ -45,11 +46,11 @@ var AppView = BaseView.extend({
...
-                    'target': $(this).attr('target')
-                }).done(function(data) {
-                    $(this).html(data['rate']);
-                }.bind(this));
+                        'metric': 'rate',
+                        'target': $(this).attr('target')
+                    })
+                    .done(function(data) { $(this).html(data['rate']); }.bind(this))
+                    .fail(function() { $(this).html('ERROR'); }.bind(this));
             });
```
rolyatmax: If instead of relying on jQuery's special handling of `this` (which can get confusing - especially when nested within several functions), you can be a little more explicit (and lose the `bind`s in the process) if you do something like this:

```
$('.rate').each(function(i, el) {
    $.get(AppState.url('/graphite'), {
        'metric': 'rate',
        'target': $(el).attr('target')
    })
    .done(function(data) { $(el).html(data['rate']); })
    .fail(function() { $(el).html('ERROR'); });
});
 ```

or even:

```
$('.rate').each(function(i, el) {
    var $el = $(el);
    $.get(AppState.url('/graphite'), {
        'metric': 'rate',
        'target': $el.attr('target')
    })
    .done(function(data) { $el.html(data['rate']); })
    .fail($el.html.bind($el, 'ERROR'));
});
 ```

nsqadmin/static/js/collections/nodes.js
mreiferson: is it because `Node` is a restricted keyword?

nsqadmin/static/js/views/app.js
```diff
@@ -133,9 +134,9 @@ var AppView = BaseView.extend({
             var node = new Node({
                 'name': nodeName
             });
-            node.tombstoneTopic(topicName).done(function() {
-                window.location.reload(true);
-            });
+            node.tombstoneTopic(topicName)
+                .done(function() { window.location.reload(true); })
+                .fail(this.handleAJAXError.bind(this));
         });
```
rolyatmax: Does this actually work?

The `this` in `.fail(this.handleAJAXError.bind(this))` is referring to whatever context `bootbox` uses when it calls this callback. (I'm guessing `this` will be incorrectly set to `window` - the result of calling an instance's method without a context.) You probably need to add `.bind(this)` to the `bootbox` callback, as well.

For example:

```
bootbox.confirm(txt, function(result) {
    ...
}.bind(this));
```

nsqadmin/static/js/collections/nodes.js
rolyatmax: Not a restricted keyword.

`Nodejs` is a _holy_ word - maybe that's what you're thinking of?

nsqadmin/static/js/views/channel.js
```diff
@@ -21,11 +21,18 @@ var ChannelView = BaseView.extend({
...
+                    this.template = require('./channel.hbs');
+                    this.render({'message': data['message']});
+                }.bind(this))
+            .fail(function(jqXHR) {
+                    this.template = require('./error.hbs');
+                    this.handleViewError(jqXHR);
+                }.bind(this))
+            .always(function() {
+                    Pubsub.trigger('view:ready');
+                });
```
rolyatmax: Muchisimos tabs, Batman. Personally, I'd unindent the bodies of these functions by one tab. But that's just a style comment. Take it or leave it :grimacing:

rolyatmax: Also, should `this.template = require('./error.hbs');` just go in the `handleViewError` method?

nsqadmin/static/js/views/base.js
```diff
@@ -83,6 +85,29 @@ var BaseView = Backbone.View.extend({
...
+    },
+
+    handleAJAXError: function(jqXHR) {
+        $('#warning, #error').hide();
+        $('#error .alert').text(this.parseErrorMessage(jqXHR));
+        $('#error').show();
+    },
+
+    handleViewError: function(jqXHR) {
+        this.render({'message': this.parseErrorMessage(jqXHR)});
```
rolyatmax: I'm seeing something like `this.template = require('./error.hbs')` appear before each time this method is called. Mutating `template` on these views puts them in a weird state. It might be a better practice to make `handleViewError` something like:
```
var errorTemplate = require('./error.hbs');

...

handleViewError: function(jqXHR) {
    this.removeSubviews();
    this.$el.html(errorTemplate({'message': this.parseErrorMessage(jqXHR)}));
}
```

nsqadmin/static/js/views/app.js
mreiferson: fixed

nsqadmin/static/js/views/app.js
mreiferson: fixed

nsqadmin/static/js/views/channel.js
mreiferson: I couldn't find a style I liked, I'll play around with it

nsqadmin/static/js/views/base.js
mreiferson: I like this

nsqadmin/static/js/views/lookup.js
```diff
@@ -20,14 +21,23 @@ var LookupView = BaseView.extend({
...
-            this.template = require('./lookup.hbs');
-            this.render({
-                'topics': _.map(data['topics'], function(v, k) {
-                    return {'name': k, 'channels': v};
-                })
-            });
-        }.bind(this));
+        $.ajax(AppState.url('/topics?inactive=true'))
+            .done(function(data) {
+                    this.template = require('./lookup.hbs');
```
rolyatmax: Swapping out templates like this seems a little strange to me (and prone to bugs). I usually have one template for a View. And if I need to insert spinners or error messages, I do that "manually". Typically, this ajax code would live in a model which would fire `Pubsub` events, like `loading`, and `error`. Views would then listen to these events and throw in error messages or spinners as needed.

All of this might seem to be more work than you want to do right now, however - especially if you're wanting to a rewrite (sayyyy... in React?)

mreiferson: Let's talk about this separately from error handling, I'm open to (simple) improvements before another React-factoring.

mreiferson:
updated @rolyatmax

nsqadmin/static/js/views/channel.js
rolyatmax: If you handled the templates like I was suggesting, you could do something like this:

```
        this.model.fetch()
            .done(function(data) {
                this.render({'message': data['message']});
            }.bind(this))
            .fail(this.handleViewError.bind(this))
            .always(Pubsub.trigger.bind(Pubsub, 'view:ready'));
```

But you'd need to make sure `template: require('./channel.hbs')` was in your View definition.

If you didn't mind `data`'s other properties getting into the render method, you could simplify further:

```
        this.model.fetch()
            .done(this.render.bind(this))
            .fail(this.handleViewError.bind(this))
            .always(Pubsub.trigger.bind(Pubsub, 'view:ready'));
```
...though it's not super explicit that `render` and `handleViewError` are both receiving arguments as the `done` and `fail` callbacks.

rolyatmax:
Updates look good to me. Have you tried producing all these error states? :grimacing:

mreiferson:
most of them :facepunch:

rolyatmax:
Sweeeeet - I'll leave the Go for a GoPro.

mreiferson:
@jehiah tested this a bunch against a real cluster - used https://github.com/apenwarr/sshuttle/ which is pretty amazing

mreiferson:
OK, I think this is RFR @jehiah

nsqadmin/static/html/index.html
```diff
@@ -13,7 +14,7 @@
     <div id="container"></div>
 
     <script type="text/javascript">
-        var USER_AGENT = 'nsqadmin';
+        var USER_AGENT = 'nsqadmin v{{.Version}}';
```
jehiah: `nsqadmin/{{.Version}}` to follow [rfc7231 section 5.5.3](http://tools.ietf.org/html/rfc7231#section-5.5.3)

jehiah:
LGTM aside from my nitpick. (sorry for the delay in review; i was offline for a few days)

mreiferson:
no worries, fixed and squashed in (remembered to update bin dater too this time)

RFM

PR #651, opened 2015-09-21
nsqadmin: fix deferred message count

jehiah:
The parsing of deferred message count for channels was broken by #323

RFR @mreiferson


mreiferson:
:+1:

PR #367, opened 2015-08-11
the great deprecation purge

mreiferson:
a WIP list of things that are currently deprecated that will be purged in `1.0`
- [x] old response format HTTP endpoints
  - [x] update HTTP docs (push nsqio/nsqio.github.io@57dbe89)
- [x] `BroadcastAddress` / `Address` dance
- [x] `--max-message-size` flag in `nsqd`
- [x] `--verbose` flag in bundled apps
- [x] `--gzip-compression` flag in `nsq_to_file`
- [x] `--round-robin` flag in `nsq_to_http`
- [x] `--max-backoff-duration` flag in `nsq_to_http` and `nsq_to_nsq`
- [x] `--throttle-fraction` flag in `nsq_to_http`
- [x] `--http-timeout-ms` flag in `nsq_to_http`
- [x] `long_id` and `short_id` attributes in `IDENTIFY`
- [x] `name` attribute in `/stats` response in `nsqd`
- [x] `--template-dir` flag in `nsqadmin`

PR #652, opened 2015-09-21
nsqadmin: fix regression when querying nsqd directly

jehiah:
This fixes the regression mentioned [here](https://github.com/nsqio/nsq/pull/642/files#r40002100)

cc: @mreiferson


mreiferson:
fix LGTM, can you push a commit removing this to #367?

PR #649, opened 2015-09-22
nsqadmin: dynamic graphite timeframe

mreiferson:
prior to #323 nsqadmin controlled the graphite integration timeframe via `?t={{timeframe}}` and persisted it via cookie. Now it's persisted via browser local storage.

The benefit of the previous approach is that you were not limited to the built in timeframe selection options and could pick any supported graphite time range (ie: `?t=3weeks`). This also meant you could hard link other people to view the dashboard in a specific timeframe which is helpful in troubleshooting in a group setting.

I think we should add the ability to select a custom timeframe (ie: a new 'other' option in the dropdown that presents a free form selection) and possibly add back the ability to pick up the timeframe option when present as a query parameter

cc: @mreiferson


nsqadmin/package.json
```diff
@@ -10,9 +10,9 @@
     "handlebars": "1.3.0",
     "hbsfy": "1.3.2",
     "browserify": "4.1.9",
-    "gulp-sass": "0.7.1",
+    "gulp-sass": "2.0.4",
```
mreiferson: I guess I had updated `node` at some point and this old version of `gulp-sass` wasn't working.

nsqadmin/static/js/app_state.js
```diff
@@ -17,7 +17,11 @@ var AppState = Backbone.Model.extend({
         this.on('change:graph_interval', function(model, v) {
             localStorage.setItem('graph_interval', v);
         });
-        this.set('graph_interval', localStorage.getItem('graph_interval') || 'off');
+
+        var qp = _.object(_.compact(_.map(window.location.search.slice(1).split('&'),
+            function(item) { if (item) { return item.split('='); } })));
```
mreiferson: @rolyatmax this is not completely absurd, right?

nsqadmin/static/js/app_state.js
```diff
@@ -17,7 +17,11 @@ var AppState = Backbone.Model.extend({
...
+        var interval = qp['t'] || localStorage.getItem('graph_interval') || 'off';
```
mreiferson: @jehiah note: there's no real parsing or validation, the code that reads this value requires it to be in the form `/[0-9]+h/`.

Do we need to support other durations?

(I should probably add a little validation here...)

jehiah: Let's just support them all and not care? `(s|seconds?|min|minutes?|h|hours?|d|days?|w|weeks?|mon|months?|y|years?)` seems to be the valid list (despite the [docs](http://graphite.readthedocs.org/en/latest/render_api.html#from-until) only showing a subset)

mreiferson: Oh, cool, then I don't even need to parse it.

jehiah: also, while here, can we default to `2h` when you have graphite configured? (we used to do that see [here](https://github.com/nsqio/nsq/pull/323/files#diff-c0e97967285692126a91af4fdca200daL130) )

mreiferson: SGTM

mreiferson:
updated @jehiah

mreiferson:
(I haven't actually tested this :grin:)

mreiferson:
Pushed another tweak to prefix the timeframes for the graphite URLs with `-`

jehiah:
@mreiferson I had to `require('underscore')` but this seems to work as expected.

mreiferson:
updated/squashed

nsqadmin/static/js/app_state.js
rolyatmax: I think that should work. Here it is without underscore (and in one loop):

```
var queryParams = window.location.search.slice(1).split('&').reduce(function(params, qStr) {
    if (qStr) {
        var pieces = qStr.split('=');
        params[pieces[0]] = pieces[1];
    }
    return params;
}, {});
```

Also, I think `array.map` is now supported by all major browsers (http://caniuse.com/#feat=es5), so there's also this:

```
var qp = _.object(_.compact(window.location.search.slice(1).split('&').map(function(item) {
    return item ? item.split('=') : null;
})));
```

PR #657, opened 2015-09-23
nsqadmin: improve error logging

jehiah:
This improves nsqadmin to show actual errors when unable to query all configured nsqd or nsqlookupd instances. Previously it left the specific error out which made it hard to troubleshoot. (@mreiferson if you'd prefer it only log the specific error i'm ok w/ that too)

This also updates the backwards compatability added in #652 to default hostname to the same value as BroadcastAddress.


mreiferson:
LGTM

Issue #661, opened 2015-09-28
nsqadmin: not respecting "use-statsd-prefixes" flag


kesutton:
I have just upgraded to NSQ 0.3.6, looking for the code fixed in #655.

However, it now seems as if the "use-statsd-prefixes" field is not being honored by nsqadmin.

Here is the output from NSQ Admin, running with both a true and false value for "use-statsd-prefixes":
```
[root@utility1 nsq]# nsqadmin --graphite-url="http://stats.crucible.dev.net:8000" --lookupd-http-address="utility1.crucible.dev.net:4161" --statsd-interval=1s --use-statsd-prefixes=true --statsd-prefix="nsq.%s"
[nsqadmin] 2015/09/28 14:33:00.074970 nsqadmin v0.3.6 (built w/go1.5.1)
[nsqadmin] 2015/09/28 14:33:00.075289 HTTP: listening on [::]:4171
[nsqadmin] 2015/09/28 14:33:01.788801 200 GET / (172.16.200.118:56701) 1.139171ms
[nsqadmin] 2015/09/28 14:33:01.922098 200 GET /static/bootstrap.min.css (172.16.200.118:56701) 14.103816ms
[nsqadmin] 2015/09/28 14:33:01.948642 200 GET /static/base.css (172.16.200.118:56702) 172.897µs
[nsqadmin] 2015/09/28 14:33:01.952700 200 GET /static/main.js (172.16.200.118:56704) 4.191885ms
[nsqadmin] 2015/09/28 14:33:01.954154 200 GET /static/vendor.js (172.16.200.118:56703) 5.683797ms
[nsqadmin] 2015/09/28 14:33:02.108851 200 GET /static/nsq_blue.png (172.16.200.118:56703) 574.021µs
[nsqadmin] 2015/09/28 14:33:02.113791 CI: querying nsqlookupd http://utility1.crucible.dev.net:4161/topics
[nsqadmin] 2015/09/28 14:33:02.117304 200 GET /api/topics (172.16.200.118:56704) 3.618575ms
[nsqadmin] 2015/09/28 14:33:02.120256 200 GET /fonts/glyphicons-halflings-regular.woff2 (172.16.200.118:56701) 189.417µs
[nsqadmin] 2015/09/28 14:33:02.132503 200 GET /static/main.js.map (172.16.200.118:56702) 12.481754ms
[...]

PR #662, opened 2015-09-29
nsqadmin: conf flags for formatting statsd counters and gauges

kesutton:
#661 

I want to reiterate that because there was a bug where the "use-statsd-prefixes" flag was not actually passed to nsqadmin, and thus useless, now would be the best time to eliminate that flag.  Unless it worked in the past, in which case I understand backwards compatibility.


nsqadmin/README.md
```diff
@@ -12,6 +12,6 @@ Read the [docs](http://nsq.io/components/nsqadmin.html)
  3. `$ gulp clean watch` or `$ gulp clean build`
  4. `$ go-bindata --debug --pkg=nsqadmin --prefix=static/build static/build/...`
  5. `$ cd ../nsqadmin && go build && ./nsqadmin`
- 6. make changes
+ 6. make changes (and repeat steps 4 and 5 to see changes)
```
mreiferson: This isn't _always_ true (specifically, it's only true when you're making changes to the Go code of `nsqadmin`).  Another way of communicating this is to say `(repeat step 5 if you make changes to any Go code)`

kesutton: Will update

mreiferson:
@jehiah what do you think?

mreiferson:
also, thanks for putting this together @kesutton!

nsqadmin/static/js/lib/handlebars_helpers.js
```diff
@@ -53,16 +72,19 @@ var genColorList = function(typ, key) {
 
 var genTargets = function(typ, node, ns1, ns2, key) {
     var targets = [];
-    var prefix = statsdPrefix(node ? node : '*', metricType(key));
+    var nsqPrefix = nsqStatsdPrefix(node ? node : '*');
```
mreiferson: style nitpick, can we drop the `nsq` on these variable/function names?

mreiferson:
I think this LGTM other than the minor comment, thanks!

kesutton:
I will fix the comment.  Did you guys come to a decision about the avlue of keeping legacy functionality that was already broken?

mreiferson:
I'm OK with it.

mreiferson:
Want to rebase/squash?

kesutton:
squashed and rebased on latest

mreiferson:
thanks

PR #666, opened 2015-10-01
dist.sh cleanups

ploxiln:
Clean up temp install dir, plus some other subjective cleanups.

(`rm -f` does not return error if the target does not exist)


mreiferson:
cool, thanks

PR #665, opened 2015-10-04
FreeBSD builds

mreiferson:
Since cross-compiling is so easy with the release of Go 1.5, could you please supply FreeBSD binaries?


mreiferson:
RFR @jehiah

PR #667, opened 2015-10-06
nsqadmin: add messages to the topics page

kpurdon:
This PR adds a basic stream\* of messages for a topic. By default it will show the latest 10 messages (or less than 10 for a default 250ms wait). New command-line options `messageTailMaxCount` and `messageTailMaxWait` control these values.

*not actually streaming (this may happen in future work)

preview image


mreiferson:
welp, that's pretty clever!

mrwoof:
hack week in the house! snaaaaaaaakes!

mreiferson:
sup @mrwoof :joy:

GeorgiCodes:
hawt!

gjohnson:
sweet!

gjohnson:
I'd almost say no streaming is more practical, at least in our case. I can't imagine you'd even be able to see the messages as they fly by.

kpurdon:
@gjohnson we got pretty close on the streaming work but not close enough (during Bitly hack week) so we'll be removing those commits and going with the non-streaming this time around. We did take into account high volume streams with some back-end and front-end limits. In addition we had a grep-tool on the FE that can stop as soon as it finds a match in the message stream. Hopefully we'll get back around to looking at streaming someday.

afd1d506990cf0b731a4b55c3ed451ae97acdca9 represents the last commit with non-streaming work complete. I'll try and get around to cleaning up this PR this week.

mreiferson:
Independent of streaming or not, my only UX request is that this be optional - i.e. the user should have to click a link or button or something to fetch messages for display.

kpurdon:
Ok, so here is my punch list:

* [ ] create branch to keep streaming work around (leave that crumb here)
* [ ] remove streaming work from this branch
* [ ] make message preview optional (button to fetch initial messages)
* [ ] *OPTIONAL:*explore grep search on message fetch (fetch until match?)
* [ ] tiny bit of cleanup on the pre-streaming work
* [ ] code review 
* [ ] happy dance

giphy 2

kpurdon:
For the future [here lies](https://github.com/kpurdon/nsq/tree/nsqadmin_streaming_messages) the streaming treasure we are leaving behind.

giphy 3

jehiah:
@kpurdon @mrwoof I think my only design comment here is that I imagine there are plenty of situations where a nsq datastream has sensitive information and exposing it via http through nsqadmin might be unexpected and a security issue. 

I think that means that at a minimum we should have a command line/config flag that enables/disables this feature (potentially with a topic regex?). I also second @mreiferson that fetching these messages separately from page load is desirable (and should keep the code simpler and allow for a longer default wait time). 

Ultimately i think a separate datastream introspection tool will the final solution that can grow to be more full featured (full disclosure: i'm biased by my previous work [here](https://github.com/jehiah/nsq_sketch)), but i think this is a fine spot to start.

xoen:
Hi All, this is nice as see the NSQ message's content could help to investigate weird issues.

On a related note, I'd love to be able to actually delete some of these NSQ messages as well. I understand that could be more involved/out of scope but it would have been extremely useful in a few occasions :)

kpurdon:
Going to go ahead and close this. I'm not likely to get back around to working on it and suspect as @jehiah noted there may be better ways (a separate introspection tool) to go at this.

PR #670, opened 2015-10-13
nsqd/nsqlookupd: IOLoop fix err shadow

judwhite:
`go tool vet -shadow` reported `lookup_protocol_v1.go:40: declaration of err shadows declaration at lookup_protocol_v1.go:25:`

In the current implementation `IOLoop` returns `nil` if there is an error other than from `reader.ReadString`. I considered if this was intentional but couldn't find a reason; if there is one please let me know.

This seems to only affect logging in `tcp.go`, however it is an exported method:

```go
	...

	err = prot.IOLoop(clientConn)
	if err != nil {
		p.ctx.nsqlookupd.logf("ERROR: client(%s) - %s", clientConn.RemoteAddr(), err)
		return
	}
}
```

Starting at line 49 in `nsqlookupd/lookup_protocol_v1.go`, is the second `err` check necessary?

```go
	_, err = protocol.SendResponse(client, []byte(err.Error()))
	if err != nil {
		break
	}

	// errors of type FatalClientErr should forceably close the connection
	if _, ok := err.(*protocol.FatalClientErr); ok {
		break
	}
```

### DIFF
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -213,7 +213,8 @@ func (s *httpServer) doPUB(w http.ResponseWriter, req *http.Request, ps httprout
 
 	var deferred time.Duration
 	if ds, ok := reqParams["defer"]; ok {
-		di, err := strconv.ParseInt(ds[0], 10, 64)
+		var di int64
+		di, err = strconv.ParseInt(ds[0], 10, 64)
 		if err != nil {
 			return nil, http_api.Err{400, "INVALID_DEFER"}
 		}
@@ -264,7 +265,8 @@ func (s *httpServer) doMPUB(w http.ResponseWriter, req *http.Request, ps httprou
 		rdr := bufio.NewReader(io.LimitReader(req.Body, readMax))
 		total := 0
 		for !exit {
-			block, err := rdr.ReadBytes('\n')
+			var block []byte
+			block, err = rdr.ReadBytes('\n')
 			if err != nil {
 				if err != io.EOF {
 					return nil, http_api.Err{500, "INTERNAL_ERROR"}
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -82,7 +82,8 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 			p.ctx.nsqd.logf("PROTOCOL(V2): [%s] %s", client, params)
 		}
 
-		response, err := p.Exec(client, params)
+		var response []byte
+		response, err = p.Exec(client, params)
 		if err != nil {
 			ctx := ""
 			if parentErr := err.(protocol.ChildErr).Parent(); parentErr != nil {
@@ -92,6 +93,7 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 
 			sendErr := p.Send(client, frameTypeError, []byte(err.Error()))
 			if sendErr != nil {
+				p.ctx.nsqd.logf("ERROR: [%s] - %s%s", client, sendErr, ctx)
 				break
 			}
 
--- a/nsqd/protocol_v2_test.go
+++ b/nsqd/protocol_v2_test.go
@@ -6,6 +6,7 @@ import (
 	"compress/flate"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -25,6 +26,7 @@ import (
 	"github.com/mreiferson/go-snappystream"
 	"github.com/nsqio/go-nsq"
 	"github.com/nsqio/nsq/internal/protocol"
+	"github.com/nsqio/nsq/internal/test"
 )
 
 func mustStartNSQD(opts *Options) (*net.TCPAddr, *net.TCPAddr, *NSQD) {
@@ -182,8 +184,8 @@ func TestMultipleConsumerV2(t *testing.T) {
 		go func(c net.Conn) {
 			resp, _ := nsq.ReadResponse(c)
 			_, data, _ := nsq.UnpackResponse(resp)
-			msg, _ := decodeMessage(data)
-			msgChan <- msg
+			recvdMsg, _ := decodeMessage(data)
+			msgChan <- recvdMsg
 		}(conn)
 	}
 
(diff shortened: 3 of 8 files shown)


judwhite:
Regarding the last code block, I noticed in [`nsqd/protocol_v2.go`](https://github.com/nsqio/nsq/blob/c3b3f73c067fe90260a5b9b3b52e70d1396f1897/nsqd/protocol_v2.go#L93) the first `err` is replaced with `sendErr`. This looks like better behavior. `sendErr` gets swallowed but the real `err` remains set, however it's still shadowed.

If you'd like we can follow up in a separate PR to keep this one small and focused.

mreiferson:
@judwhite doesn't look like it's necessary although it might have been intended to check the read error, not the send error

nsqlookupd/lookup_protocol_v1_test.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		write: func(b []byte) (int, error) {
+			return 0, errors.New("write error")
+		},
+	}
+
+	prot := &LookupProtocolV1{ctx: &Context{nsqlookupd: &NSQLookupd{opts: &Options{}}}}
+	err := prot.IOLoop(fakeConn)
+
+	equal(t, err == nil, false)
```
mreiferson: this can just be `equal(t, err, nil)`

judwhite:
Test results with unmodified `nsqd/protocol_v2.go` and `nsqlookupd/lookup_protocol_v1.go`

```
$ go test
--- FAIL: TestIOLoopReturnsClientErrWhenSendFails (0.02s)
        nsqd_test.go:59: nsqd v0.3.6 (built w/go1.5.1)
        nsqd_test.go:59: ID: 446
        nsqd_test.go:59: PROTOCOL(V2): [fe80::18d6:cdef:3d9f:977c/28] [INVALID_COMMAND]
        nsqd_test.go:59: ERROR: [fe80::18d6:cdef:3d9f:977c/28] - E_INVALID invalid command
                INVALID_COMMAND
        nsqd_test.go:59: PROTOCOL(V2): [fe80::18d6:cdef:3d9f:977c/28] exiting ioloop
        nsqd_test.go:45: protocol_v2_test.go:1518:

                        nexp: <nil>

                        got:  <nil>

--- FAIL: TestIOLoopReturnsClientErrWhenSendSucceeds (0.01s)
        nsqd_test.go:59: nsqd v0.3.6 (built w/go1.5.1)
        nsqd_test.go:59: ID: 446
        nsqd_test.go:59: PROTOCOL(V2): [fe80::18d6:cdef:3d9f:977c/28] [INVALID_COMMAND]
        nsqd_test.go:59: ERROR: [fe80::18d6:cdef:3d9f:977c/28] - E_INVALID invalid command
                INVALID_COMMAND
        nsqd_test.go:59: PROTOCOL(V2): [fe80::18d6:cdef:3d9f:977c/28] exiting ioloop
        nsqd_test.go:45: protocol_v2_test.go:1518:

                        nexp: <nil>

                        got:  <nil>

FAIL
exit status 1
FAIL    github.com/nsqio/nsq/nsqd       14.141s
```

```
$ go test
--- FAIL: TestIOLoopReturnsClientErrWhenSendFails (0.01s)
        nsqlookupd_test.go:45: nsqlookupd v0.3.6 (built w/go1.5.1)
        nsqlookupd_test.go:45: ERROR: [fe80::18d6:cdef:3d9f:977c/28] - E_INVALID invalid command
                INVALID_COMMAND
        nsqlookupd_test.go:45: CLIENT(fe80::18d6:cdef:3d9f:977c/28): closing
        nsqlookupd_test.go:31: lookup_protocol_v1_test.go:89:

                        nexp: <nil>

                        got:  <nil>

--- FAIL: TestIOLoopReturnsClientErrWhenSendSucceeds (2.00s)
        nsqlookupd_test.go:45: nsqlookupd v0.3.6 (built w/go1.5.1)
        nsqlookupd_test.go:45: ERROR: [fe80::18d6:cdef:3d9f:977c/28] - E_INVALID invalid command
                INVALID_COMMAND
        ... *snip 400 lines - hits timeout because IOLoop never exits*
        nsqlookupd_test.go:45: ERROR: [fe80::18d6:cdef:3d9f:977c/28] - E_INVALID invalid command
                INVALID_COMMAND
        nsqlookupd_test.go:22: lookup_protocol_v1_test.go:87:

                        exp: false

                        got: true

FAIL
exit status 1
FAIL    github.com/nsqio/nsq/nsqlookupd 2.829s
```

judwhite:
@mreiferson RFR

In short this is meant to ensure `IOLoop` returns `FatalClientErr` when it receives this error from `Exec`. Current nsqd implementation returns `nil`, nsqlookupd either returns `nil` or doesn't exit the loop depending on what SendResponse returns.

The error is already logged right after `Exec` returns an error, so I don't think the return of `nil` was that big of a deal. The non-breaking loop in nsqlookupd is probably NBD either since typically the only TCP connection is going to be nsqd.

Maybe the most useful thing from this is getting closer to reducing noise from `go tool vet -shadow`.

nsqd/http.go
```diff
@@ -213,8 +213,8 @@ func (s *httpServer) doPUB(w http.ResponseWriter, req *http.Request, ps httprout
 
 	var deferred time.Duration
 	if ds, ok := reqParams["defer"]; ok {
-		di, err := strconv.ParseInt(ds[0], 10, 64)
-		if err != nil {
+		di, parseErr := strconv.ParseInt(ds[0], 10, 64)
+		if parseErr != nil {
```
mreiferson: Are these being reported as errors in your linting?  Given the fact that an error here will return, I don't like having to complicate the naming of the error variable here.

i.e. my preference here, below and in `nsqd.go` are to keep these as they were.

nsqd/protocol_v2_test.go
```diff
@@ -227,11 +229,10 @@ func TestClientTimeout(t *testing.T) {
 		default:
 			_, err := nsq.ReadResponse(conn)
 			if err != nil {
-				goto done
+				break loop
```
mreiferson: this seems like a stylistic change that feels inconsistent with the rest of the codebase, would also prefer dropping this.

nsqd/protocol_v2_test.go
```diff
@@ -1452,6 +1453,83 @@ func runAuthTest(t *testing.T, authResponse, authSecret, authError, authSuccess
...
+		},
+		remoteAddr: func() net.Addr {
+			return netAddr
+		},
+		setDeadline:      func(time.Time) error { return nil },
+		setWriteDeadline: func(time.Time) error { return nil },
+		setReadDeadline:  func(time.Time) error { return nil },
+		closeFunc:        func() error { return nil },
+	}
+}
```
mreiferson: is it valuable to put this (`fakeNetAddr`, `fakeNetConn`) somewhere that would make it more discoverable and re-usable (thinking somewhere in the `internal` dir)?

nsqlookupd/lookup_protocol_v1.go
```diff
@@ -45,8 +45,8 @@ func (p *LookupProtocolV1) IOLoop(conn net.Conn) error {
 			}
 			p.ctx.nsqlookupd.logf("ERROR: [%s] - %s%s", client, err, ctx)
 
-			_, err = protocol.SendResponse(client, []byte(err.Error()))
-			if err != nil {
+			_, sendErr := protocol.SendResponse(client, []byte(err.Error()))
```
mreiferson: It seems like this error is now "lost" (i.e. it's not even logged).  Should we at least log it?

nsqd/http.go
judwhite: Yes, it is. Would you be okay with `var di int` and `di, err = ...` instead?

mreiferson: sure

nsqd/protocol_v2_test.go
judwhite: I thought I saw `continue loop` and `break loop` elsewhere, must've been somewhere else. I'll revert and update.

nsqd/protocol_v2_test.go
judwhite: I'll try it out and remove the `read` implementation, leaving that for the individual `Test` functions.

nsqlookupd/lookup_protocol_v1.go
judwhite: Will do.

judwhite:
@mreiferson updated and squashed

nsqd/protocol_v2.go
```diff
@@ -92,6 +93,7 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 
 			sendErr := p.Send(client, frameTypeError, []byte(err.Error()))
 			if sendErr != nil {
+				p.ctx.nsqd.logf("ERROR: [%s] - %s%s", client, sendErr, ctx)
```
mreiferson: should we do something like we do on https://github.com/nsqio/nsq/pull/670/files#diff-12a30208ec1e011f7cd53fe10c77575cR110 instead?

mreiferson:
Cool, this is looking good :100:

nsqd/protocol_v2.go
judwhite: Probably not. `Exec` is going to return a `FatalClientErr` and we don't want to overwrite it on our way out.

judwhite: Let me clarify without editing - if `Exec` returned a `FatalClientError` we probably want to returned error to be the one which determined `IOLoop` is going to exit. In this case, `sendErr` is secondary.

If `Exec` returned another type of error it may not have necessarily exited, but we are now that `p.Send` has failed. In this case, if we want to return the error which caused the loop to exit it would be `sendErr`. A few if statements and we could have this behavior.

I think, please correct me if I'm wrong, but at this time all errors returned by `Exec` are `FatalClientErr`.

PR #672, opened 2015-10-15
nsqd: add msg header size to diskqueue MaxMsgSize

judwhite:
- purpose: diskqueue was enforcing a maxmsgsize of the whole message
  including header, while protocol enforces maxmsgsize on the body. this
  could lead to successful writes to the memory queue but failed writes
  to the diskqueue, intermittently if the body size was within the
  26 bytes range.
- channel / topic tests assert configuration is passed correctly to
  diskqueue; it doesn't test PutMessage directly because enforcing msg
  size is the responsibility of protocol.
- add maxMsgSize to diskqueue writeOne error message.


mreiferson:
LGTM

PR #673, opened 2015-10-17
[WIP] HTTP/S Auth

judwhite:
opening to get early feedback.

---

- [x] Determine flags
  - `--auth-required=tcp,http,https`
- [ ] Determine new permissions (pause, empty, delete, tombstone, etc)
- [ ] HTTP/S auth
- [ ] Settle on client auth type's API
- [ ] TTL for HTTP/S (storing secret/permissions), or query auth server for each request?
- [x] Relax handling of unknown permissions, or negotiation with auth server?
  - Unknown permissions logged as warning
- [ ] nsqlookupd / nsqadmin updates (maybe in separate PR)
- [ ] Update documentation with flags, permissions, change to allow https auth server

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -74,6 +74,29 @@ func (t *tlsVersionOption) String() string {
 	return strconv.FormatInt(int64(*t), 10)
 }
 
+type authRequiredOption uint16
+
+func (t *authRequiredOption) Set(s string) error {
+	s = strings.ToLower(s)
+	for _, v := range strings.Split(s, ",") {
+		switch v {
+		case "tcp":
+			*t |= nsqd.FlagTCPProtocol
+		case "http":
+			*t |= nsqd.FlagHTTPProtocol
+		case "https":
+			*t |= nsqd.FlagHTTPSProtocol
+		default:
+			return fmt.Errorf("unknown auth-required value '%s'", v)
+		}
+	}
+	return nil
+}
+
+func (t *authRequiredOption) String() string {
+	return strconv.FormatInt(int64(*t), 10)
+}
+
 func nsqFlagset() *flag.FlagSet {
 	flagSet := flag.NewFlagSet("nsqd", flag.ExitOnError)
 
@@ -88,6 +111,9 @@ func nsqFlagset() *flag.FlagSet {
 
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> to query auth server (may be given multiple times)")
+
+	var authRequired authRequiredOption
+	flagSet.Var(&authRequired, "auth-required", "protocols which require auth when --auth-http-address is set (can be specified multiple times or comma separated 'tcp,http,https', default 'tcp')")
 
 	flagSet.String("broadcast-address", "", "address that will be registered with lookupd (defaults to the OS hostname)")
 	lookupdTCPAddrs := app.StringArray{}
--- a/internal/auth/authorizations.go
+++ b/internal/auth/authorizations.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/url"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/nsqio/nsq/internal/http_api"
@@ -88,13 +89,21 @@ func QueryAnyAuthd(authd []string, remoteIP, tlsEnabled, authSecret string) (*St
 	return nil, errors.New("Unable to access auth server")
 }
 
+func addScheme(url string) string {
+	lowerUrl := strings.ToLower(url)
+	if !strings.HasPrefix(lowerUrl, "http://") && !strings.HasPrefix(lowerUrl, "https://") {
+		url = fmt.Sprintf("http://%s", url)
+	}
+	return url
+}
+
 func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error) {
 	v := url.Values{}
 	v.Set("remote_ip", remoteIP)
 	v.Set("tls", tlsEnabled)
 	v.Set("secret", authSecret)
 
-	endpoint := fmt.Sprintf("http://%s/auth?%s", authd, v.Encode())
+	endpoint := fmt.Sprintf("%s/auth?%s", addScheme(authd), v.Encode())
 
 	var authState State
 	client := http_api.NewClient(nil)
@@ -105,10 +114,11 @@ func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error) 
 	// validation on response
 	for _, auth := range authState.Authorizations {
 		for _, p := range auth.Permissions {
+			// TODO: Add more granular permissions, for example delete, pause, etc
 			switch p {
 			case "subscribe", "publish":
 			default:
-				return nil, fmt.Errorf("unknown permission %s", p)
+				log.Printf("Warning: unknown permission %s", p)
 			}
 		}
 
--- a/internal/auth/authorizations.go
+++ b/internal/auth/authorizations.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/nsqio/nsq/internal/http_api"
@@ -24,6 +25,13 @@ type State struct {
 	Identity       string          `json:"identity"`
 	IdentityURL    string          `json:"identity_url"`
 	Expires        time.Time
+}
+
+var authStateCache map[string]State
+var authStateCacheMutex sync.RWMutex
+
+func init() {
+	authStateCache = make(map[string]State)
 }
 
 func (a *Authorization) HasPermission(permission string) bool {
@@ -105,7 +113,13 @@ func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error) 
 
 	endpoint := fmt.Sprintf("%s/auth?%s", addScheme(authd), v.Encode())
 
-	var authState State
+	authStateCacheMutex.RLock()
+	authState, ok := authStateCache[endpoint]
+	authStateCacheMutex.RUnlock()
+	if ok && !authState.IsExpired() {
+		return &authState, nil
+	}
+
 	client := http_api.NewClient(nil)
 	if err := client.GETV1(endpoint, &authState); err != nil {
 		return nil, err
@@ -138,5 +152,10 @@ func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error) 
 	}
 
 	authState.Expires = time.Now().Add(time.Duration(authState.TTL) * time.Second)
+
+	authStateCacheMutex.Lock()
+	authStateCache[endpoint] = authState
+	authStateCacheMutex.Unlock()
+
 	return &authState, nil
 }
--- a/nsqd/auth.go
+++ b/nsqd/auth.go
@@ -24,12 +24,10 @@ func (a *AuthService) Auth(secret string) error {
 }
 
 func (a *AuthService) QueryAuthd() error {
-	var tlsEnabled string
-	tls := atomic.LoadInt32(&a.TLS)
-	if tls == 1 {
+	tls := atomic.LoadInt32(&a.TLS) == 1
+	tlsEnabled := "false"
+	if tls {
 		tlsEnabled = "true"
-	} else {
-		tlsEnabled = "false"
 	}
 
 	authState, err := auth.QueryAnyAuthd(a.AuthHTTPAddresses, a.RemoteIP, tlsEnabled, a.AuthSecret)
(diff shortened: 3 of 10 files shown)


apps/nsqd/nsqd.go
```diff
@@ -88,6 +88,8 @@ func nsqFlagset() *flag.FlagSet {
 
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> to query auth server (may be given multiple times)")
+	flagSet.Bool("http-auth-required", false, "require auth for HTTP clients")
+	flagSet.Bool("https-auth-required", false, "require auth for HTTPS clients")
```
mreiferson: I've got to think about this some more, but I'm not excited about the idea of adding two more flags for this.

It might mean we need to rethink the existing combination of TLS/Auth related flags, choose good/flexible ones, and deprecate existing ones.  This might introduce some short term implementation pain but open up the door for getting this right going forward.

internal/auth/authorizations.go
```diff
@@ -88,13 +89,21 @@ func QueryAnyAuthd(authd []string, remoteIP, tlsEnabled, authSecret string) (*St
...
+}
+
 func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error) {
 	v := url.Values{}
 	v.Set("remote_ip", remoteIP)
 	v.Set("tls", tlsEnabled)
 	v.Set("secret", authSecret)
 
-	endpoint := fmt.Sprintf("http://%s/auth?%s", authd, v.Encode())
+	endpoint := fmt.Sprintf("%s/auth?%s", addScheme(authd), v.Encode())
```
mreiferson: :+1:

internal/auth/authorizations.go
```diff
@@ -108,6 +117,7 @@ func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error)
 			switch p {
 			case "subscribe", "publish":
 			default:
+				// TODO: Can we add more granular permissions, for example delete, pause, etc?
```
mreiferson: sure, why not?

nsqd/client_v2.go
```diff
@@ -499,6 +503,7 @@ func (c *clientV2) UpgradeTLS() error {
 	c.Writer = bufio.NewWriterSize(c.tlsConn, c.OutputBufferSize)
 
 	atomic.StoreInt32(&c.TLS, 1)
+	atomic.StoreInt32(&c.ClientAuth.TLS, 1)
```
mreiferson: it feels leaky to package up the auth related metadata into `ClientAuth` but then have `clientV2` modify struct fields directly.

nsqd/protocol_v2.go
```diff
@@ -504,7 +504,7 @@ func (p *protocolV2) AUTH(client *clientV2, params [][]byte) ([]byte, error) {
 		return nil, protocol.NewFatalClientErr(err, "E_BAD_BODY", "AUTH failed to read body")
 	}
 
-	if client.HasAuthorizations() {
+	if client.ClientAuth.HasAuthorizations() {
```
mreiferson: Related, I think this is an indication that we might want to _embed_ `ClientAuth` in `clientV2` so that we don't have to change the way we access these methods (feels redundant to write `client.ClientAuth.<blah>`).

apps/nsqd/nsqd.go
judwhite: Agreed. Originally I was mirroring the `--*-address` flags.

One option might be to have an `--auth-required` or `--require-auth` flag. It could be specified multiple times, or it could be of the form `tcp,http,https` with a default of `tcp` to preserve existing functionality. The comma separated options breaks convention with existing flags, however without it you wouldn't be able to both preserve existing functionality and allow disabling auth for `tcp` without some funky logic.

What do you think?

nsqd/client_v2.go
judwhite: Agreed. I think `TLS` should live in `clientV2` and be passed to the auth functions.

nsqd/protocol_v2.go
judwhite: I agree it doesn't look pleasant. I was thinking of `ClientAuth` as a service which `clientV2` depends on. I'm not sure if that makes it a good candidate to embed as part of the type or if better naming would address the issue while making the separation clear.

internal/auth/authorizations.go
```diff
@@ -108,6 +117,7 @@ func QueryAuthd(authd, remoteIP, tlsEnabled, authSecret string) (*State, error)
...
 				return nil, fmt.Errorf("unknown permission %s", p)
 			}
```
judwhite: If an auth server returns additional permissions it would cause `unknown permission` to occur on previous versions of `nsqd`. Do you think this should be relaxed, for example log the unknown permission and move on?

Another thing I thought about was a sort of "feature negotiation" with the auth server; either passing the `nsqd` or `nsqlookupd` version or passing the set of permissions it's able to handle. We could delegate the responsibility of trimming the list and/or erroring to the auth server. I'm not sold this is worthwhile, but without some kind of change (relaxing or negotiation) the auth server and `nsqd` versions will need to be kept in sync - unless a different `secret` is used by the client.

apps/nsqd/nsqd.go
mreiferson: That's kinda where I was headed too, let's experiment with that approach?

FWIW, it's consistent with how `--e2e-processing-latency-percentile` works.

nsqd/protocol_v2.go
mreiferson: Right, a better name might also be sufficient, let's see how it plays out.

internal/auth/authorizations.go
mreiferson: I think logging is sufficient. I think the current implementation of erroring here is unnecessary, and the logging might help debug situations where you're not getting the permissions you expect (typos, etc.).

nsqd/auth.go
```diff
@@ -0,0 +1,55 @@
...
+	} else {
+		authService.tlsEnabled = "true"
+	}
+
+	err := authService.queryAuthd()
+	if err != nil {
+		return nil, err
+	} else {
+		return authService, nil
+	}
```
judwhite: TODO: Don't want to call auth service in `NewAuthService`

nsqd/client_v2.go
```diff
@@ -223,9 +224,11 @@ func (c *clientV2) Stats() ClientStats {
...
-	if c.AuthState != nil {
-		identity = c.AuthState.Identity
-		identityURL = c.AuthState.IdentityURL
+	var authed bool
+	if c.Auth != nil && c.Auth.AuthState != nil {
+		identity = c.Auth.AuthState.Identity
+		identityURL = c.Auth.AuthState.IdentityURL
+		authed = true
 	}
 	c.RUnlock()
```
judwhite: TODO: Smells funny.

nsqd/http.go
```diff
@@ -114,6 +114,43 @@ func (s *httpServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+func (s *httpServer) checkAuth(req *http.Request, cmd, topicName, channelName string) error {
+	var protocol uint16
+	var tls int32
+	if req.TLS != nil {
+		protocol = FlagHTTPSProtocol
+		tls = 1
+	} else {
+		protocol = FlagHTTPProtocol
+		tls = 0
+	}
```
judwhite: TODO: Might be better extracted into its own method.

nsqd/protocol_v2.go
```diff
@@ -504,21 +506,23 @@ func (p *protocolV2) AUTH(client *clientV2, params [][]byte) ([]byte, error) {
...
 	}
 
-	if !client.ctx.nsqd.IsAuthEnabled() {
+	if !client.ctx.nsqd.IsAuthEnabled(FlagTCPProtocol) {
 		return nil, protocol.NewFatalClientErr(err, "E_AUTH_DISABLED", "AUTH Disabled")
 	}
 
-	if err := client.Auth(string(body)); err != nil {
+	client.Auth, err = NewAuthService(opts.AuthHTTPAddresses, string(body), client.remoteIP, client.TLS)
+	if err != nil {
```
judwhite: TODO: Not happy about `protocol_v2` setting `client.Auth`.

judwhite:
@mreiferson can I get your feedback on the flags?

nsqd/http.go
```diff
@@ -114,6 +114,51 @@ func (s *httpServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	var tls int32
+	if req.TLS != nil {
+		isAuthEnabled = s.ctx.nsqd.IsAuthEnabledHTTPS
+		tls = 1
+	} else {
+		isAuthEnabled = s.ctx.nsqd.IsAuthEnabledHTTP
+		tls = 0
+	}
+
+	if isAuthEnabled() {
```
mreiferson: style nitpick, but let's invert this and return early so the primary code path is less indented.

mreiferson:
@judwhite :+1: on the flag handling, came out nice!

nsqd/http.go
judwhite: done, updated `protocol_v2` `CheckAuth` also.

nsqd/http.go
```diff
@@ -114,6 +114,58 @@ func (s *httpServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	}
+	return s.ctx.nsqd.IsAuthEnabledHTTP()
+}
+
+func (s *httpServer) isTLS(req *http.Request) int32 {
+	if req.TLS != nil {
+		return 1
+	}
+	return 0
+}
```
mreiferson: I feel like these two methods above might be a little prematurely factored.

1. they're not really re-used
2. `isTLS` doesn't technically need the `s *httpServer` receiver
3. it was initially confusing when seeing _another_ `isAuthEnabled` variant attached to `httpServer`

I vote for just consolidating this into `checkAuth`.  What do you think?

nsqd/http.go
```diff
@@ -114,6 +114,58 @@ func (s *httpServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	err := authService.Auth("TODO_SECRET")
+	if err != nil {
+		// we don't want to leak errors contacting the auth server to untrusted clients
+		s.ctx.nsqd.logf("ERROR: [%s] Auth Failed %s", s, err)
+		return http_api.Err{500, "E_AUTH_FAILED"}
+	}
+	// TODO: We'll need something like this if the secret isn't passed with the request
+	/*if !auth.HasAuthorizations() {
+		return http_api.Err{401, "E_AUTH_FIRST"}
+	}*/
```
mreiferson: Can you explain this commented out TODO?

judwhite: Reminder to check the Header or query param for the `secret` and return a 401 if not present.

Should a query param be supported or just a header? Any preference for naming?

nsqd/http.go
judwhite: Maybe. Lots left to do on the HTTP/S side, still experimenting here.

nsqd/http.go
mreiferson: Header, `X-AuthToken`?

judwhite: :+1: Any preference between `X-AuthToken` or `X-NSQ-AuthToken` ?

judwhite:
@mreiferson Ready for feedback on adding a cache to `internal/auth/authorizations.go`. Thinking behind this is to allow `httpServer` to call `auth.QueryAnyAuthd` without making an HTTP request to the auth server for every incoming request.

Another options would be to move the cache to `httpServer`.

internal/auth/authorizations.go
```diff
@@ -25,6 +27,13 @@ type State struct {
 	Expires        time.Time
 }
 
+var authStateCache map[string]State
+var authStateCacheMutex sync.RWMutex
+
+func init() {
+	authStateCache = make(map[string]State)
+}
```
mreiferson: This is a good start, but I don't think we can keep this "global", I think we should attach it to some of the other structs we're using for auth.

Similarly, this might be addressed as a side effect, I'm not sure `QueryAuthd` is the right spot to do the caching.  It feels like a responsibility of a higher level abstraction, or at least a wrapper for `QueryAuthd`?

judwhite:
Dropping a note for myself. The `AUTH` portion, at least for HTTP, might come out better if it's broken out into a Gateway layer, and written with re-use in `nsqlookupd` in mind. The HTTP layer is already sort of acting like this, need to think about this some more.

mreiferson:
@judwhite how are you feeling about this? :grin:

judwhite:
@mreiferson at the moment I don't have a use case or much time to put into this change. thanks for the feedback so far, I'll request a reopen if I get the time.

mreiferson:
SGTM, thanks @judwhite

nsqd/http.go
judwhite: @mreiferson I may want to re-open this. Was there a reason we didn't go for the standard `Authorization` header?

jehiah: no specific reason i can think of

PR #676, opened 2015-10-29
nsqd/nsqlookupd: support running as windows service

judwhite:
I tried making the least modifications to the existing flow as possible.

Tested flipping the `main.go` and `main_windows.go` files, both builds run, passes smoke test, and the service starts/stops as expected and responds to `^C`.

On Windows it'll run from the command line and respond to `^C`. If it's running as a Windows Service it'll respond to start/stop commands and do a graceful shutdown during reboot/restart.

FWIW we've been running code [similar](https://github.com/judwhite/nsq-0.3.2-win/commits/master) to this in production for a few months. This version makes less modifications.

Example installation:

```
sc create nsqlookupd binpath= "c:\nsq\nsqlookupd.exe" start= auto DisplayName= "nsqlookupd"
sc description nsqlookupd "nsqlookupd 0.3.6"
sc failure nsqlookupd actions= restart/5000/restart/60000/restart/300000// reset= 120
sc start nsqlookupd

sc create nsqd binpath= "c:\nsq\nsqd.exe -mem-queue-size=0 -lookupd-tcp-address=127.0.0.1:4160 -data-path=c:\nsq\data" start= auto 
sc description nsqd "nsqd 0.3.6"
sc failure nsqd actions= restart/5000/restart/60000/restart/300000// reset= 120
sc start nsqd
```

The missing piece for our Windows friends is logging to disk, unfortunately ` 2>>nsqd.log` doesn't cut it when running as a service. Thankfully we've never needed them. :smirk: 

:pray:

### DIFF
--- a/apps/nsqlookupd/main.go
+++ b/apps/nsqlookupd/main.go
@@ -1,0 +1,22 @@
+// +build !windows
+
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func main() {
+	nsqlookupd := start()
+	if nsqlookupd == nil {
+		return
+	}
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+	<-signalChan
+
+	nsqlookupd.Exit()
+}
--- a/Godeps
+++ b/Godeps
@@ -8,3 +8,4 @@ github.com/mreiferson/go-snappystream   028eae7ab5c4c9e2d1cb4c4ca1e53259bbe7e504
 github.com/bitly/timer_metrics          afad1794bb13e2a094720aeb27c088aa64564895
 github.com/blang/semver                 9bf7bff48b0388cb75991e58c6df7d13e982f1f2
 github.com/julienschmidt/httprouter     6aacfd5ab513e34f7e64ea9627ab9670371b34e7
+golang.org/x/sys/windows                354f231ae1a9ca2d3a6a1a7c5d40b1213d761675
--- a/apps/nsqd/main.go
+++ b/apps/nsqd/main.go
@@ -1,0 +1,22 @@
+// +build !windows
+
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func main() {
+	nsqd := start()
+	if nsqd == nil {
+		return
+	}
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
+	<-signalChan
+
+	nsqd.Exit()
+}
--- a/apps/nsqd/main_windows.go
+++ b/apps/nsqd/main_windows.go
@@ -1,0 +1,47 @@
+// +build windows
+
+package main
+
+import (
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/nsqio/nsq/internal/winservice"
+	"github.com/nsqio/nsq/nsqd"
+)
+
+type program struct {
+	nsqd *nsqd.NSQD
+}
+
+func main() {
+	prg := &program{}
+	if err := winservice.Run(prg); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (p *program) Start() error {
+	p.nsqd = start()
+	if p.nsqd == nil {
+		// --version
+		os.Exit(0)
+	}
+	return nil
+}
+
+func (p *program) Stop() error {
+	if p.nsqd != nil {
+		p.nsqd.Exit()
+	}
+	return nil
+}
+
+func (p *program) BeforeStart(e winservice.Environment) error {
+	if !e.IsAnInteractiveSession() {
+		dir := filepath.Dir(os.Args[0])
+		return os.Chdir(dir)
+	}
+	return nil
+}
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -7,10 +7,8 @@ import (
 	"log"
 	"math/rand"
 	"os"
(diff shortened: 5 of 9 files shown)


judwhite:
@mreiferson ready for review

mreiferson:
@jud will take a look today

judwhite:
@mreiferson friendly ping, when you have a chance. let me know if you need any more info or justification. long story short this is to make nsqd production ready on Windows.

mreiferson:
@judwhite sorry, a little behind - I'll try to take a pass today

apps/nsqlookupd/main.go
```diff
@@ -0,0 +1,22 @@
...
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func main() {
+	daemon := start()
```
mreiferson: in the `main_windows.go` file we called this var `nsqd`, which I think I prefer.

mreiferson: sorry, did not realize this was for `nsqlookupd`, in which case it should be `nsqlookupd` :)

mreiferson:
Minor comments aside, the changes themselves are surgical, nice work :+1:.

At a higher level, it's _really_ unfortunate that we have to do this at all, and I don't feel great about catering to a specific OS like this.

Is it possible to write a generic Windows service binary wrapper that would `exec` nsqd, capture stderr/stdout and listen/respond to the Windows service hooks?  Why does this _have_ to be baked in?

What do you think @jehiah?

jehiah:
@mreiferson ditto your comment. If there are no other realistic paths to simple out-of-the-box windows setup, then i'm :+1:. It is unfortunate.

judwhite:
Thanks guys, believe me I also wish there was more parity between OS's here.

I'll make the variable name change (`daemon` was taken from `apps/nsqlookupd/nsqlookupd.go`). I'm also going to add an `os.Chdir` in the Windows service to change the cwd to the binary's location. Otherwise dat files end up in `C:\Windows\System32` unless you specify `-data-path`. Although hilarious, I don't think this should be the default behavior.

Thanks again.

mreiferson:
@jud any thoughts on:

> Is it possible to write a generic Windows service binary wrapper that would exec nsqd, capture stderr/stdout and listen/respond to the Windows service hooks? Why does this have to be baked in?

Jud:
@mreiferson Probably have better luck pinging @judwhite

mreiferson:
hah, sorry!

judwhite:
@jud welcome! :grinning: 

@mreiferson afaik the best solution is [Non-Sucking Service Manager](https://nssm.cc) which should tell you something about how many people have tried. Their page admits versions prior to 2.19 sucked. It may be possible, but it's an undertaking for sure. fwiw the Windows Service code has 100% coverage and imo good asserts, though I get the goal. One of the things we really appreciate about Go is the single binary, even if we pulled it off we lose that benefit. Long story short I wouldn't feel comfortable doing this by myself, but I'm open to ideas to make this better.

mreiferson:
> One of the things we really appreciate about Go is the single binary, even if we pulled it off we lose that benefit. Long story short I wouldn't feel comfortable doing this by myself, but I'm open to ideas to make this better.

It still is a single binary though? I see this more as "process supervision", like I would use `daemontools` on linux.

judwhite:
Squashed, RFR. Summary of changes:
- [Changed `daemon` variable name to `nsqlookupd`](https://github.com/nsqio/nsq/pull/676/files#diff-14161ed9e943756703377c17dc099101R1)
- [Windows Service: Call `os.Chdir` to prevent files ending up in C:\Windows\System32 when `-data-path` is not specified](https://github.com/nsqio/nsq/pull/676/files#diff-4730cef7ba8fe0ae5ac6bfda561976cfR41)

Matt, NSSM is the closest you can get on Windows afaik. It requires administrator access which is not something most people like to grant without a really good reason.

mreiferson:
From offline discussion in IRC:

I made the suggestion that @jud / we (nsqio) maintain a separate windows-only `nsqd` and `nsqadmin` repo that would include all of these OS-specific features.

This PR could make whatever internal changes are necessary to support that.  Specifically, we might want to improve how `nsqd` and `nsqlookupd` handle command line flags, so that @jud's windows binary app can import and use that code.

Thoughts @jehiah ?

Jud:
@mreiferson you're killin' me.

mreiferson:
LOL, sorry man :(

mreiferson:
@judwhite @jehiah let's circle back on https://github.com/nsqio/nsq/pull/676#issuecomment-161453879 and make a decision here?

judwhite:
Hey, I thought about this some more.

I feel pretty strongly about the following:
- `nsqd` and `nsqlookupd` should support running as a Windows service
- The code should live in the nsqio organization

Do you agree with the above?

I think something could be done with an internal package and `// +build` flags to abstract this away a little. `nsqd` and `nsqlookupd` would no longer have build flags but would have to tie into the internal package. You'd get a more uniform look at the binary level and move the responsibility of handling `SIG_TERM`, or registering as a Windows service, into the internal package. I think the end product would look more like the Windows `.go` file because it has to handle the start/stop signals, but we can make it look a little less Windows-y. Does that make sense, and is it worthwhile?

mreiferson:
#718 is where it's at, closing, thanks @judwhite!

elvarb:
Are there any plans to add this to the other NSQ tools?

judwhite:
@elvarb which ones are you interested in? I'll raise a PR for those and see if Matt/Jehiah want to do the whole lot

elvarb:
Mostly interested in nsq_to_nsq for now, using it to move data from a local nsqd instance (one per server) to a central nsqd server

PR #683, opened 2015-11-11
*: increase HTTP client transport timeout

alvarolm:
Client is implemented in the authorization of nsqd for requesting it, sometimes my authorization server has unexpected latency increases which I can't control, making the default 2 seconds timeout useless.


mreiferson:
@alvarolm thanks for the PR.  Unfortunately, this timeout in question affects many different HTTP operations.  Why don't we make it configurable as a command line flag (default to the current `2s`)?

mreiferson:
opened #715 to track this, closing

PR #690, opened 2015-11-30
Fix defaultID calculating error

zieckey:
defaultID has 12 bits which give us up to 4096 IDs.


mreiferson:
@zieckey "worker ID" has 10 bits

PR #700, opened 2015-12-31
nsqd: stats call times out requests during high load

zachbadgett:
While maxing out (reading and writing) my local nsqd with many topics (9 in this case), I noticed calling the stats api causes all requests from my consumers and producers to timeout and the stats call, if completed, takes around 14 seconds.

I believe this is caused by unneeded locking in the GetStats method. Removing the deferred and moving the RUnlock fixes the issue. During high load the stats call remained under 2ms and the consumers and producers no longer timeout. Tested with the race detector and everything looks good.


jehiah:
nice. thanks!

mreiferson:
:joy: nice catch

PR #706, opened 2016-01-14
nsqd: enable block profile

absolute8511:


mreiferson:
@absolute8511 this is already available at runtime via the HTTP pprof endpoints http://nsq.io/components/nsqd.html#get-debugpprofblock, closing.

absolute8511:
block profile will not be enabled by default if you did not `SetBlockProfileRate`

mreiferson:
@absolute8511 sorry! didn't realize that - it seems like this is something you don't want on all the time (hence the default being off).  Perhaps we need another `/debug/...` endpoint to temporarily enable it?

absolute8511:
Maybe using `/debug/...` would be better, but I am not sure is it safe to set this dynamically.

mreiferson:
> Maybe using /debug/... would be better, but I am not sure is it safe to set this dynamically.

According to http://grokbase.com/p/gg/golang-nuts/152ra4r1vz/go-nuts-trying-to-use-the-pprof-block-endpoint, that's the way it's expected to be used.

nsqd/http.go
```diff
@@ -99,11 +100,23 @@ func newHTTPServer(ctx *context, tlsEnabled bool, tlsRequired bool) *httpServer
 	router.Handler("GET", "/debug/pprof/heap", pprof.Handler("heap"))
 	router.Handler("GET", "/debug/pprof/goroutine", pprof.Handler("goroutine"))
 	router.Handler("GET", "/debug/pprof/block", pprof.Handler("block"))
+	router.HandlerFunc("PUT", "/debug/blockrate",
```
mreiferson: let's name this `/debug/setblockrate` and keep this on one line

mreiferson: also, let's log this, something like:

```go
router.Handle("PUT", "/debug/setblockrate", http_api.Decorate(s.setBlockRateHandler, log, http_api.PlainText))
```

nsqd/http.go
```diff
@@ -99,11 +100,23 @@ func newHTTPServer(ctx *context, tlsEnabled bool, tlsRequired bool) *httpServer
...
 	router.Handler("GET", "/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
 
 	return s
 }
 
+func HandleBlockRate(w http.ResponseWriter, req *http.Request) {
+	rate, err := strconv.Atoi(req.FormValue("rate"))
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "block rate value error: %s", err.Error())
```
mreiferson: after this is converted to the custom HTTP handler, we can just:

```go
return nil, http_api.Err{400, fmt.Sprintf("invalid block rate %s", err)}
```

nsqd/http.go
```diff
@@ -99,11 +100,23 @@ func newHTTPServer(ctx *context, tlsEnabled bool, tlsRequired bool) *httpServer
...
+		return
+	}
+	runtime.SetBlockProfileRate(rate)
```
mreiferson: should we return output in the success case?

absolute8511: Since the SetBlockProfileRate has no return value, I think we no need to return output.

nsqd/http.go
```diff
@@ -99,11 +100,21 @@ func newHTTPServer(ctx *context, tlsEnabled bool, tlsRequired bool) *httpServer
 	router.Handler("GET", "/debug/pprof/heap", pprof.Handler("heap"))
 	router.Handler("GET", "/debug/pprof/goroutine", pprof.Handler("goroutine"))
 	router.Handler("GET", "/debug/pprof/block", pprof.Handler("block"))
+	router.Handle("PUT", "/debug/setblockrate", http_api.Decorate(HandleBlockRate, log, http_api.PlainText))
 	router.Handler("GET", "/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
 
 	return s
 }
 
+func HandleBlockRate(w http.ResponseWriter, req *http.Request, ps httprouter.Params) (interface{}, error) {
```
mreiferson: Last comment!  This doesn't need to be exported, let's name it `setBlockRateHandler` please

mreiferson:
@absolute8511 thanks for this - after that last naming change can you squash down to one commit please?

absolute8511:
@mreiferson done.

mreiferson:
Thanks!

PR #709, opened 2016-01-15
nsqd: optimize GetTopic / fix data race

absolute8511:
In `AggregateChannelE2eProcessingLatency` the channels should be copied out with lock and this lock should be outside the stream merge. Also the GetTopic is hot while pub message, so we should avoid write lock as more as possible.

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -430,9 +430,17 @@ func (n *NSQD) Exit() {
 // GetTopic performs a thread safe operation
 // to return a pointer to a Topic object (potentially new)
 func (n *NSQD) GetTopic(topicName string) *Topic {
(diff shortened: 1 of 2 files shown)


jehiah:
Good catch on the race in `AggregateChannelE2eProcessingLatency`. Do you have more changes you want to make in this PR? 

Can you add a bench test and share before/after metrics for performance change from the `GetTopic()` hot path?

absolute8511:
OK, I will add some benchmark test later.

mreiferson:
also LGTM, I am interested in benchmarks but can you also squash down into two separate commits

mreiferson:
thanks for all these fixes @absolute8511

mreiferson:
:+1:

PR #710, opened 2016-01-15
should allow POST method, since the newer go tool pprof need it.

absolute8511:


jehiah:
@absolute8511 is `POST` only needed for `/debug/pprof/symbol` not the other endpoints?

absolute8511:
That's right, it is needed by the go tool along with go 1.5

mreiferson:
cool, LGTM

PR #705, opened 2016-01-10
nsq_to_file: rewrite function hasArg

KerwinKoo:
Recently I'm reading the source code of NSQ.I think using `flag.Visit` is a good idea to implement the `hasArg` func.

PR #712, opened 2016-01-16
nsq_apps: update hasArg function

KerwinKoo:
I'm realy not good at using GitHub,so I reforked `nsq` replace #705 

    Recently I'm reading the source code of NSQ.
    I think using flag.Visit is a good idea to implement the hasArg func.


mreiferson:
thanks, fixed the build in #713

PR #718, opened 2016-02-19
nsqd/nsqlookupd: support running as windows service

judwhite:
This approach expunges the `main.go` and `main_windows.go` files, keeping only the original `nsqd.go` and `nsqlookupd.go` which can be built on all platforms.

The `// +build` responsibility is moved to the `github.com/judwhite/go-svc/svc` package. This package exposes two interfaces and a single `Run` function. Replaces #676.

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -7,13 +7,13 @@ import (
 	"log"
 	"math/rand"
 	"os"
-	"os/signal"
+	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/BurntSushi/toml"
+	"github.com/judwhite/go-svc/svc"
 	"github.com/mreiferson/go-options"
 	"github.com/nsqio/nsq/internal/app"
 	"github.com/nsqio/nsq/internal/version"
@@ -174,7 +174,26 @@ func (cfg config) Validate() {
 	}
 }
 
+type program struct {
+	nsqd *nsqd.NSQD
+}
+
 func main() {
+	prg := &program{}
+	if err := svc.Run(prg); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (p *program) Init(env svc.Environment) error {
+	if env.IsWindowsService() {
+		dir := filepath.Dir(os.Args[0])
+		return os.Chdir(dir)
+	}
+	return nil
+}
+
+func (p *program) Start() error {
 	flagSet := nsqFlagset()
 	flagSet.Parse(os.Args[1:])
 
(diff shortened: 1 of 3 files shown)


mreiferson:
this is beautiful

apps/nsqd/nsqd.go
```diff
@@ -182,12 +201,9 @@ func main() {
 
 	if flagSet.Lookup("version").Value.(flag.Getter).Get().(bool) {
 		fmt.Println(version.String("nsqd"))
-		return
+		os.Exit(0)
 	}
 
-	signalChan := make(chan os.Signal, 1)
-	signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)
```
mreiferson: It's fine for now, but perhaps users of `go-svc` want to specify what signals they want to handle?  Right now it's hardcoded https://github.com/judwhite/go-svc/blob/master/svc/svc_other.go#L19

mreiferson:
LGTM, any concerns about this approach @jehiah?

This should also make it trivial to package official windows binaries, if you want to update the builds in `dist.sh` @judwhite ?

judwhite:
I tried running `dist.sh` unmodified on CentOS 6.7. I'm getting a test failure (tried 3x) and `mktemp` is grumpy. Happen to know what's causing this? If not I'll look into it.

```
nsqadmin/http_test.go#TestHTTPEmptyChannelPOST

        http_test.go:53: http_test.go:490:
                        exp: 1
                        got: 0
```

```
... building v0.3.7-alpha for linux/amd64
mktemp: too few X's in template `nsq'
```

jehiah:
:thumbsup:

mreiferson:
> Happen to know what's causing this? If not I'll look into it.

I've seen that test flake out in travis, too.  It's probably time dependent :(

re: `mktemp` - I only ever run it from OSX, so it might depend on BSD's `mktemp`?

ploxiln:
Indeed, OS X / FreeBSD `mktemp` and linux `mktemp` are slightly different. Here's a way to get the same result on either platform:

```
mktemp -d ${TMPDIR:-/tmp}/nsq.XXXXXX
```

Less ugly linux-only variants, if you're curious (the former does something slightly different on OS X, and the latter doesn't work on OS X):

```
mktemp -d -t nsq.XXXXXX
mktemp --tmpdir nsq.XXXXXX
```

mreiferson:
@ploxiln to the rescue

judwhite:
@mreiferson I'm using `/vendor` in `go-svc` so you'll need to build with Go 1.6 or Go 1.5 with `GO15VENDOREXPERIMENT=1` when targeting `windows`.

If you want to build with older versions of Go I think adding `golang.org/x/sys/windows` to `Godeps` should be enough.

mreiferson:
hmmm, was thinking about building the next release with Go 1.6 but that might be a little aggressive.

mreiferson:
@judwhite just tested `dist.sh` w/ `windows`, this LGTM

I'll open a separate PR for those updates + the flakey test, thanks!

judwhite:
Tried out the official release on Windows, only thing is the binaries need to have a `.exe` extension. After renaming everything works :100:. Thanks for your patience with this one.

PR #722, opened 2016-02-26
nsqadmin: fix connected durations >1hr

jehiah:
I noticed that connected durations longer than 1hr were not displaying correctly in nsqadmin.

cc: @mreiferson


mreiferson:
whooops

mreiferson:
don't we need to push up the built assets and recompile the `go-bindata`?

PR #730, opened 2016-04-09
nsqd: persisted data repeatedly consumed

mreiferson:
hi，
i have a question about nsqd persisted meta file update.
my nsqd version:
> nsqd v0.3.7 (built w/go1.6)

when nsqd start up, it load persisted data from disk, and then subscriber consumed the msg, but the nsqd not update the persist meta file, and then if nsqd crash when it restart, my subscriber will consume the same messages again. I know nsqd guarantees "messages are delivered at least onceAnchor link for: messages are delivered at least once"， but is this scene can improve？

### DIFF
--- a/nsqd/diskqueue.go
+++ b/nsqd/diskqueue.go
@@ -577,7 +577,6 @@ func (d *diskQueue) ioLoop() {
 	for {
 		// dont sync all the time :)
 		if count == d.syncEvery {
-			count = 0
 			d.needSync = true
 		}
 
@@ -586,6 +585,7 @@ func (d *diskQueue) ioLoop() {
 			if err != nil {
 				d.logf("ERROR: diskqueue(%s) failed to sync - %s", d.name, err)
 			}
+			count = 0
 		}
 
 		if (d.readFileNum < d.writeFileNum) || (d.readPos < d.writePos) {
@@ -607,6 +607,7 @@ func (d *diskQueue) ioLoop() {
 		// the Go channel spec dictates that nil channel operations (read or write)
 		// in a select are skipped, we set r to d.readChan only when there is data to read
 		case r <- dataRead:
+			count++
 			// moveForward sets needSync flag if a file is removed
 			d.moveForward()
 		case <-d.emptyChan:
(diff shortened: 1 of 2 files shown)


mreiferson:
RFR

nsqd/diskqueue.go
```diff
@@ -569,15 +569,15 @@ func (d *diskQueue) handleReadError() {
...
+	var wcount int64
 	var r chan []byte
 
 	syncTicker := time.NewTicker(d.syncTimeout)
 
 	for {
 		// dont sync all the time :)
-		if count == d.syncEvery {
-			count = 0
+		if wcount == d.syncEvery {
```
jehiah: don't we want `(wcount + rcount) == d.syncEvery` ? 

And more generally, do we need to distinguish between read and write counts, can we always use a single `count` var?

mreiferson: > can we always use a single count var?

That should simplify things.

mreiferson:
ready @jehiah

jinhao:
@mreiferson  
can release a new version of nsq

I try to get the latest code from github, and rebuild the nsq
but the nsqd cannot commuicate with nsqlookup
it seems new nsqd use protocol 'V2', but nsqlookupd still use 'V1'
nsqlookupd log

    [nsqlookupd] 2016/04/11 10:01:01.732021 CLIENT(127.0.0.1:63929): desired protocol magic '  V1'
    [nsqlookupd] 2016/04/11 10:01:01.732123 ERROR: [127.0.0.1:63929] - E_BAD_BODY IDENTIFY    missing fields
    [nsqlookupd] 2016/04/11 10:01:01.732179 CLIENT(127.0.0.1:63929): closing
    [nsqlookupd] 2016/04/11 10:01:01.732192 ERROR: client(127.0.0.1:63929) - E_BAD_BODY IDENTIFY missing fields

judwhite:
@jinhao That looks like you have `nsqd -lookupd-tcp-address` set to connect to the nsqlookupd HTTP port (4161). Can you try 4160 instead? If that's not the case can you paste your `nsqd` and `nsqlookupd` args?

PR #736, opened 2016-04-13
bugfix and set default value to: worker-id, broadcast-address, tls-min-version...

FlamingTree:
Though `nsqd.NewOptions` set default to some options(worker-id, broadcast-address, tls-min-version, etc), According to logic of `go-options.Resolve`,  eventually they will be reset by default flag value(if option not set in command line and config file), so should set default value to these option on `nsqFlagSet`. It seems `nsqd.NewOptions`'s default value has no effect.

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -75,13 +75,14 @@ func (t *tlsVersionOption) String() string {
 }
 
 func nsqFlagset() *flag.FlagSet {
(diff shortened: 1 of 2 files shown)


FlamingTree:
It seems directory `nsqd/test/certs`'s certs expired, so travis ci failed. can you update these files?

mreiferson:
see #738, thanks!

PR #738, opened 2016-04-15
nsqd: update and fix for latest go-options

mreiferson:
this is a replacement for #736, thanks @FlamingTree!


mreiferson:
Actually cleans this up a bunch, RFR @jehiah

nsqd/options.go
```diff
@@ -75,61 +71,18 @@ type Options struct {
...
 		QueueScanDirtyPercent:    0.25,
 
-		MsgTimeout:    60 * time.Second,
-		MaxMsgTimeout: 15 * time.Minute,
-		MaxMsgSize:    1024 * 1024,
-		MaxBodySize:   5 * 1024 * 1024,
-		MaxReqTimeout: 1 * time.Hour,
 		ClientTimeout: 60 * time.Second,
 
-		MaxHeartbeatInterval:   60 * time.Second,
```
jehiah: removing these presents a challenge for those importing `nsqd` directly though right?

jehiah: Thinking on this a little more I think it'd be better to preserve the defaults here in `NewOptions()` and just change the order in `main()`  to get `NewOptions()` first and pass that in to `nsqFlagset()`, so that the defaults come from the `nsqd` package, not `apps/nsqd`

mreiferson: Good points, agreed, will update.

mreiferson:
updated

jehiah:
@mreiferson can you look into why CI isn't happy yet? (LGTM otherwise)

mreiferson:
let's see how this run does

jehiah:
:+1: LGTM. squash?

mreiferson:
:hammer:d

PR #742, opened 2016-04-16
nsqd: lock tweaks

mreiferson:
(Pushing up rebased branches I've had laying around locally for ages)


mreiferson:
Obviated by #743

PR #751, opened 2016-04-22
dist.sh: strip release builds of debug info

ploxiln:
results in 30% smaller binaries

backtraces should still be fine
inspired by / see also: https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/


ploxiln:
unrelated ... it would be nice if version was bumped to something like "0.3.8-pre" just after tagging 0.3.7

mreiferson:
word

PR #766, opened 2016-06-01
docker: Adds support for alpine as base image

kenjones-cisco:
Adds a variant for NSQ leveraging alpine as the base image.

Alpine is equally as small as busybox (~5 MB) but has more features
including the ability to install other packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM busybox
-
-ADD dist/docker/bin/ /nsq_bin/
-RUN cd /    && ln -s /nsq_bin/* . \
- && cd /bin && ln -s /nsq_bin/* .
+FROM alpine:3.4
 
 EXPOSE 4150 4151 4160 4161 4170 4171
 
 VOLUME /data
 VOLUME /etc/ssl/certs
+
+COPY dist/docker/bin/ /usr/local/bin/
+RUN ln -s /usr/local/bin/*nsq* / \
+    && ln -s /usr/local/bin/*nsq* /bin/


kenjones-cisco:
busybox is unfortunately very limited as there is no option for curl and wget is all but broken. So the ability to extend the base nsqio/nsq image is painful. This provides an equally simple base image that leverages busybox binaries internally while providing the option for more if needed. 

Many of the major official images within Docker Hub are now including an alpine variant or using alpine exclusively.

To maintain backwards compatibility with what existed, I just added a separate docker file and updated the dist to build a new image with a different tag indicating alpine.

mreiferson:
I would be amenable to switching the official image over to alpine, what do you think @jehiah?

cc @ploxiln

jehiah:
sounds good to me

kenjones-cisco:
Should I update the PR to replace the existing Dockerfile then?

mreiferson:
Yes please!

Dockerfile
```diff
@@ -1,10 +1,8 @@
-FROM busybox
-
-ADD dist/docker/bin/ /nsq_bin/
-RUN cd /    && ln -s /nsq_bin/* . \
- && cd /bin && ln -s /nsq_bin/* .
```
mreiferson: We can't drop the symlinks for backwards compat

mreiferson:
great, thanks!

Do you mind squashing this down to one commit?

kenjones-cisco:
@mreiferson done

mreiferson:
LGTM

PR #776, opened 2016-07-24
nsqd: configurable HTTP client transport timeout

### LINKED ISSUE #680: nsq_to_http: improve --http-timeout
by default http-timeout(20s), on the same deadlineConn(api_request.go), the second request's timeout is less than 20s.
api_request.go:
```go
func (c *deadlinedConn) Read(b []byte) (n int, err error) {
	// it is not correct to set deadline on read operation, 
	// http connection doesn't support reading timeout
	// instead, we can set  http.Transport.ResponseHeaderTimeout.
	c.Conn.SetReadDeadline(time.Now().Add(c.Timeout))
	return c.Conn.Read(b)
}

func (c *deadlinedConn) Write(b []byte) (n int, err error) {
	// ok, it's right
	c.Conn.SetWriteDeadline(time.Now().Add(c.Timeout))
	return c.Conn.Write(b)
}

func NewDeadlineTransport(timeout time.Duration) *http.Transport {
	transport := &http.Transport{
		Dial: func(netw, addr string) (net.Conn, error) {
			c, err := net.DialTimeout(netw, addr, timeout)
			if err != nil {
				return nil, err
			}
			return &deadlinedConn{timeout, c}, nil
		},
		//  http reading timeout is here
		ResponseHeaderTimeout: timeout,
	}
	return transport
}
```

### LINKED ISSUE #715: nsqd: configurable HTTP client transport timeout
see https://github.com/nsqio/nsq/pull/683#issuecomment-155947122

kenjones-cisco:
Adds configuration options HTTPClientConnectTimeout and
HTTPClientRequestTimeout to control the connection and request timeout
repectively of the HTTP client.

Also added to the following app binaries:

- nsqadmin
- nsq_stat
- nsq_to_file
- nsq_to_http

Closes #715
Closes #680

### DIFF
--- a/apps/nsq_stat/nsq_stat.go
+++ b/apps/nsq_stat/nsq_stat.go
@@ -22,14 +22,16 @@ import (
 )
 
 var (
-	showVersion      = flag.Bool("version", false, "print version")
-	topic            = flag.String("topic", "", "NSQ topic")
-	channel          = flag.String("channel", "", "NSQ channel")
-	statusEvery      = flag.Duration("status-every", -1, "(deprecated) duration of time between polling/printing output")
-	interval         = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
-	countNum         = numValue{}
-	nsqdHTTPAddrs    = app.StringArray{}
-	lookupdHTTPAddrs = app.StringArray{}
+	showVersion        = flag.Bool("version", false, "print version")
+	topic              = flag.String("topic", "", "NSQ topic")
+	channel            = flag.String("channel", "", "NSQ channel")
+	statusEvery        = flag.Duration("status-every", -1, "(deprecated) duration of time between polling/printing output")
+	interval           = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
+	httpRequestTimeout = flag.Duration("http-client-request-timeout", 5*time.Second, "timeout for HTTP request")
+	countNum           = numValue{}
+	nsqdHTTPAddrs      = app.StringArray{}
+	lookupdHTTPAddrs   = app.StringArray{}
 )
 
 type numValue struct {
@@ -55,9 +57,9 @@ func init() {
 	flag.Var(&countNum, "count", "number of reports")
 }
 
-func statLoop(interval time.Duration, topic string, channel string,
-	nsqdTCPAddrs []string, lookupdHTTPAddrs []string) {
-	ci := clusterinfo.New(nil, http_api.NewClient(nil))
+func statLoop(interval time.Duration, connectTimeout time.Duration, requestTimeout time.Duration,
+	topic string, channel string, nsqdTCPAddrs []string, lookupdHTTPAddrs []string) {
+	ci := clusterinfo.New(nil, http_api.NewClient(nil, connectTimeout, requestTimeout))
 	var o *clusterinfo.ChannelStats
 	for i := 0; !countNum.isSet || countNum.value >= i; i++ {
 		var producers clusterinfo.Producers
@@ -149,6 +151,16 @@ func main() {
 		log.Fatal("--interval should be positive")
 	}
 
+	connectTimeout := *httpConnectTimeout
+	if int64(connectTimeout) <= 0 {
+		log.Fatal("--http-client-connect-timeout should be positive")
+	}
+
+	requestTimeout := *httpRequestTimeout
+	if int64(requestTimeout) <= 0 {
+		log.Fatal("--http-client-request-timeout should be positive")
+	}
+
 	if countNum.isSet && countNum.value <= 0 {
 		log.Fatal("--count should be positive")
 	}
@@ -171,7 +183,7 @@ func main() {
 	termChan := make(chan os.Signal, 1)
 	signal.Notify(termChan, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
 
-	go statLoop(intvl, *topic, *channel, nsqdHTTPAddrs, lookupdHTTPAddrs)
+	go statLoop(intvl, connectTimeout, requestTimeout, *topic, *channel, nsqdHTTPAddrs, lookupdHTTPAddrs)
 
 	<-termChan
 }
--- a/apps/nsq_to_file/nsq_to_file.go
+++ b/apps/nsq_to_file/nsq_to_file.go
@@ -45,6 +45,9 @@ var (
 	rotateSize     = flag.Int64("rotate-size", 0, "rotate the file when it grows bigger than `rotate-size` bytes")
 	rotateInterval = flag.Duration("rotate-interval", 0*time.Second, "rotate the file every duration")
 
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
+	httpRequestTimeout = flag.Duration("http-client-request-timeout", 5*time.Second, "timeout for HTTP request")
+
 	nsqdTCPAddrs     = app.StringArray{}
 	lookupdHTTPAddrs = app.StringArray{}
 	topics           = app.StringArray{}
@@ -404,8 +407,9 @@ func (t *TopicDiscoverer) allowTopicName(pattern string, name string) bool {
 	return match
 }
 
-func (t *TopicDiscoverer) syncTopics(addrs []string, pattern string) {
-	newTopics, err := clusterinfo.New(nil, http_api.NewClient(nil)).GetLookupdTopics(addrs)
+func (t *TopicDiscoverer) syncTopics(addrs []string, pattern string,
+	connectTimeout time.Duration, requestTimeout time.Duration) {
+	newTopics, err := clusterinfo.New(nil, http_api.NewClient(nil, connectTimeout, requestTimeout)).GetLookupdTopics(addrs)
 	if err != nil {
 		log.Printf("ERROR: could not retrieve topic list: %s", err)
 	}
@@ -438,13 +442,14 @@ func (t *TopicDiscoverer) hup() {
 	}
 }
 
-func (t *TopicDiscoverer) watch(addrs []string, sync bool, pattern string) {
+func (t *TopicDiscoverer) watch(addrs []string, sync bool, pattern string,
+	connectTimeout time.Duration, requestTimeout time.Duration) {
 	ticker := time.Tick(*topicPollRate)
 	for {
 		select {
 		case <-ticker:
 			if sync {
-				t.syncTopics(addrs, pattern)
+				t.syncTopics(addrs, pattern, connectTimeout, requestTimeout)
 			}
 		case <-t.termChan:
 			t.stop()
@@ -472,6 +477,16 @@ func main() {
 
 	if *channel == "" {
 		log.Fatal("--channel is required")
+	}
+
+	connectTimeout := *httpConnectTimeout
+	if int64(connectTimeout) <= 0 {
+		log.Fatal("--http-client-connect-timeout should be positive")
+	}
+
+	requestTimeout := *httpRequestTimeout
+	if int64(requestTimeout) <= 0 {
+		log.Fatal("--http-client-request-timeout should be positive")
 	}
 
 	var topicsFromNSQLookupd bool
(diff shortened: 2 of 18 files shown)


mreiferson:
@kenjones-cisco thanks for this!

I have a high level comment first.

If we're going to land this, we should probably just do #680 as well, because the timeout as implemented is really a read/write deadline on the connection and has unexpected behavior.

If we do #680, we should introduce two configurable values, `--http-client-connect-timeout` and `--http-client-request-timeout`. They will apply to the custom `Dial` and to the [HTTP client's](https://golang.org/pkg/net/http/#Client) `Timeout` property, respectively.

What do you think?

kenjones-cisco:
@mreiferson That makes sense. Let me take a look.

mreiferson:
Also, let's not forget to update `nsqd`'s example config file with these options.

kenjones-cisco:
@mreiferson changes made, just saw the comment about the example config file. Based on #715 I thought it was command line options. For nsqd and nsqadmin, the options should be specified for the configuration file as well?

mreiferson:
> For nsqd and nsqadmin, the options should be specified for the configuration file as well?

Might only be `nsqd` - it uses [`go-options`](https://github.com/mreiferson/go-options) to resolve configuration coming from multiple sources (the struct's defaults, command line flags, and config files).

internal/http_api/api_request.go
```diff
@@ -19,7 +19,6 @@ type deadlinedConn struct {
 }
 
 func (c *deadlinedConn) Read(b []byte) (n int, err error) {
-	c.Conn.SetReadDeadline(time.Now().Add(c.Timeout))
```
mreiferson: we should drop the one in `Write` too

internal/http_api/api_request.go
```diff
@@ -38,6 +37,7 @@ func NewDeadlineTransport(timeout time.Duration) *http.Transport {
 			}
 			return &deadlinedConn{timeout, c}, nil
 		},
+		ResponseHeaderTimeout: requestTimeout,
```
mreiferson: I don't _think_ we need to set this, considering we're bounding the _entire_ request by the same timeout.

apps/nsq_stat/nsq_stat.go
```diff
@@ -27,6 +27,8 @@ var (
 	channel          = flag.String("channel", "", "NSQ channel")
 	statusEvery      = flag.Duration("status-every", -1, "(deprecated) duration of time between polling/printing output")
 	interval         = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
+	httpRequestTimeout = flag.Duration("http-client-request-timeout", 2*time.Second, "timeout for HTTP read/write (each)")
```
mreiferson: The description should be `timeout for HTTP request`, it's not the read/write anymore after this set of changes.

(repeat for all locations of this description)

kenjones-cisco:
@mreiferson  Looks like the DiskQueue test failure is back; otherwise all changes based on feedback are completed.

apps/nsq_stat/nsq_stat.go
```diff
@@ -27,6 +27,8 @@ var (
 	channel          = flag.String("channel", "", "NSQ channel")
 	statusEvery      = flag.Duration("status-every", -1, "(deprecated) duration of time between polling/printing output")
 	interval         = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
+	httpRequestTimeout = flag.Duration("http-client-request-timeout", 2*time.Second, "timeout for HTTP request")
```
mreiferson: Didn't think of this before, but now that we're setting a real request timeout, which encapsulates connection time, we should set the default for request timeout something greater than connect, say `5s`?

cc @jehiah

kenjones-cisco:
@mreiferson Just let me know when a default is decided.

mreiferson:
@kenjones-cisco let's go with `5s` for now, @jehiah can chime in if he wants.

kenjones-cisco:
@mreiferson all done

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -42,9 +42,12 @@ var (
 	numPublishers = flag.Int("n", 100, "number of concurrent publishers")
 	mode          = flag.String("mode", "hostpool", "the upstream request mode options: multicast, round-robin, hostpool (default), epsilon-greedy")
 	sample        = flag.Float64("sample", 1.0, "% of messages to publish (float b/w 0 -> 1)")
-	httpTimeout   = flag.Duration("http-timeout", 20*time.Second, "timeout for HTTP connect/read/write (each)")
-	statusEvery   = flag.Int("status-every", 250, "the # of requests between logging status (per handler), 0 disables")
-	contentType   = flag.String("content-type", "application/octet-stream", "the Content-Type used for POST requests")
+	// deprecate? in favor of http-client-connect-timeout, http-client-request-timeout
```
mreiferson: Good catch, can you put the `// TODO: ` prefix on this so it shows up in searches?

apps/nsq_stat/nsq_stat.go
```diff
@@ -22,14 +22,16 @@ import (
...
-	interval         = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
-	countNum         = numValue{}
-	nsqdHTTPAddrs    = app.StringArray{}
-	lookupdHTTPAddrs = app.StringArray{}
+	showVersion        = flag.Bool("version", false, "print version")
+	topic              = flag.String("topic", "", "NSQ topic")
+	channel            = flag.String("channel", "", "NSQ channel")
+	statusEvery        = flag.Duration("status-every", -1, "(deprecated) duration of time between polling/printing output")
+	interval           = flag.Duration("interval", 2*time.Second, "duration of time between polling/printing output")
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 5*time.Second, "timeout for HTTP connect")
```
mreiferson: I meant for the connect timeout to remain at `2s` and bump request timeout to `5s`. It doesn't really make sense for them to be equal to each other.

contrib/nsqd.cfg.example
```diff
@@ -21,6 +21,11 @@ nsqlookupd_tcp_addresses = [
     "127.0.0.1:4160"
 ]
 
+## duration to wait before HTTP client connection timeout
+http_client_connect_timeout = "2s"
+
+## duration to wait before HTTP client request timeout
+http_client_request_timeout = "2s"
```
mreiferson: missed a `5s` bump

internal/http_api/api_request.go
```diff
@@ -46,12 +44,13 @@ type Client struct {
 	c *http.Client
 }
 
-func NewClient(tlsConfig *tls.Config) *Client {
-	transport := NewDeadlineTransport(2 * time.Second)
+func NewClient(tlsConfig *tls.Config, clientTimeout time.Duration, requestTimeout time.Duration) *Client {
```
mreiferson: `s/clientTimeout/connectTimeout`

internal/http_api/api_request.go
```diff
@@ -19,12 +19,10 @@ type deadlinedConn struct {
 }
 
 func (c *deadlinedConn) Read(b []byte) (n int, err error) {
-	c.Conn.SetReadDeadline(time.Now().Add(c.Timeout))
 	return c.Conn.Read(b)
 }
 
 func (c *deadlinedConn) Write(b []byte) (n int, err error) {
-	c.Conn.SetWriteDeadline(time.Now().Add(c.Timeout))
```
mreiferson: Now that we've removed these, we need to set `ResponseHeaderTimeout` to our configured request timeout on the `http.Transport` instantiation below on line 30. `NewDeadlineTransport` will need to take both connect timeout and request timeout.

mreiferson: (otherwise `DeadlineTransport` won't actually have request timeouts)

kenjones-cisco: I had that previously, and your previous comments said it shouldn't be needed as entire request is now being bounded.

mreiferson: Right, sorry about the back and forth. What I realized is that, when `DeadlineTransport` is used for proxying (`nsqadmin`) there is no other request timeout being applied (unlike when it's used with an HTTP client).

We could set the `ResponseHeaderTimeout` only in the case where it's used for proxying?

kenjones-cisco:
@mreiferson Let me know if there is any more feedback.

apps/nsq_to_file/nsq_to_file.go
```diff
@@ -45,6 +45,9 @@ var (
 	rotateSize     = flag.Int64("rotate-size", 0, "rotate the file when it grows bigger than `rotate-size` bytes")
 	rotateInterval = flag.Duration("rotate-interval", 0*time.Second, "rotate the file every duration")
 
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 5*time.Second, "timeout for HTTP connect")
```
mreiferson: should be `2s`

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -42,9 +42,12 @@ var (
 	numPublishers = flag.Int("n", 100, "number of concurrent publishers")
 	mode          = flag.String("mode", "hostpool", "the upstream request mode options: multicast, round-robin, hostpool (default), epsilon-greedy")
 	sample        = flag.Float64("sample", 1.0, "% of messages to publish (float b/w 0 -> 1)")
-	httpTimeout   = flag.Duration("http-timeout", 20*time.Second, "timeout for HTTP connect/read/write (each)")
-	statusEvery   = flag.Int("status-every", 250, "the # of requests between logging status (per handler), 0 disables")
-	contentType   = flag.String("content-type", "application/octet-stream", "the Content-Type used for POST requests")
+	// TODO: remove; deprecated in favor of http-client-connect-timeout, http-client-request-timeout
+	httpTimeout        = flag.Duration("http-timeout", 20*time.Second, "timeout for HTTP connect/read/write (each)")
+	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 20*time.Second, "timeout for HTTP connect")
```
mreiferson: should be 2s

nsqadmin/http.go
```diff
@@ -83,7 +84,7 @@ func NewHTTPServer(ctx *Context) *httpServer {
 	router.Handle("GET", "/static/:asset", http_api.Decorate(s.staticAssetHandler, log, http_api.PlainText))
 	router.Handle("GET", "/fonts/:asset", http_api.Decorate(s.staticAssetHandler, log, http_api.PlainText))
 	if s.ctx.nsqadmin.getOpts().ProxyGraphite {
-		proxy := NewSingleHostReverseProxy(ctx.nsqadmin.graphiteURL, 20*time.Second)
+		proxy := NewSingleHostReverseProxy(ctx.nsqadmin.graphiteURL, 20*time.Second, 20*time.Second)
```
mreiferson: these should use options?

nsqadmin/nsqadmin.go
```diff
@@ -143,7 +143,7 @@ func (n *NSQAdmin) handleAdminActions() {
 		if err != nil {
 			n.logf("ERROR: failed to serialize admin action - %s", err)
 		}
-		httpclient := &http.Client{Transport: http_api.NewDeadlineTransport(10 * time.Second)}
+		httpclient := &http.Client{Transport: http_api.NewDeadlineTransport(10*time.Second, 10*time.Second)}
```
mreiferson: same here, use options?

nsqlookupd/nsqlookupd_test.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/nsqio/nsq/internal/http_api"
 )
 
+const (
+	ConnectTimeout = 5 * time.Second
```
mreiferson: minor, but 2s for consistency

mreiferson:
@kenjones-cisco found a few more inconsistencies, thanks for your patience!

kenjones-cisco:
Some were previously hard-coded to a specific value so I left those to avoid breaking existing functionality. But I can make the changes.

mreiferson:
> Some were previously hard-coded to a specific value so I left those to avoid breaking existing functionality. But I can make the changes.

@kenjones-cisco appreciate that, and I think we can still accomplish that by preserving those previously hard-coded values as the defaults for the flags, for those applications. What will be new (and potentially unexpected) is that we are now introducing a separate connection timeout in those  cases. I believe `2s` is a reasonable default.

kenjones-cisco:
@mreiferson all changes completed

mreiferson:
LGTM, thanks!

PR #778, opened 2016-07-30
nsqlookupd: improve test coverage

kenjones-cisco:
Adds test cases to improve code coverage to protect against potential
compatibility conflicts moving forward.

Existing test cases refactored to leverage common test helpers within
the internal/test package to avoid code duplication and simplify
the testing effort.

### DIFF
--- a/internal/test/assertions.go
+++ b/internal/test/assertions.go
@@ -17,10 +17,10 @@ func Equal(t *testing.T, expected, actual interface{}) {
 }
 
 func NotEqual(t *testing.T, expected, actual interface{}) {
-	if !reflect.DeepEqual(expected, actual) {
+	if reflect.DeepEqual(expected, actual) {
 		_, file, line, _ := runtime.Caller(1)
-		t.Logf("\033[31m%s:%d:\n\n\tvalue should not equal %#v\033[39m\n\n",
-			filepath.Base(file), line, actual)
+		t.Logf("\033[31m%s:%d:\n\n\tnexp: %#v\n\n\tgot:  %#v\033[39m\n\n",
+			filepath.Base(file), line, expected, actual)
 		t.FailNow()
 	}
 }
--- a/nsqlookupd/options.go
+++ b/nsqlookupd/options.go
@@ -16,7 +16,7 @@ type Options struct {
 	InactiveProducerTimeout time.Duration `flag:"inactive-producer-timeout"`
 	TombstoneLifetime       time.Duration `flag:"tombstone-lifetime"`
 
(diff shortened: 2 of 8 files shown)


kenjones-cisco:
Initial efforts, so far but wanted to start the PR so that it is visible and if there is any feedback on the work being done that it can be shared sooner rather than later.

I'll remove the [WIP] from the title once I think it is ready for merging. All feedback welcome!

internal/test/assertions.go
```diff
@@ -17,10 +17,10 @@ func Equal(t *testing.T, expected, actual interface{}) {
 }
 
 func NotEqual(t *testing.T, expected, actual interface{}) {
-	if !reflect.DeepEqual(expected, actual) {
+	if reflect.DeepEqual(expected, actual) {
```
mreiferson: hah, guess we didn't use that very much 😉

nsqlookupd/options.go
```diff
@@ -16,7 +18,7 @@ type Options struct {
 	InactiveProducerTimeout time.Duration `flag:"inactive-producer-timeout"`
 	TombstoneLifetime       time.Duration `flag:"tombstone-lifetime"`
 
-	Logger logger
+	Logger app.Logger
```
mreiferson: `nsqlookupd` is technically a public package and `app` is now an `internal` one - seems like that might be undesirable?

mreiferson:
@kenjones-cisco thanks, this is really great 😍 ❤️ 💯

nsqlookupd/options.go
kenjones-cisco: Any other idea for providing the common types w/o duplicating the same within each of the public packages?

mreiferson: It's such a small interface, I'm ok with a little copy/paste.

kenjones-cisco:
ok, `logger.go` restored and the use of `internal/app/logger.go` has been removed from `nsqlookupd`

mreiferson:
@kenjones-cisco thanks, we've been trying to slowly remove our use of `go-simplejson` throughout the codebase (finished in #367).

It's my fault for not mentioning this earlier on this PR, but would you mind removing `go-simplejson` here so we don't add more back into the code base? I would understand if you're not up for refactoring this, I can handle it in a rebase of #367 if necessary.

kenjones-cisco:
Sure no problem.

kenjones-cisco:
Removed `go-simplejson` from `nsqlookupd` tests.

mreiferson:
😍👍🏻

PR #781, opened 2016-08-04
nsqadmin: improve test coverage

kenjones-cisco:
Adds test cases to improve code coverage to protect against potential
compatibility conflicts moving forward.

Existing test cases refactored to leverage common test helpers within
the internal/test package to avoid code duplication and simplify
the testing effort.

### DIFF
--- a/nsqadmin/nsqadmin_test.go
+++ b/nsqadmin/nsqadmin_test.go
@@ -1,0 +1,58 @@
+package nsqadmin
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/nsqio/nsq/internal/test"
+)
+
+func TestNoLogger(t *testing.T) {
+	opts := NewOptions()
+	opts.Logger = nil
+	opts.HTTPAddress = "127.0.0.1:0"
+	opts.NSQLookupdHTTPAddresses = []string{"127.0.0.1:4161"}
+	nsqlookupd := New(opts)
+
+	nsqlookupd.logf("should never be logged")
+}
+
+func TestNeitherNSQDAndNSQLookup(t *testing.T) {
+	if os.Getenv("BE_CRASHER") == "1" {
+		opts := NewOptions()
+		opts.Logger = nil
+		opts.HTTPAddress = "127.0.0.1:0"
+		New(opts)
+		return
+	}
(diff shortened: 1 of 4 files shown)


mreiferson:
@kenjones-cisco 😍 

You interested in taking a pass here in fixing `TestDiskQueueSyncAfterRead` flakiness so we can stabilize these runs?

kenjones-cisco:
@mreiferson 
Interestingly enough I've not encountered this issue a single time when testing locally. I don't know if maybe due to the fact I use containers to execute all the tests against Go 1.5 & Go 1.6, or maybe my SSD drive means writes happen fast enough not that the timer never hits. 

I'll see if I can figure out how to reproduce this error locally, then I'll submit a PR to fix it.

(It is rather annoying that it causes each of my PRs to keep showing as failed!)

mreiferson:
Same, it's definitely a timing issue, perhaps there's a better way to write the test but I'm also fine with adjusting timing in the test so that it smooths out the failures.

nsqadmin/nsqadmin_test.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd.Env = append(os.Environ(), "BE_CRASHER=1")
+	err := cmd.Run()
+	test.Equal(t, "exit status 1", fmt.Sprintf("%v", err))
+	if e, ok := err.(*exec.ExitError); ok && !e.Success() {
+		return
+	}
+	t.Fatalf("process ran with err %v, want exit status 1", err)
+}
+
+func TestBothNSQDAndNSQLookup(t *testing.T) {
```
mreiferson: Can you talk through the approach here, it's not immediately obvious how this works?

kenjones-cisco: I based it on this http://stackoverflow.com/a/33404435

Basically it spawns another process to execute the test in and then verifies the exit code is what is expected. Allows for testing scenarios where the outcome is a call to `os.Exit(1)` work as expected.

mreiferson: OK, but what is the test actually testing? With certain params we expect a certain exit code?

kenjones-cisco: Verify that when calling New() without all the required inputs then it will result in `os.Exit(1)`

mreiferson:
👍

Issue #779, opened 2016-08-02
[PROPOSAL]: Leverage OpenAPI Specification for APIs


kenjones-cisco:
Provides clear documentation of each API and how to use them that is not tied to any specific vendor or technology.

There are tools that can consume the spec and generate the server component and the client component. [go-swagger](http://goswagger.io/)

Also tools for displaying the spec as html, and allows for interacting with the API if server is available.

nsqlookupd APIs as example:
https://gist.github.com/kenjones-cisco/5b36a7c54ddd548264178ca657e00bd7

To view an HTML representation of the example nsqlookupd spec provided, copy the yml file into the Swagger Editor to visually see what the documentation could look like.

Tasks:
- [ ] Create API spec for nsqd
- [ ] Create API spec for nsqlookupd
- [ ] Create API spec for nsqadmin
- [ ] Use specs for documentation on http://nsq.io

Considerations:
- [ ] Apply REST API patterns for APIs
- [ ] Evaluate code generation for Server components
- [ ] Evaluate code generation for Client components

References:
- [Open APIs](https://openapis.org/)
- [Specification](https://openapis.org/specification)
- [Swagger Editor](http://editor.swagger.io/#/)

kenjones-cisco:
Screen caps in HTML format

![nsqlookupd-spec-snap1](https://cloud.githubusercontent.com/assets/7967510/17331889/dbec9e54-589a-11e6-9daa-23215f4b7112.PNG)
![nsqlookupd-spec-snap2](https://cloud.githubusercontent.com/assets/7967510/17331891/dd99bc8c-589a-11e6-9fcf-97a94dc2295a.PNG)
![nsqlookupd-spec-snap3](https://cloud.githubusercontent.com/assets/7967510/17331895/e09dccca-589a-11e6-9288-013a198e9d49.PNG)

mreiferson:
It sounds promising for specification and documentation, I'm a bit weary of using it to generate the actual code though. In your experience, if we omitted code generation, would that compromise the effectiveness of this approach?

kenjones-cisco:
The code generation is only one of the possible benefits, there parts of what the code generation does that could be just used directly such that the spec is part of the run time to avoid the dreaded documentation and the code not being in alignment.

That is why I only put that under the Considerations list vs. the task list. I believe after creating the specs for the APIs and making use for the documentations, then the second part could be making the spec an active part of the API implementation (like routing, validation, etc.).

[...]

PR #782, opened 2016-08-04
nsqlookupd: adds API spec

kenjones-cisco:
Documents APIs using OpenAPI Specification standard.
#779

### DIFF
--- a/spec/nsqlookup.yml
+++ b/spec/nsqlookup.yml
@@ -1,0 +1,448 @@
+swagger: '2.0'
+info:
+  version: '0.4.0'
+  title: 'nsqlookupd'
+  description: |
+    nsqlookupd is the daemon that manages topology information. Clients query nsqlookupd to discover nsqd producers for a specific topic and nsqd nodes broadcasts topic and channel information.
+  license:
+    name: Free Public License 1.0.0
+    url: https://raw.githubusercontent.com/nsqio/nsq/master/LICENSE
+  contact:
+    name: nsqio
+    url: http://nsq.io/
+
+schemes:
+  - http
+consumes:
+  - application/json
+produces:
+  - application/json
+
+paths:
+  /ping:
+    get:
+      operationId: pingHandler
+      summary: Monitoring endpoint, should return OK
+      description: |
+        Monitoring endpoint, should return OK
+      tags:
+        - nsqlookupd
+      produces:
+        - text/plain; charset=utf-8
+      responses:
+        200:
+          description: OK
+
+  # TODO: Alternate path based on REST
+  # /topics/{topic}/nodes:
+  /lookup:
+    get:
+      operationId: doLookup
+      summary: Returns a list of producers for a topic
+      description: |
+        Returns a list of producers for a topic
+      tags:
+        - nsqlookupd
+      parameters:
+        - name: topic
+          in: query
+          description: the topic to list producers for
+          type: string
+          required: true
+      responses:
+        200:
+          description: List of producers
+          schema:
+            type: object
+            properties:
+              channels:
+                type: array
+                items:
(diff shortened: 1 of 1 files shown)


spec/nsqlookup.yml
```diff
@@ -0,0 +1,448 @@
...
+  # TODO: Alternate path based on REST
+  # /topics/{topic}/nodes/{node}:
+  #   delete:
+  /topic/tombstone:
+    # TODO: Consider 'delete' vs. 'post'
+    post:
+      operationId: doTombstoneTopicProducer
+      summary: Tombstones a specific producer of an existing topic.
+      description: |
+        Handles scenario of deleting a topic from a subset of nodes due to the additional complexities. Because of the way consumers query nsqlookupd and connect to all producers you enter into race conditions with attempting to remove the information from the cluster and consumers discovering that node and reconnecting (thus pushing updates that the topic is still produced on that node). The solution in these cases is to use “tombstones”. A tombstone in nsqlookupd context is producer specific and lasts for a configurable --tombstone-lifetime time. During that window the producer will not be listed in /lookup queries, allowing the node to delete the topic, propagate that information to nsqlookupd (which then removes the tombstoned producer), and prevent any consumer from re-discovering that node.
```
ploxiln: long lines could use some wrapping
see http://stackoverflow.com/a/21699210/2807852
(link also good for "why I hate yaml")

spec/nsqlookup.yml
```diff
@@ -0,0 +1,448 @@
...
+                type: string
+                default: INVALID_REQUEST
+                enum:
+                  - INVALID_REQUEST
+                  - MISSING_ARG_TOPIC
+        # TODO: 404 TOPIC_NOT_FOUND
+
+  # TODO: Alternate path based on REST
+  # /topics/{topic}/nodes/{node}:
+  #   delete:
```
ploxiln: Does everyone agree on these TODOs? It might be best to keep them out of the spec for now.

I'll also put in my 2c now: I prefer RPC style paths to REST style paths. Treating paths as objects is usually a fantasy that needs to be worked-around, and is also less convenient and less efficient for parsing and routing in other tools (e.g. nginx, grep).

I recently-ish worked on a big-ish team project which involved a "REST API" app, and that's my experience. Ended up a lot messier than just using params for params.

spec/nsqlookup.yml
```diff
@@ -0,0 +1,448 @@
...
+          type: string
+          pattern: '^[\.a-zA-Z0-9_-]+(#ephemeral)?$'
+          minLength: 1
+          maxLength: 64
+          required: true
+      responses:
+        # TODO: Consider 201 or 204
+        200:
+          description: Success
+        # TODO: Consider TOPIC_ALREADY_EXISTS
```
iameugenejo: Can we consider 409 (Conflict) for TOPIC_ALREADY_EXISTS, so we can ignore the response by just inspecting the HTTP status code instead of parsing the body?

spec/nsqlookup.yml
```diff
@@ -0,0 +1,448 @@
...
+          type: string
+          pattern: '^[\.a-zA-Z0-9_-]+(#ephemeral)?$'
+          minLength: 1
+          maxLength: 64
+          required: true
+      responses:
+        # TODO: Consider 201 or 204
+        200:
+          description: Success
+        # TODO: Consider CHANNEL_ALREADY_EXISTS
```
iameugenejo: I'd like to suggest considering 409 (Conflict) for CHANNEL_ALREADY_EXISTS as well.

ploxiln:
Looks like this isn't being updated. Someone can try again later if they are interested in actually using such a thing.

PR #786, opened 2016-08-11
Add clientConn.Close() in tcpServer Handle

JackDrogon:
Add clientConn.Close() in tcpServer Handle. Before clientConn is not closed when leave Handle function from IOLoop with error.

### DIFF
--- a/nsqlookupd/tcp.go
+++ b/nsqlookupd/tcp.go
@@ -12,6 +12,8 @@ type tcpServer struct {
 }
 
 func (p *tcpServer) Handle(clientConn net.Conn) {
+	defer clientConn.Close()
+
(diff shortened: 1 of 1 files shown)


mreiferson:
@JackDrogon thanks for the patch, but I don't think this is necessary.

All code paths in `IOLoop` pass through https://github.com/nsqio/nsq/blob/master/nsqd/protocol_v2.go#L117

JackDrogon:
I saw that, the protocol v2 is used by nsq client and nsqd. But nslookupd and nsqd is also used protocol v1. This had a error.

JackDrogon:
The error appeares in nsqlookupd.

JackDrogon:
OK. Please move to PR #787

PR #787, opened 2016-08-12
nsqlookupd: Add conn.Close() in Tcp IOLoop

JackDrogon:
As #786 , I remove conn.Close() in nsqlookupd tcp IOLoop like nsqd.


mreiferson:
LGTM, thanks!

PR #788, opened 2016-08-18
apps/to_nsq: add --rate to throttle message creation

jehiah:
This adds a `--rate=int` option to `to_nsq` which will throttle the message creation to that rate per second:

```
  -rate int
    	Throttle messages to n/second. 0 to disable
```

This is helpful when you want to load a large dataset into a message driven pipeline w/o causing a backlog of messages to process on the NSQ side.

### DIFF
--- a/apps/to_nsq/to_nsq.go
+++ b/apps/to_nsq/to_nsq.go
@@ -11,7 +11,9 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"sync/atomic"
 	"syscall"
+	"time"
 
 	"github.com/nsqio/go-nsq"
 	"github.com/nsqio/nsq/internal/app"
@@ -20,7 +22,7 @@ import (
 
 var (
 	topic     = flag.String("topic", "", "NSQ topic to publish to")
-	delimiter = flag.String("delimiter", "\n", "character to split input from stdin (defaults to '\n')")
+	delimiter = flag.String("delimiter", "\n", "character to split input from stdin")
 
 	destNsqdTCPAddrs = app.StringArray{}
 )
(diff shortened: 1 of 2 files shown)


apps/to_nsq/to_nsq.go
```diff
@@ -63,11 +66,42 @@ func main() {
 		log.Fatal("--nsqd-tcp-address required")
 	}
 
+	var throttleEnabled bool
+	if *rate >= 1 {
+		throttleEnabled = true
+	}
```
mreiferson:
```go
throttleEnabled := *rate >= 1
```

apps/to_nsq/to_nsq.go
```diff
@@ -63,11 +66,42 @@ func main() {
...
+	balance := int64(1)
+	interval := (1000 * time.Millisecond) / time.Duration(*rate)
```
mreiferson: `time.Second`?

mreiferson:
I feel like there are some not-so-important race conditions, so otherwise LGTM

apps/to_nsq/to_nsq.go
jehiah: LOL. i know what i was thinking, but i'm not going to mention it. 😊

jehiah:
comments addressed. 🔨 'd and RFM

mreiferson:
❤️

judwhite:
Have you looked at [`golang.org/x/time/rate`](https://godoc.org/golang.org/x/time/rate)? It supports rates less than 1/s, saves a goroutine and a few lines of code at the cost of an added dependency.

jehiah:
@judwhite interesting. Nope never seen/noticed that package before.

PR #792, opened 2016-09-08
nsqadmin: fix root ca verification

joshuarubin:
`nsqadmin` needs to set the `tlsCertPool` with the root ca certificate as the `RootCAs` field of `tls.Config` and not `ClientCAs` since it is verifying the `nsqd` server and not validating incoming client connections itself.

Here is the documentation from the `tls` package:

>	// RootCAs defines the set of root certificate authorities
>	// that clients use when verifying server certificates.
>	// If RootCAs is nil, TLS uses the host's root CA set.
>	RootCAs *x509.CertPool
>
>	// ClientCAs defines the set of root certificate authorities
>	// that servers use if required to verify a client certificate
>	// by the policy in ClientAuth.
>	ClientCAs *x509.CertPool


mreiferson:
@joshuarubin thanks! I guess we're missing a test here, any interest in adding one?

joshuarubin:
test added

mreiferson:
It's (obviously) been broken for a long time, but after scanning the code I don't even think it's making `https` requests internally! 😭

joshuarubin:
Anything I can do to help move this along?

mreiferson:
Ahh, I forgot that it "negotiates" HTTPS automatically.

LGTM

PR #805, opened 2016-11-03
nsq: fix requeue accounting

sdbaiguanghe:


mreiferson:
I _think_ this is correct!

sdbaiguanghe:
My point is correct or the original?
All of my colleagues think the **Channel. Messages** should means number of messages published to this channel(**PUB** & **DPUB**)
and **Channel. Requeued** should be the number of **REQ** received.

mreiferson:
Yes, I agree.

Thoughts @jehiah

jehiah:
👍  this means REQ counts are incremented immediately instead of when it's re-added to the queue for processing. That probably makes more intuitive sense anyway.

nsqd/topic.go
```diff
@@ -275,6 +275,7 @@ func (t *Topic) messagePump() {
 				chanMsg.deferred = msg.deferred
 			}
 			if chanMsg.deferred != 0 {
+				atomic.AddUint64(&channel.messageCount, 1)
```
mreiferson: I don't think we want this change, though, because when the deferred message is finally published this will be incremented.

sdbaiguanghe: But when the defered message finally published the messageCount do not increase.
```
    if chanMsg.deferred != 0 {
        atomic.AddUint64(&channel.messageCount, 1)
        channel.StartDeferredTimeout(chanMsg, chanMsg.deferred)
        continue
    }
```

**StartDeferredTimeout()** will  push the message into channel.deferredPQ and when the timeout message poped from channel.deferredPQ it will be published through **doRequeue()**. The messageCount hasn't changed in the process.
I reviewed the code so many times and verified it in practice

sdbaiguanghe: @mreiferson @jehiah  Am i wrong？

mreiferson: Ahhh, I missed that this was an increment for the _channel_ message count. Yes, it looks like you are correct. However, it doesn't make sense that the topic is now responsible for atomically incrementing this channel's counter, we should push this responsibility into the channel logic somewhere.

mreiferson:
@sdbaiguanghe given that these are regressions, would you mind adding test cases for each of these scenarios?

Thanks!

mreiferson:
see #832

PR #808, opened 2016-11-04
nsqd: fix diskqueue benchmark test, maxMsgSize too small

yejingx:
`BenchmarkDiskQueueGet4096` will put invalid messages if maxMsgSize is `1<<10`.


judwhite:
Nice catch. Checking `err` from [here](https://github.com/nsqio/nsq/pull/808/files#diff-a6550fab3b9f698ffc711512486f7d60R561) might've revealed this sooner.

mreiferson:
thanks!

PR #816, opened 2016-11-23
nsqadmin: Fix handling of IPv6 broadcast addresses

### LINKED ISSUE #815: Nsqadmin fails to load node data with IPv6 broadcast addresses
When nsqd instances are configured with IPv6 addresses e.g. `broadcast-address=fdeb:5ba3:18b3::1` then when nsqadmin attempts to load the node stats it does a naive mapping of the broadcast address to an HTTP/TCP address which gives invalid results.  In the above case it would attempt to fetch from `http://fdeb:5ba3:18b3::1:4151/stats` which fails at the URI parsing stage.

Note that if you set the broadcast address to `[fdeb:5ba3:18b3::1]` this fixes nsqadmin behaviour but breaks nsq behaviour on the node as the address gets double-bracketed at some point and ends up trying to connect to `[[fdeb:5ba3:18b3::1]]:4150`.

magnetised:
Fixes #815

### DIFF
--- a/internal/clusterinfo/types.go
+++ b/internal/clusterinfo/types.go
@@ -3,7 +3,9 @@ package clusterinfo
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/clusterinfo/types.go
```diff
@@ -77,11 +78,29 @@ func (p *Producer) Address() string {
 }
 
 func (p *Producer) HTTPAddress() string {
-	return fmt.Sprintf("%s:%d", p.BroadcastAddress, p.HTTPPort)
+	return addressWithPort(p.BroadcastAddress, p.HTTPPort)
```
mreiferson: go has builtins for this, let's use [`net.JoinHostPort`](https://golang.org/pkg/net/#JoinHostPort)

magnetised:
@mreiferson Done. standard library FTW.

mreiferson:
Re-triggered those flakey test runs. In the meantime would you mind squashing down to one commit?

mreiferson:
apparently I don't know how to use git or github anymore ^^^

mreiferson:
@magnetised thanks!

magnetised:
@mreiferson I'd forgotten all about this - thanks for the unintentional email-storm reminder :)

magnetised:
@mreiferson any idea when the next release is?

mreiferson:
> @mreiferson any idea when the next release is?

Trying to get things in shape for one!

magnetised:
@mreiferson great. good luck

PR #825, opened 2016-12-19
Add CORS header to PUB response

ericzhill:
Publishing to NSQ via a web browser directly, AJAX calls fail due to CORS because of no header in the PUB response.  This two line addition adds that header into the response.


judwhite:
I'm 👎 on this, you probably shouldn't publish directly from a browser on the internet. There are a few other ways you could accomplish the goal, you could proxy NSQ with nginx or another web server and add the CORS header there, or you could publish a message to the domain of the site you're on and forward to the NSQ publisher on the server side.

mreiferson:
@ericzhill thanks for the request but I agree with @judwhite, this is the job of a proxy in front of `nsqd`.

PR #828, opened 2016-12-23
nsqd: fixing race with sending messages and topic/channel associations changing

### LINKED ISSUE #826: nsqd: not properly creating channels registered in lookupd on seeing a new topic

stephensearles:
fixes #826

### DIFF
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -226,8 +226,28 @@ func (t *Topic) messagePump() {
 		backendChan = t.backend.ReadChan()
 	}
 
+	doUpdate := func() {
+		chans = chans[:0]
+		for _, c := range t.channelMap {
+			chans = append(chans, c)
+		}
+		if len(chans) == 0 || t.IsPaused() {
+			memoryMsgChan = nil
+			backendChan = nil
+		} else {
+			memoryMsgChan = t.memoryMsgChan
+			backendChan = t.backend.ReadChan()
+		}
+	}
+
 	for {
 		select {
+		case <-t.channelUpdateChan:
+			t.RLock()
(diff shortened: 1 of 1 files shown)


nsqd/topic.go
```diff
@@ -226,30 +226,36 @@ func (t *Topic) messagePump() {
...
+			continue
+
 		case msg = <-memoryMsgChan:
 		case buf = <-backendChan:
 			msg, err = decodeMessage(buf)
 			if err != nil {
 				t.ctx.nsqd.logf("ERROR: failed to decode message - %s", err)
 				continue
 			}
-		case <-t.channelUpdateChan:
```
stephensearles: I think the crux of the problem was actually that the select statement here could possibly never choose this path, or more likely, choose one of the message chan paths before choosing this one.

nsqd/topic.go
```diff
@@ -263,34 +269,49 @@ func (t *Topic) messagePump() {
 			goto exit
 		}
 
-		for i, channel := range chans {
```
stephensearles: and then this loop happens while some other process could be trying to change the set of channels. I'm not sure that's a problem, but putting this under the topic's lock will help make channel operations more atomic.

mreiferson:
@stephensearles thanks for the PR and the investigation.

You're correct that it's theoretically possible for channel creation to be handled out-of-order from message routing (in the Go runtime we trust), but I don't think it's a practical issue worth solving and certainly not at the expense of larger critical sections between locks. More specifically, the edge cases where this would occur imply that other conditions exist that could cause messages to be "lost" to a new channel (e.g. publishing to the topic while creating said channel).

I'm 👎

PR #830, opened 2016-12-28
nsqd: switch to golang/snappy

### LINKED ISSUE #532: nsqd: message stalling with Snappy enabled
Hi,

Root problem:

Most of the messages that go through our system do so rapidly (tens or hundreds of milliseconds) but some messages stall for multiple seconds.  The length of the stall is tied to the nsqd client heartbeat.  

The stalled messages don't really stand out (roughly 6k of json).  We send other large messages that don't seem to stall.

My own investigation so far suggests that this area of the code in `protocol_v2.go` is related:

```
func (p *protocolV2) Send(client *clientV2, frameType int32, data []byte) error {
    client.Lock()

    var zeroTime time.Time
    if client.HeartbeatInterval > 0 {
        client.SetWriteDeadline(time.Now().Add(client.HeartbeatInterval))
    } else {
        client.SetWriteDeadline(zeroTime)
    }

    _, err := util.SendFramedResponse(client.Writer, frameType, data)
    if err != nil {
        client.Unlock()
        return err
    }

    if frameType != frameTypeMessage {
        err = client.Flush()
    }

    client.Unlock()

    return err
}
```

Specifically, I believe that certain messages aren't being flushed properly.  The heartbeat hits the client and then that flushes the stalled messages through.

Have you seen this behavior before?
Is there any reason that a message might get stuck in util.SendFramedResponse?
Is there a reason we don't explicitly flush frameTypeMessage?

mreiferson:
fixes #532

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -4,7 +4,7 @@ github.com/nsqio/go-nsq                 642a3f9935f12cb3b747294318d730f56f4c34b4
 github.com/bitly/go-simplejson          18db6e68d8fd9cbf2e8ebe4c81a78b96fd9bf05a
 github.com/bmizerany/perks/quantile     6cb9d9d729303ee2628580d9aec5db968da3a607
 github.com/mreiferson/go-options        7ae3226d3e1fa6a0548f73089c72c96c141f3b95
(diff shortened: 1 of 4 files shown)


mreiferson:
RFR @jehiah 

P.S. this also has the side effect of enabling CRC checks

PR #831, opened 2016-12-28
nsqd: bump go-options / log when we can't get channels for pre-creation

mreiferson:
Addresses proposal (1) and (2) of #826

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -444,7 +444,10 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	// this makes sure that any message received is buffered to the right channels
 	lookupdHTTPAddrs := n.lookupdHTTPAddrs()
 	if len(lookupdHTTPAddrs) > 0 {
-		channelNames, _ := n.ci.GetLookupdTopicChannels(t.name, lookupdHTTPAddrs)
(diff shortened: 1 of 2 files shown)


mreiferson:
RFR @jehiah 

cc @ploxiln @stephensearls

nsqd/nsqd.go
```diff
@@ -444,7 +444,10 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	// this makes sure that any message received is buffered to the right channels
 	lookupdHTTPAddrs := n.lookupdHTTPAddrs()
 	if len(lookupdHTTPAddrs) > 0 {
-		channelNames, _ := n.ci.GetLookupdTopicChannels(t.name, lookupdHTTPAddrs)
+		channelNames, err := n.ci.GetLookupdTopicChannels(t.name, lookupdHTTPAddrs)
+		if err != nil {
+			n.logf("ERROR: failed to get channels for pre-creation for topic %s - %s", t.name, err)
```
jehiah: Technically we failed to query all lookupd's so we don't know if the result is deficient in any way. We may have still gotten channels for pre-creation from available lookupd's. Rephrase this and/or make different error messages based on `channelNames`?

mreiferson: Fair point, will update 👍

mreiferson: how's this?

PR #832, opened 2016-12-28
nsqd: more intuitive channel message accounting

mreiferson:
Fixes #805


mreiferson:
RFR @jehiah 

cc @sdbaiguanghe

PR #837, opened 2016-12-29
nsqlookupd: pass defaults from opts into flags

### LINKED ISSUE #826: nsqd: not properly creating channels registered in lookupd on seeing a new topic

mreiferson:
This updates `nsqlookupd` so that default values from the options struct
are passed into flag defaults since all flag values implement `flag.Getter`
and the recent change to `go-options` (mreiferson/go-options#15) was insufficient.

RFR @jehiah
cc @ploxiln

Fixes #826

### DIFF
--- a/apps/nsqlookupd/nsqlookupd.go
+++ b/apps/nsqlookupd/nsqlookupd.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path/filepath"
 	"syscall"
-	"time"
 
 	"github.com/BurntSushi/toml"
 	"github.com/judwhite/go-svc/svc"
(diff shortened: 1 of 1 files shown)


ploxiln:
I can confirm that this fixes the default broadcast address for nsqlookupd.

Check out the interface specifier on that ipv6 link-local address (it's pretty cool that this ipv6 stuff works):

> [nsqlookupd] 2016/12/29 01:53:56.504125 200 GET /channels?topic=t1 ([fe80::3e15:c2ff:fec0:28a8%en0]:50923) 9.42µs

I'll separately look into why there was no error printed when no nsqlookupd could be contacted.

mreiferson:
> I'll separately look into why there was no error printed when no nsqlookupd could be contacted.

When `nsqlookupd` returned an empty broadcast address `nsqd` would skip it entirely, see https://github.com/nsqio/nsq/blob/master/nsqd/lookup.go#L188

ploxiln:
Yeah I figured something like that.

Options I was thinking of:

  * log error if there was at least one lookupd address configured, but none could be contacted
  * log error if broadcast address is empty (right where you point out)
  * just use tcp address (without port) if blank broadcast address

mreiferson:
@jehiah 🍏 💚 📗

PR #843, opened 2017-01-01
nsqd: replace missing requeue exit check; cleanup doRequeue

### LINKED ISSUE #760: nsqd: channel, RequeueMessage should check exitFlag before doRequeue
It looks like [this code](https://github.com/nsqio/nsq/blob/bf6cbc3c087e8ed3abd7d7e657da42508c77ad8a/nsqd/channel.go#L360-L362) should check `c.Exiting()`. This changed in https://github.com/nsqio/nsq/commit/de57a276b503550feab246dedd24e618bd79d49e in #701.

What do you think?

mreiferson:
Fixes #760

RFR @jehiah
cc @judwhite

### DIFF
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -277,7 +277,7 @@ func (c *Channel) IsPaused() bool {
 func (c *Channel) PutMessage(m *Message) error {
 	c.RLock()
 	defer c.RUnlock()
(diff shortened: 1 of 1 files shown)


judwhite:
woah, forgot about this. this is why I should open PR's instead of issues 👍

PR #844, opened 2017-01-05
nsqd: remove --worker-id, replace with --node-id

ploxiln:
This is one of the for-1.0 changes contemplated in #741 

This is the minimal-effort approach. If `--worker-id` is specified with an int, parsing options will fail, and the user will hopefully notice the change instead of just specifying "true" or "false" to work-around the failure. Or is backwards-compatibility (with just a "deprecated" warning) desired?


ploxiln:
added another change contemplated in #741, removing the ID from the metadata filename

nsqd/nsqd.go
```diff
@@ -370,6 +379,11 @@ func (n *NSQD) PersistMetadata() error {
 		return err
 	}
 
+	err = os.Remove(oldFileName)
```
jehiah: My concern with this approach is that it breaks the ability to roll back from a nsqd upgrade (old nsqd would not see new file). 

As much as i hate things sticking around, I'd rather see a phased deprecation where one release persists both versions with defined precedence on startup (or requiring equality between two files), and a subsequent release remove the old file.

ploxiln: sounds good. will attempt

apps/nsqd/nsqd.go
```diff
@@ -80,7 +80,8 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Bool("version", false, "print version string")
 	flagSet.Bool("verbose", false, "enable verbose logging")
 	flagSet.String("config", "", "path to config file")
-	flagSet.Int64("worker-id", opts.ID, "unique seed for message ID generation (int) in range [0,4096) (will default to a hash of hostname)")
+	flagSet.Int64("node-id", opts.ID, "unique part for message IDs, (int) in range [0,1024) (default is hash of hostname)")
```
mreiferson: has this range been wrong this whole time? sheesh 😭

ploxiln: Funny, but it was checked correctly in `nsqd.New()` anyway. Maybe everybody just lets the default "hash of hostname" take care of it.

nsqd/nsqd.go
mreiferson: does that mean we should have a point release between where we are and `1.0`?

jehiah: strictly speaking I think we have 3 options w/r/t version promises. We

a) backport a few small deprecation/forwards compatible things that get released as 0.3.9/0.4 and complete removal in 1.0 (documenting the intermediate version requirements for clean transitions) 
b) 1.0 deprecates and provides forwards compatibility w/ removal in 2.0
c) we just cold turkey switch in 1.0 and give strong documentation for the transition risk

ploxiln:
Pushed up change to maintain old and new metadata files in parallel. It wasn't possible for metadata loading to abort nsqd startup before, I changed that so it is ...

ploxiln:
(squashed in minor fixes, left major commits for discussion)

nsqd/nsqd.go
jehiah: an idea: I don't know what it means for windows, but we could just drop a symlink in when we remove the old filename; that way i think we preserve rollback. We could then refuse to run if both are present (and one isn't a symlink) to make sure rolling forward again is clean. The symlink would mean most importantly we only update a single file going forward when persisting the metdata

mreiferson: My preference is (a), I dislike the idea of maintaining compatibility (and this symlink proposal sounds questionable).

That said, I don't think it's necessary to remove the the old file? If we leave it alone, the rollback path becomes easier.

nsqd/nsqd.go
```diff
@@ -267,24 +268,45 @@ type meta struct {
...
+		return nil
+	}
+	if err == nil && errID == nil {
+		if bytes.Compare(data, dataID) != 0 {
+			return fmt.Errorf("metadata in %s and %s do not match (delete one)", fn, fnID)
+		}
+	}
+	if err != nil {
+		fn = fnID
+		data = dataID
```
mreiferson: I've looked at this block multiple times and I still can't really make sense of it.

Is there a simpler way to implement this? What if `LoadMetadata` took a filename and the _caller_ did some of this compatibility logic?

ploxiln: I use err and errID as a shorthand to indicate if fn or fnID exist. If it was any other error it would have aborted/returned already

ploxiln: ... so if `err != nil` here, then `fn` did not exist and `data` is empty, but `fnID` did exist and `dataID` was successfully read into. Below code can get whichever was found using just `fn` and `data`

nsqd/nsqd.go
ploxiln: The concern is that some messages will be processed before the version rollback, so the old metadata file will have stale info, be missing new messages, have references to queue files that don't exist anymore...

nsqd/nsqd.go
ploxiln: all that said I'll be glad to refactor once I figure out what the final strategy will be :)

mreiferson: The location of the comment is misleading, I was referring to the whole new set of changes related to error handling, not this specific line of if block.

re: strategy, it seems like regardless of version choices, we're going to land forward/backward compatible code, so these changes will be necessary in some form.

ploxiln:
The "what to do with the old metadata file" discussion is by default hidden as a "comment on outdated changes" so I'll continue it here. I do kinda like the symlink idea. By the way, on both macOS and Linux you use rename to atomically replace a plain file with a symlink and vice-versa. Older nsqd will read the file via the symlink, save to the .tmp file, then move it over the symlink.

@judwhite does nsqd normally run as "administrator" when run as a service, and is it reasonable to have windows users run it as administrator when run manually? Any thoughts on the windows symlink situation?

ploxiln:
some interesting reading:
https://github.com/golang/go/issues/8914#issuecomment-154285625
https://github.com/golang/go/blob/release-branch.go1.7/src/internal/syscall/windows/syscall_windows.go#L130

go 1.5+ apparently uses `MoveFileEx()` with the appropriate flags on Windows, so we might not need an `atomicRename()` wrapper to do the special thing on windows anymore

http://www.weirdnet.nl/apple/rename.html
macOS rename() had atomicity issues until recently

jehiah:
I would like to make sure we support forwards and backwards and forwards again safely. The  scenario i am concerned about is nuanced, but it's:

a) running nsqd persisting to  nsqd.1.dat
b) upgrade; uses nsqd.dat; some persistence for nsqd.1.dat -> nsqd.dat so rollback works
c) rollback; loads nsqd.1.dat; nsqd.dat is stale
d) roll forward again; nsqd.dat incorrectly used.

I think there are two sane options to support this back and forth scenario safely. da) refuse to startup if both files are present (And old isn't a symlink). db) use the file w/ the newer timestamp (if timestamps are equal; refuse to start if file contents differ)

For me, this is a real aspect of trying deploying a canary build; reverting, and then deploying an updated canary again. I also don't think this is extremely unusual.

mreiferson:
Given that we decided to support --node-id for 1.0, and the main concern is compatibility, the file name change feels unnecessarily complex and not worth it.

Let's just keep the file name as is, it only matters for pedantic correctness.

jehiah:
I just hate having to explain to how to predict what a new hosts ID will be when I see files be attempted to be moved to a new host. I really want to see that go!

ploxiln:
I think of the data-path as a portable package of nsqd state. It doesn't make sense that if you decide to change the node-id, or move the state to a new host (and use automatic node-id), then you have to rename just this one file. It would be nice to clean this up, and now would be the time to start the transition (though not necessarily in _this_ PR).

By the way, I think it would be fine to drop backwards compatibility in a 1.x minor update, I don't think we have to wait until 2.0. This is because 1.0 will be forwards-compatible with that, it would only break compatibility with pre-1.0 - so this is a bit different than supporting/deprecating a flag that you can still use in 1.0.

mreiferson:
OK, those seem rare but reasonable arguments.

I'll leave code review to @jehiah, good catch on the rename changes!

nsqd/nsqd.go
```diff
@@ -365,11 +396,26 @@ func (n *NSQD) PersistMetadata() error {
...
 	f.Close()
 
-	err = atomicRename(tmpFileName, fileName)
+	err = os.Rename(tmpFileName, fileName)
 	if err != nil {
 		return err
 	}
 
+	stat, err := os.Lstat(fileNameID)
+	if err != nil || (stat.Mode()&os.ModeSymlink) == 0 {
```
ploxiln: If you're curious about the spacing around `&`, gofmt does this to emphasize the operator precedence. I guess that means I don't need the parentheses, I can remove them if preferred.
https://github.com/golang/go/issues/12720

nsqd/nsqd_test.go
```diff
@@ -158,6 +165,106 @@ func TestStartup(t *testing.T) {
...
+	new_meta := `
+	{
+	  "topics": [{
+	    "name": "t2", "paused": false,
+	    "channels": [{"name": "c2", "paused": false}]
+	  }],
+	  "version": "1.0.0-alpha"
+	}`
+
+	tmpDir, err := ioutil.TempDir("", "nsq-test-")
```
ploxiln: `TempDir()` does the random-suffix part automatically. Docs say:

> Multiple programs calling TempDir simultaneously will not choose the same directory.

ploxiln:
I tested on Windows 7. Everything seems to work, when nsqd is run as Administrator (I used an Administrator PowerShell, and a windows build of curl to test). When not run as Administrator it fails to PersistMetadata() on startup and aborts, as expected. Still waiting for windows folks to comment on the symlink / Administrator / service / container situation on Windows, to see how acceptable this is. cc @judwhite @elvarb @esiqveland

nsqd/nsqd_test.go
mreiferson: Cool, another nice find/cleanup

mreiferson:
@ploxiln I'm inclined to land this and folks can follow up with opinions if we need to make changes before release.

Want to squash?

ploxiln:
Sounds good.

IMHO these commits are logically split into separate topics, but I have no problem squashing them all together as is preferred. Done.

mreiferson:
@ploxiln nah, I didn't mean down to one, I meant logical. I thought there was one commit that didn't have a prefix that you were planning on cleaning up?

ploxiln:
heh, well, I could squash the tests changes into the metadata change. will do ...

ploxiln:
If this turns out to be a significant problem for Windows users, one idea is to skip the symlink rollback-safety on Windows. But yeah, we'll figure that out later if needed...

judwhite:
I'll pull down this PR and check it out. Running as non-admin is a must have for us (we don't run services as root) ;)

ploxiln:
adobe / symantec / mcafee / comodo etc tend to install high-privilege services to offer priv escalation opportunities to browser content etc, so I wasn't sure ... I guess we'll need a windows mode for this

judwhite:
Running as non-admin on Windows 10, go 1.7.4

```
2017/01/12 17:10:10 ERROR: failed to persist metadata - symlink nsqd.dat nsqd.815.dat.5407116895147581727.tmp: A required privilege is not held by the client.
```

judwhite:
Also "that's just the way it is" (on Windows) http://security.stackexchange.com/questions/10194/why-do-you-have-to-be-an-admin-to-create-a-symlink-in-windows

ploxiln:
see also this blog post I linked in an earlier discussion:
https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/#A3YPFCQzK3UKSCC1.97

> Now in Windows 10 Creators Update, a user (with admin rights) can first enable Developer Mode, and then any user on the machine can run the mklink command without elevating

But to get the same effect from a `CreateSymbolicLink()` call, you need to also pass the (brand-new I assume) flag `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE`.

Anyway, I have another, slightly messier proposal to handle the Windows case, pushing up now.

judwhite:
I would be okay with an upgrade to 1.0 that can't automatically rollback by replacing the binary. If the binary format of the .dat file changed I don't there'd be an expectation of supporting binary-drop rollback for that scenario, unless offered as a utility to rollback the file format (and name).

What does the user have to do to a rollback to pre-1.0, rename the file to include the node id? I might be selfish here but I think a major version upgrade requires reading the docs.

ploxiln:
The only change is to the filename. The attempt is to avoid "silent" (meta)data loss if someone does something rash - handle the common case, abort in other cases. It's definitely arguable how much should be done ... but we were trying to settle on something :)

judwhite:
@ploxiln Works on Win as non-admin. Why is there a `nsqd.dat` and `nsqd.815.dat` on a fresh data path? I could read the code and figure out "why", but users won't do that and probably be confused. At worst visible inefficiency can hurt confidence. I can jump in if you want me involved, or I can sit back and tell you what it looks like from an outside perspective.

ploxiln:
It was deemed simpler to not detect and handle fresh and upgrade as separate cases. Thus the old metadata filename is written even to a fresh data path. One alternative would be to detect whether the old metadata file already existed, and only update/replace in that case. (Another alternative would be to say "RTFM and don't rollback without manual renaming".)

On non-Windows systems, it's just a one-time symlink, which isn't so bad ...

judwhite:
@ploxiln IMO the symlink is kind of ugly regardless of platform. I like the RTFM approach. Tools can help with a rollback.

ploxiln:
miscellaneous responses:

  * I find symlinks to be beautiful
  * It is possible for windows users to create the symlink manually (e.g. with an Administrator powershell or install script) and then nsqd will notice it and do nothing, thus no inefficiency
  * The first version of this PR just deleted the old metadata file
  * I'd like to get @jehiah and @mreiferson to weigh in with their latest opinions

ploxiln:
* I don't think this metadata file is in the most critical performance path - it does not include anything about messages or diskqueue state, just topic and channel names and paused state

judwhite:
> I'd like to get jehiah and mreiferson to weigh in with their latest opinions

Maybe we could schedule some time on IRC or Skype. I'm sure there are other 1.0 topics we could discuss better in real time.

mreiferson:
I'm more concerned about backwards compatibility rather than _being able to rollback_, which seems like an even more rare case that arguably @jehiah can solve on his own with a bit of scripting 😁.

I think in its most simple form, my ideal scenario is that this PR lands the rename changes to `nsqd.dat`, but does so by simple one-time renaming the existing file. If the file does not exist it just writes to `nsqd.dat`.

mreiferson:
We've gone back and forth a bit, and I'm pretty sure @jehiah is out on break and won't be able to respond for at least a week, so I think given the updates from @judwhite about Windows/admin expectations we should just wait for @jehiah to get back since he felt strongly about this.

ploxiln:
@jehiah seems to be back ... any arguments? should I go back to a simpler one-way version?

nsqd/nsqd.go
```diff
@@ -308,12 +337,32 @@ func (n *NSQD) LoadMetadata() {
...
+		return err
+	}
+
+	_, err = f.Write(data)
+	if err != nil {
+		f.Close()
+		return err
+	}
+
+	f.Sync()
```
jehiah: we should handle `Sync()` error here

nsqd/nsqd_test.go
```diff
@@ -23,8 +23,15 @@ const (
 	RequestTimeout = 5 * time.Second
 )
 
+func newMetadataFile(opts *Options) string {
```
jehiah: Can you move these two functions into `nsqd.go` and replace the two separate uses there?

nsqd/nsqd.go
ploxiln: done

nsqd/nsqd_test.go
ploxiln: done

ploxiln:
I threw in another variation of windows support - just deleting the old file and not having "rollback compat" on windows - which might be a good compromise if the biggest windows user of nsq doesn't want the rollback compat file.

If we can agree on something here, I can revert commits as needed and squash back down a bit ...

nsqd/nsqd.go
```diff
@@ -425,19 +425,20 @@ func (n *NSQD) PersistMetadata() error {
 	// if no symlink (yet), race condition:
 	// crash right here may cause next startup to see metadata conflict and abort
 
+	// on Windows, Administrator privs are needed to Symlink, so skip rollback compat
+	if runtime.GOOS == "windows" {
+		if err == nil || !os.IsNotExist(err) {
+			err = os.Remove(fileNameID)
+		}
+		return err
+	}
```
judwhite: @ploxiln if `os.IsNotExist(err)` is `true` you'll return a value for `err`, is that intentional, or would you prefer to return `nil`?

judwhite:
I think parity in the system between OS's is more important than my personal taste. Though we won't get there with symlinks, if there are going to be two files on Linux I'm fine with the same on Windows. I'll defer; whatever the core maintainers think is best.

nsqd/nsqd.go
ploxiln: That's a bug, I was trying to be too concise, nice catch.
Looks like I'll back-out this windows-path variant anyway.

ploxiln:
I threw out the latest windows variant I mentioned yesterday, and squashed everything down as I found appropriate, just now.

mreiferson:
> as there is no easy way to generate the proper node ID that's currently auto generated. That makes it infeasible to implement out of band scripting for a rollback where you move nsqd.dat to the right nsqd.%d.dat for rollback because you don't know the value for %d.

@jehiah what do you mean? It's deterministic, why can't it be scripted? I agree the symlink makes rollback _easier_, obviously.

mreiferson:
If you want the symlink, fine, let's just land this already. My only requirements is that I don't think the version should be `1.0` that includes code that handles symlinks (creation _or_ cleanup).

ploxiln:
So in 1.0, you want it to

  * try to read from new metadata file, if success, done
  * try to read from old metadata file only if new one does not exist
  * write new metadata file (and do nothing else)

mreiferson:
IMO it should read/write to the new metadata file and do nothing else.

ploxiln:
It seems harsh if one wouldn't be able to upgrade from 0.3.8 to 1.0 without either manually renaming the metadata file or losing topics/channels - unless there would be a pre-1.0 transition release series that was around for 6 months.

I think the minimum that makes sense is to just fall back to reading the old metadata file if the new one does not yet exist. That's really not much code. And it's sufficient if you never rollback ...

ploxiln:
For sake of discussion ... I've pushed up a commit that reduces this back down to what I think the minimum for 1.0 should be.

ploxiln:
With the "minimum back-compat" code, the rollback procedure would be something like:

  1. stop nsqd
  2. `[ -e "$DATAPATH"/nsqd.[0-9]*.dat ] || exit 1`
  3. `mv -v "$DATAPATH"/nsqd.dat "$DATAPATH"/nsqd.[0-9]*.dat`
  4. upgrade and start nsqd

I'd also like to point out that with either the symlink-style or minimal-style back-compat code, 1.0 being backward-compatible with 0.x would not necessitate all of 1.x keeping any back-compat code - if 1.0 was ever run it would write the new metadata file, which later 1.x would find even without any back-compat code. So this is not comparable to keeping a function present in a library.

ploxiln:
ping

I think there are two options:

  * remove my last commit, merge with symlink-based rollback-compat, make a pre-1.0 stable release series soon
  * squash my last commit in, merge with one-time-upgrade-compat only, make a 1.0 release soon

mreiferson:
Sorry I wasn't more clear, I meant to land your changes including symlinks and stamp a non-1.0 release, then remove that temporary code and work towards 1.0.

ploxiln:
I removed that last commit. This should be ready to go.

ploxiln:
ping

jehiah:
<3 I heart a 1.0-migration release build for those who want easy rollout.

jehiah:
FYI: i'm working to push this build out on the Bitly stack. I'm using the snippet below to prep our ability to rollback seamlessly (without causing complications rolling forward again)

```
DATA_DIR="/data/nsqd"
if [ -e "$DATA_DIR/nsqd.dat" ] && [ ! -h "$DATA_DIR/nsqd.dat" ]; then
    OLD_NAME=$(find $DATA_DIR -type l -lname "$DATA_DIR/nsqd.dat")
    if [ -n "$OLD_NAME" ]; then
        echo "[`date +'%Y%m%d %H:%M:%S'`] rolling back $DATA_DIR/nsqd.dat back to $OLD_NAME"
        mv $DATA_DIR/nsqd.dat $OLD_NAME || exit 1
    fi
fi
```

mreiferson:
@jehiah you building that with go 1.8 too by any chance?

jehiah:
@mreiferson no unfortunately; this is my last pre-req task i want to finish before i start working on 1.8 on our stack.

mreiferson:
NP, just thinking in terms of a release for these changes I usually use the latest stable version of Go at the time.

PR #847, opened 2017-01-11
nsqd: move diskqueue into nsqio/go-diskqueue

### LINKED ISSUE #764: nsqd: consider publicly exporting diskQueue
First off, thank you for DiskQueue!  It's the best go implementation of a disk-backed channel-esque data structure that I've found!

I have been using it and it's great, but I'm sure you can imagine that hand-applying updates sucks!

Would you please consider making diskQueue publicly exported so it can be used as a "normal" dependency?

In case it is helpful, here is a copy of the ported `DiskQueue` I'm currently using (obviously there is a lot more stuff I ported over, including interfaces and unit-tests, etc):

https://gist.github.com/jaytaylor/71ae326116d9d5492250022025a2015a

Also I am curious - is there a way to get the disk queue to delete or clear out the .dat files?  I've invoked `.Empty()` which reset the metadata, but the .dat file lives on at the same size as before (even though `Depth()==0`).

mreiferson:
This moves the `diskqueue` package into https://github.com/nsqio/go-diskqueue

Fixes #764


ploxiln:
just btw, conflicts with #844

mreiferson:
We'll land this after #844

mreiferson:
rebased @jehiah

PR #850, opened 2017-01-24
*: new --log-prefix flag / option

ploxiln:
instantiate Logger in nsqd.New() instead of nsqd.NewOptions(), only if not already present in opts

More options spam, I understand if it's not desired. But if you don't mind this, I'll do this for nsqlookupd and nsqadmin as well.

(why? it's redundant with the service / container name in my setup. not a big deal)

### DIFF
--- a/apps/nsqadmin/main.go
+++ b/apps/nsqadmin/main.go
@@ -21,6 +21,7 @@ var (
 
 	config      = flagSet.String("config", "", "path to config file")
 	showVersion = flagSet.Bool("version", false, "print version string")
+	logPrefix   = flagSet.String("log-prefix", "[nsqadmin] ", "log message prefix")
 
 	httpAddress = flagSet.String("http-address", "0.0.0.0:4171", "<addr>:<port> to listen on for HTTP clients")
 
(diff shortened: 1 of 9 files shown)


mreiferson:
OK, fine with me, but we should add it for all 3 binaries?

ploxiln:
That's my suggestion, do it for all three - was just testing the waters with one easier-to-review piece first. I'll add changes for the other two shortly ...

ploxiln:
Done for nsqd, nsqlookupd, nsqadmin.

I just realized, previously one would be able to set `Logger` to `nil` in options to disable logging, now that won't work, you'll have to create a no-op Logger. All these apps have `logf()` like:

```go
func (n *NSQD) logf(f string, args ...interface{}) {
	if n.getOpts().Logger == nil {
		return
	}
	n.getOpts().Logger.Output(2, fmt.Sprintf(f, args...))
}
```

(On the upside, we can get rid of that conditional now...)

ploxiln:
squashed in removal of the now (unfortunately) useless nil checks in `logf()`

Issue #737, opened 2016-04-14
nsqadmin: root_url


freeseacher:
Please add support for root_url in nsqadmin. to make it work with reverse proxy better. 
tnx!

mreiferson:
sounds fine to me, PRs welcome!

jhowtan:
Hey we need this too! Seems like there's a PR in for it, https://github.com/nsqio/nsq/pull/856
Could we get this in?

mreiferson:
@jhowtan looks like there's some discussion on that PR about approach. If you're interested in taking a pass at it please do!

mreiferson:
closing in favor of some resolution in #856

PR #856, opened 2017-03-02
nsqadmin: base path

blinklv:
It's difficult for me to access nsqadmin server through nginx (or other reverse proxies), because almost all URLs of HTTP request are based on root path('/'). Most of the time We can't use root path directly, so I add base-path option to change the base path of request URL. 

**eg:**

**command**
`nsqadmin -base-path='/i/am/new/root'`

**original request url**
` /api/topics/foo/bar`

**new request url**
`/i/am/new/root/api/topics/foo/bar`

If you think it's a good idea, you can take it.

(*Maybe I don't know other ways to solve this problem. If you know, please tell me.*)


mreiferson:
cc @jehiah thoughts?

ploxiln:
With a proxy like nginx, you can translate all URLs to remove a common prefix before passing the request through to nsqadmin. The problem is the URLs in the response body, particularly if they're constructed at view time by javascript - but those can work if they're all relative. Currently, all links I've spot checked are absolute paths (but not including domain). If there are any 301 or 302 redirects in nsqadmin, that would be a trickier problem, but I don't think there are any.

So an alternative config-free "fix" (which would need someone to work on it) is to make all nsqadmin links relative. Just mentioning it as an option :)

mreiferson:
Yea, making the links relative would be a simpler set of changes.

@blinklv would you be interested in taking a pass at that implementation?

jehiah:
@mreiferson I'm 👍 for either approach

blinklv:
@mreiferson Of course,  I can try it.

mreiferson:
ping #737

AlmogBaku:
any update on this?

blinklv:
About six months ago, I tried to implement this feature by using *relative links* instead of *root-relative links*.  But the relationship of these sites are more complicated than what I think, the set of changes aren't simpler, so I stick the original way, it has already satisfied my needs. Recently, I simplify some codes of `http.go` file to make it more elegant, which means a simpler set of changes.  😄

MarcErdmann:
Will this be merged and released?

mreiferson:
I started to dig into this one, initially trying to refactor it to use relative links, and I came away feeling similar to @blinklv — that's it's actually easier to just root everything to some base path.

So I think instead I'll just get this rebased and do a bit of cleanup.

mreiferson:
I'm gonna test this out a little more, but I've pushed up my revisions.

ploxiln:
Looks reasonable. I think we'll probably want to defer merging until after v1.1.0 final is released?

(and it'll need to be rebased and bindata regenerated)

nsqadmin/nsqadmin.go
```diff
@@ -130,11 +131,28 @@ func New(opts *Options) *NSQAdmin {
...
+	if len(p) == 0 {
+		return "/"
+	}
+	// add leading slash
+	if p[0] != '/' {
+		p = "/" + p
+	}
+	// add trailing slash
+	if p[len(p)-1] != '/' {
+		p = p + "/"
```
ploxiln: I think the templated urls and such may look a bit better if instead `BasePath` does _not_ end in `/` and we write the `/` after it everywhere. But I understand if you disagree, and it's OK this way too.

mreiferson: Well, except for `/`. I got this from Backbone's code and was trying to create some parity.

I do agree that aesthetically, the sites where we use `basePath` would look better with a preceding `/`. But really it's all just fugly.

blinklv:
I haven't paid my attention to this for a long time. Is there anything I can help now?  (ง •̀_•́)ง

blinklv:
I update my branch (*merge **master** into it*) and hope these changes won't affect your work. @mreiferson

ploxiln:
the rebuild of bindata.go removed over 200 more lines than it added, which is suspicious ...

mreiferson:
It's ok @blinklv, I can handle the rebase and `bindata.go` update! Thanks :)

blinklv:
I added `-debug` option when I first rebuilt `bindata.go` file, so many lines were removed. Now I have replaced it with the released version. @ploxiln 😉

nsqadmin/nsqadmin.go
mreiferson: We should make a decision on this...

mreiferson:
rebased and rebuilt `bindata.go`

nsqadmin/nsqadmin.go
ploxiln: I've noted my preference, and I leave the choice up to you :)

ploxiln: (I think `BasePath` and `base_path` could be blank in the default case, and the root path could be written as `{{BasePath}}/` and `{{base_path}}/`)

ploxiln: (I mean, in the config as `"/"`, but in this function turned into `""` by removing the trailing (and only) slash)

mreiferson:
made this prettier @ploxiln, PTAL

nsqadmin/http.go
```diff
@@ -66,47 +67,51 @@ func NewHTTPServer(ctx *Context) *httpServer {
...
-	router.Handle("GET", "/counter", http_api.Decorate(s.indexHandler, log))
-	router.Handle("GET", "/lookup", http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp(""), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("topics"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("topics/:topic"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("topics/:topic/:channel"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("nodes"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("nodes/:node"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("counter"), http_api.Decorate(s.indexHandler, log))
+	router.Handle("GET", bp("lookup"), http_api.Decorate(s.indexHandler, log))
```
mreiferson: I missed these, they should probably all get the `/` prefix

nsqadmin/static/js/app_state.js
```diff
@@ -30,8 +31,15 @@ var AppState = Backbone.Model.extend({
         this.set('graph_interval', interval);
     },
 
-    url: function(url) {
-        return '/api' + url;
+    basePath: function(p) {
+        // if base path is / then don't prefix
+        var bp = this.get('BASE_PATH') == '/' ? '' : this.get('BASE_PATH');
+        // trim trailing /
+        return (bp + p).replace(/\/$/, '');
```
mreiferson: Need to do a little testing but I _think_ there's a bug here when base path is `/` and this is passed a path `/`

mreiferson: such as https://github.com/nsqio/nsq/pull/856/files#diff-6c66351008f5616a7bf2b9cc9859dbbfR27

mreiferson: Actually, that link is to the Go template function, not the JS one, but I think my original point still stands.

mreiferson:
pushed those fixes up...

nsqadmin/static/js/app_state.js
```diff
@@ -30,8 +31,15 @@ var AppState = Backbone.Model.extend({
         this.set('graph_interval', interval);
     },
 
-    url: function(url) {
-        return '/api' + url;
+    basePath: function(p) {
+        // if base path is / then don't prefix
+        var bp = this.get('BASE_PATH') == '/' ? '' : this.get('BASE_PATH');
+        // remove trailing /, but guarantee at least /
+        return (bp + p).replace(/\/$/, '') || '/';
```
ploxiln: this removes the trailing slash from `p` ... did you intend to remove the trailing slash from `bp`?

ploxiln: There's some unexpected mess and complexity in this approach ... I think it can be simplified to "just remove trailing slash from basepath".

If the base path is `"/"` then we want `""`
If the base path is `"/admin/"` then we want `"/admin"`

Then, you can append anything to that, e.g. `"/topics"`, or just `"/"` to get the root URL path.

ploxiln: I guess this function is pretty much the only problem here, the rest looks good.

In Go land, `path.Join()` takes care of everything after the leading `/`.

So I guess this is the only place where you might have to worry about a double-slash between `BASE_PATH` and `p`.

This logic seems a bit much for what should still be a trivial function. I think my initial suggestion here was appropriate, just remove trailing slash from `BASE_PATH`.

ploxiln: ... e.g.

```javascript
return this.get('BASE_PATH').replace(/\/$/, '') + p;
```

but since this is the only place `BASE_PATH` is used, you could do the regex just once, in `index.html`:

```js
var BASE_PATH = {{basePath ""}}.replace(/\/$/, '');
```

and then here it's just

```js
return this.get('BASE_PATH') + p;
```

mreiferson: The case this is addressing is, e.g. where `BASE_PATH` is `/base` and in one of our handlebars templates we have `{{basePath "/"}}`. What we _want_ is `/base` _not_ `/base/`, but if `BASE_PATH` is `/` we want `/` not `//`.

Basically, the answer to your original question is that I intended to remove any potential trailing slash from the joined path.

Also, FWIW, I think `BASE_PATH` the global var that's passed into `AppState` should always represent the normalized value of whatever was passed to `nsqadmin --base-path`.

If you can come up with an easier way to do this, I'm all ears.

ploxiln: Look at the go function `bp()` in:

```go
router.Handle("GET", bp("/"), http_api.Decorate(s.indexHandler, log))
```

- that code actually gets you `/base/`, so if you wanted `/base`, then it's wrong. Personally, I think `/base/` is good.

---

Notice that your remove-trailing-slash regex _only ever removes a slash from `p`_ because the characters at the end of `(bp + p)` are always from `p`. So the only value of `p` this could be useful for is `p == "/"` in which case you could instead handle it like `if (p == "/") return this.get('BASE_PATH')`. (Recall that `bp` is the user-supplied base path and `p` is the app-meaningful part from our code.)

---

There is a super easy way. Normalize the BasePath to start with a `/`, and then to not end with a `/`, including turning `"/"` into `""`. Now a plain string concatenation works everywhere - in Go, in Javascript, everywhere. No conditional logic, no path.Join(), nothing special. You just have to accept that the root page path is `/base/`, and all the conditionals go away.

nsqadmin/nsqadmin.go
```diff
@@ -130,11 +131,24 @@ func New(opts *Options) *NSQAdmin {
 		}
 	}
 
+	opts.BasePath = normalizeBasePath(path.Clean(opts.BasePath))
```
ploxiln: I didn't initially realize that `path.Clean()` removed the trailing slash. It's part of the normalization logic so it should probably go in `normalizeBasePath()`.

nsqadmin/static/js/app_state.js
ploxiln: I swear I tested `path.Join()` to see how it would join a slash, but I must have remembered wrong, and I was wrong here, the result of `bp("/")` would indeed be just `"/base"`.

I just thought of this whole thing in an apparently opposite way from how you. Your original way was probably quite simple, and the way I originally thought of it was also quite simple. But different enough that I didn't understand the purpose of some of your logic.

Anyway, whatever works, works.

ploxiln: Why would we choose a path for a new page that ends in `/`, only for our code to remove that `/` here?

There still are special cases for `/` in both variables - but I think a formulation like this is more clear / less tricky:

```js
var bp = this.get('BASE_PATH');
if (bp == '/')
  return p;
if (p == '/')
  return bp;
return bp + p;
```

ploxiln: fwiw, I think Backbone can handle trailing slash:

> Trailing slashes are treated as part of the URL, and (correctly) treated as a unique route when accessed. docs and docs/ will fire different callbacks.

http://backbonejs.org/#Router-routes

I'm a bit more surprised that it can handle the empty string when the leading slash is removed from `this.route(bp('/'), 'topics');` but I think that would happen either way. (It's fun to think through this case for the default base-path: that `bp()` calls this `basePath()`, `BASE_PATH` is "/" so prefix with empty string, remove "/", that results in empty string so return "/" instead, finally `bp()` removes first slash, resulting in empty string.)

mreiferson: Yes, I know that Backbone can support it, I didn't mean to imply otherwise. I just think it's inconsistent with all the other routes that are _not_ currently configured to allow it (although, again, we never link to them without a trailing slash).

This feels like a good baseline so I think we should land this. If you're so inclined, I look forward to your follow up PR where I encourage you to explore this and any other rabbit hole that piques your interest 😉. I am totally willing to land something that demonstrates a simpler overall approach.

ploxiln: yes, let's merge this

nanom1t:
How to set base path in nsqadmin config?

ploxiln:
You can use the command-line flag `--base-path=/whatever` or in the config file it would be `base_path = "/whatever"`

This was merged after version 1.1.0 was tagged and released, so you will have to build nsqadmin from the master branch in order to get this feature. see https://nsq.io/deployment/installing.html#compiling

Issue #769, opened 2016-06-16
nsqadmin: configuration update API


kenjones-cisco:
Currently nsqd has an API for updating the configuration attribute `nsqlookupd_tcp_addresses`. The same feature is needed for nsqadmin.

Also each component nsqd, nsqadmin, and nsqlookupd all support a configuration file option `config`, add support for SIGHUP to trigger reloading the configuration file and continue processing with the updated configurations.

mreiferson:
@kenjones-cisco our thinking was it's much safer to restart `nsqadmin` so using some wrapping service that `TERM`s the process on config changes would be reasonable.

What do you think?

kenjones-cisco:
I'm currently using containerpilot (https://www.joyent.com/containerpilot) that handles registering each instance of nsqd, nsqlookupd, and nsqadmin into consul. For nsqd, when the registered service "nsqlookupd" changes the onChange event is triggered and and I use the API to update the configurations. That allows each nsqd to show up within each nsqlookupd.

For nsqadmin, I need a similar option that doesn't result is shutting down admin to restart it. So either an API like nsqd or the ability to send a signal to reload. Otherwise it would cause the entire container to shutdown and then that results in being dependent on the container scheduler to restart or recreate the container for nsqadmin.

Because of the investment in service discovery, being able to auto configure any component as instances go up and down within the environment.

I've used consul-template in other places where there are many configurations that will change, but for something like nsqadmin that seems overkill.

mreiferson:
@kenjones-cisco thanks for the explanation.

> I've used consul-template in other places where there are many configurations that will change, but for something like nsqadmin that seems overkill.

I guess this feels like the perfect use case for `nsqadmin`, since it feels reasonably low risk to restart as a mechanism for receiving configuration updates.

In any event, I'm fine with adding a config API similar to `nsqd`. Are you interested in taking a pass at this?

mreiferson:
see #773 / #775

PR #829, opened 2017-03-05
nsqadmin: restrict config change API

mreiferson:
#769 introduced API endpoints to nsqadmin to update the runtime config value for `nsqlookupd` to give better options for integrating into dynamic environments where restarting the service to pickup new configs is not practical (aka it's running in a container, etc). This also modeled the API after `/config/:opt` endpoints on `nsqd`

Unfortunately `nsqadmin` has a different exposure risk as it's a UI is meant to provide restricted administration changes, unlike `nsqd` which is expected to be on a trusted network. Changing the config for endpoints nsqadmin connects to provides an undesirable vector where you could iteratively change the config and probe internal network endpoints, etc.

Some options to better handle this:

a) nsqadmin could listen on a separate port for the config change API
b) add a configurable option to enable/disable the `/config/:opt` API
c) remove the `PUT /config/:opt` and support `HUP` reloading instead (not quite as flexible)
d) ....

### DIFF
--- a/nsqadmin/options.go
+++ b/nsqadmin/options.go
@@ -28,6 +28,8 @@ type Options struct {
 	HTTPClientTLSCert               string `flag:"http-client-tls-cert"`
 	HTTPClientTLSKey                string `flag:"http-client-tls-key"`
 
+	AllowConfigFromCIDR string `flag:"allow-config-from-cidr"`
+
 	NotificationHTTPEndpoint string `flag:"notification-http-endpoint"`
 
 	Logger Logger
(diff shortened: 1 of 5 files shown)


mreiferson:
Got sick of waiting 😁, RFR @jehiah

nsqadmin/options.go
```diff
@@ -43,5 +45,6 @@ func NewOptions() *Options {
 		StatsdInterval:           60 * time.Second,
 		HTTPClientConnectTimeout: 2 * time.Second,
 		HTTPClientRequestTimeout: 5 * time.Second,
+		AllowConfigFromCIDR:      "127.0.0.1/8",
```
mreiferson: This default is technically a breaking change, but probably the _right thing to do_.

ploxiln: Agreed, this is what I would have made the default. And the /config API, introduced in a4015c2aba012af68037dccc8c834a5e365daef8, was never in a release ... so you could consider this to not be breaking any previous release.

mreiferson: Ahh, good point, forgot about that!

PR #868, opened 2017-03-24
nsqd: clamp requeue timeout to valid range instead of dropping connection	

### LINKED ISSUE #865: nsqd: connection closed on receiving a requeue timeout higher than the MaxReqTimeout
Running on 0.3.8, however the same issue is on 1.0.0.

I believe the correct behavior here should be returning an error to the client, not killing the connection (`protocol.NewClientErr` vs `protocol.NewFatalClientErr`).

https://github.com/nsqio/nsq/blob/fc261f90aa324921f908fb8dbae105cfe5078d08/nsqd/protocol_v2.go#L715-L718

A snippet from logs scoped to one failing client:
```
[nsqd] 2017/03/23 22:56:42.363204 [10.50.102.108:57424] IDENTIFY: {ShortID:e24143c26459 LongID:e24143c26459 ClientID:e24143c26459.1 Hostname: HeartbeatInterval:0 OutputBufferSize:0 OutputBufferTimeout:0 FeatureNegotiation:true TLSv1:false Deflate:false DeflateLevel:0 Snappy:false SampleRate:0 UserAgent:nsq.js/0.15.1 MsgTimeout:900000}
[nsqd] 2017/03/23 22:56:42.591672 ERROR: [10.50.102.108:57424] - E_INVALID REQ timeout 3897854000000 out of range 0-3600000000000
[nsqd] 2017/03/23 22:56:42.591710 PROTOCOL(V2): [10.50.102.108:57424] exiting ioloop
[nsqd] 2017/03/23 22:56:42.591744 ERROR: client(10.50.102.108:57424) - E_INVALID REQ timeout 3897854000000 out of range 0-3600000000000
[nsqd] 2017/03/23 22:56:42.591807 PROTOCOL(V2): [10.50.102.108:57424] exiting messagePump
```

tsholmes:
Fixes #865 

This will return an error to the client when the requested requeue timeout is higher than the `MaxReqTimeout` option, instead of dropping the connection.

### DIFF
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -712,9 +712,18 @@ func (p *protocolV2) REQ(client *clientV2, params [][]byte) ([]byte, error) {
 	}
 	timeoutDuration := time.Duration(timeoutMs) * time.Millisecond
 
-	if timeoutDuration < 0 || timeoutDuration > p.ctx.nsqd.getOpts().MaxReqTimeout {
-		return nil, protocol.NewFatalClientErr(nil, "E_INVALID",
-			fmt.Sprintf("REQ timeout %d out of range 0-%d", timeoutDuration, p.ctx.nsqd.getOpts().MaxReqTimeout))
+	maxReqTimeout := p.ctx.nsqd.getOpts().MaxReqTimeout
+	clampedTimeout := timeoutDuration
+
+	if timeoutDuration < 0 {
+		clampedTimeout = 0
+	} else if timeoutDuration > maxReqTimeout {
+		clampedTimeout = maxReqTimeout
+	}
+	if clampedTimeout != timeoutDuration {
+		p.ctx.nsqd.logf("PROTOCOL(V2): [%s] REQ timeout %d out of range 0-%d. Setting to %d",
+			client, timeoutDuration, maxReqTimeout, clampedTimeout)
+		timeoutDuration = clampedTimeout
 	}
 
 	err = client.Channel.RequeueMessage(client.ID, *id, timeoutDuration)
--- a/nsqd/protocol_v2_test.go
+++ b/nsqd/protocol_v2_test.go
@@ -1403,6 +1403,66 @@ func TestBadFin(t *testing.T) {
 	test.Equal(t, "E_INVALID Invalid Message ID", string(data))
 }
 
+func TestReqTimeoutRange(t *testing.T) {
+	opts := NewOptions()
+	opts.Logger = test.NewTestLogger(t)
+	opts.Verbose = true
+	opts.MaxReqTimeout = 1 * time.Minute
+	tcpAddr, _, nsqd := mustStartNSQD(opts)
+	defer os.RemoveAll(opts.DataPath)
+	defer nsqd.Exit()
+
+	topicName := "test_req" + strconv.Itoa(int(time.Now().Unix()))
+
+	conn, err := mustConnectNSQD(tcpAddr)
+	test.Nil(t, err)
+	defer conn.Close()
+
+	identify(t, conn, nil, frameTypeResponse)
+	sub(t, conn, topicName, "ch")
+
+	topic := nsqd.GetTopic(topicName)
+	channel := topic.GetChannel("ch")
+	msg := NewMessage(topic.GenerateID(), []byte("test body"))
+	topic.PutMessage(msg)
+
+	_, err = nsq.Ready(1).WriteTo(conn)
+	test.Nil(t, err)
+
+	resp, err := nsq.ReadResponse(conn)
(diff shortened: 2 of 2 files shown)


tsholmes:
Alternatively, we could fallback to a requeue timeout of 0 when it is <0, and a requeue timeout of `MaxReqTimeout` when it is too high.

tsholmes:
ready for review

mreiferson:
> Alternatively, we could fallback to a requeue timeout of 0 when it is <0, and a requeue timeout of MaxReqTimeout when it is too high.

I think we should do this. It seems more desirable than _not_ requeueing the message at all, right?

mreiferson:
e.g. I sincerely doubt client libraries are written in a way that would handle non-fatal REQ failure.

tsholmes:
Yeah you're probably right. I'll update it to do that

tsholmes:
ready for review

nsqd/protocol_v2.go
```diff
@@ -712,9 +712,10 @@ func (p *protocolV2) REQ(client *clientV2, params [][]byte) ([]byte, error) {
 	}
 	timeoutDuration := time.Duration(timeoutMs) * time.Millisecond
 
-	if timeoutDuration < 0 || timeoutDuration > p.ctx.nsqd.getOpts().MaxReqTimeout {
-		return nil, protocol.NewFatalClientErr(nil, "E_INVALID",
```
mreiferson: I guess I'm curious now if it should still return some indication that this happened rather than _silently_ proceeding?

tsholmes: We have no way of telling the client about it, since we don't want to return an error. We could log it on the server?

ploxiln: That could be noisy, but still seems like the best option.

tsholmes: Yeah it's as noisy as the previous behavior, where we would log the fatal error and the connection drop every time someone did this ¯\\_(ツ)_/¯

judwhite: @mreiferson I think hanging up at `IDENTIFY` and `REQ` is worth thinking about some more. We actively monitor client errors, whereas early de-queues are not something we'd figure out right away. I would be happy with "`nsqd` requeued your message, but not exactly as you requested" but there's no mechanism for that. Our use case is we have some last ditch effort `REQ`'s in the `12h` range, if `nsqd` was configured incorrectly we'd see these come around again in an hour. Something to consider.

mreiferson: @judwhite `nsqd` does hang up on a bad `IDENTIFY`. On `REQ`, I agree this new behavior is more correct because the previous behavior would result in the intended message timing out rather than being requeued. In most cases that's probably less desirable if the actual time-to-reprocess is important.

tsholmes: I think the biggest issue here is that if the client has a max-in-flight > 1, any other message that was in flight would also be timed out, which can cause problems when you want exactly (or as close as possible to) 1 successful processing of each message.

judwhite: @mreiferson @tsholmes `MaxReqTimeout` is the maximum `REQ` delay the client can specify, it's not related to `-msg-timeout`. Is that right?

ploxiln: correct, it's a separate option:

```
$ nsqd --help
...
  -max-msg-timeout duration
    	maximum duration before a message will timeout (default 15m0s)
...
  -max-req-timeout duration
    	maximum requeuing timeout for a message (default 1h0m0s)
...
  -msg-timeout string
    	duration to wait before auto-requeing a message (default "1m0s")
```

judwhite: @mreiferson I think I see what you're saying. When `REQ` failed previously `nsqd` continued the message timeout and it gets requeued anyway with 0 delay (is that right?). I suppose 'it depends' if you'd rather see errors on the client or have the server override your request parameters without notification.

spruce:
Is there a way to get the max-req-timeout from a client? I don't see a possibility. Is that correct? That would mean a client can't notify the user that he can't requeue / delay with such a high timeout?

ploxiln:
I think that's correct.

Consider that currently, if nsqd restarts, the delays are forgotten, and messages are queued for delivery ASAP. So if you really need to process a message after a certain later time, you probably want an efficient or cached way to determine that in the consumer, and requeue for later delivery repeatedly until that time is reached (and keep in mind max attempts). Messy, but NSQD was not originally designed as a scheduler, so the requeue delay was originally intended for backoff (and not precise timing).

judwhite:
> Is there a way to get the max-req-timeout from a client? 

@spruce http://127.0.0.1:4151/debug/pprof/cmdline if you need a dirty hack

PR #869, opened 2017-03-28
change the id generate method

halleystar:


mreiferson:
We no longer use a global ID channel as of #741

PR #873, opened 2017-03-30
nsqd: remove back-compat for flag duration ints

ploxiln:
I noticed that --statsd-interval and --msg-timeout are the only two duration flags
that have special functionality to be backwards-compatible with
plain int arguments (meaning seconds or milliseconds).

Is it too late to remove more backwards-compatibility stuff before nsq-1.0?


mreiferson:
LGTM

PR #874, opened 2017-03-31
nsqd: expose memory stats under /stats

### LINKED ISSUE #846: nsqd: expose memory stats under /stats
Currently the `/stats` endpoint exposes most stats that the `statsd` integration provides. However, it looks like memory stats are not exposed.

Enabling `--statsd-mem-stats` has no effect on the number of stats exposed.

Would you be open to expose these?

sparklxb:


mreiferson:
@sparklxb this is great! thank you for the contribution 😍 

I have one request — can we consolidate this new `getMemStats` code with the similar code being used https://github.com/sparklxb/nsq/blob/5a888824fd0b762594575b3525c7abd1df9bb2ed/nsqd/statsd.go#L118-L138?

Thanks!

mreiferson:
ping #846

sparklxb:
Done

mreiferson:
fantastic, mind squashing down to one commit and removing the `resolve #846: ` prefix?

ploxiln:
Just as an aside - the github "Squash and merge" mode works, I've used it once or twice. It lets you modify the commit title and message, then it puts the new commit on top of the target branch (e.g. master) without any merge commit.

Also interestingly, it marks the pull request as closed and merged, but without changing the pull-request branch, so the new commit is only in the target branch and the pull request still shows the old commits.

mreiferson:
Yea, I know, I wish there were an option to do all of that _with_ the merge commit, which I've gotten used to and found valuable when browsing history/lineage.

jehiah:
ditto. for better or worse i like the merge commit.

sparklxb:
It works, following https://help.github.com/articles/changing-a-commit-message/#amending-older-or-multiple-commit-messages

PR #878, opened 2017-04-05
fix: channel race condition of ephemeral channel destroy & recreate

slayercat:
@see https://github.com/nsqio/go-nsq/issues/206


slayercat:
seems won't work. close for now.

Issue #879, opened 2017-04-06
nsqd/nsqdmin: introspection of producer connections


jehiah:
`nsqadmin` does a great job at exposing the consumers on a given topic/channel but there is no way to introspect the producers connected via TCP to a nsqd node.

nsqd should expose a `/producer_stats` endpoint that lists the connections it has that are not subscribe to a topic. It should list the same client information available in `/stats` (tls, remote address, connection time, user_agent, auth, etc). It should also list a lifetime connection stat for number of messages published.

nsqadmin should have a new page (or section on the individual `/node/:node` page) that exposes this information.

sparklxb:
Help wanted?

jehiah:
@sparklxb if you are interested, that would be super awesome!

mreiferson:
👍 but why wouldn't this data just be added to `/stats` on `nsqd`?

jehiah:
Well i'm slightly worried about performance implications of that kitchen sink endpoint for larger topic/channel/client deployments; I'd rather more narrow endpoints, and get to the point of forcing more narrow topic/channel stats queries.

mreiferson:
Meh, you can already filter by topic/channel although it's unclear how that filtering would interact with producer connections.

PR #881, opened 2017-04-10
nsqd: add producer client connections to /stats

sparklxb:
add /producer_stats endpoint for nsqd. #879

### DIFF
--- a/nsqd/client_v2.go
+++ b/nsqd/client_v2.go
@@ -55,6 +55,8 @@ type clientV2 struct {
 	FinishCount   uint64
 	RequeueCount  uint64
 
+	pubCounts map[string]uint64
+
 	writeLock sync.RWMutex
 	metaLock  sync.RWMutex
 
@@ -141,6 +143,8 @@ func newClientV2(id int64, conn net.Conn, ctx *context) *clientV2 {
 
 		// heartbeats are client configurable but default to 30s
 		HeartbeatInterval: ctx.nsqd.getOpts().ClientTimeout / 2,
+
+		pubCounts: make(map[string]uint64),
 	}
 	c.lenSlice = c.lenBuf[:]
 	return c
@@ -211,6 +215,13 @@ func (c *clientV2) Stats() ClientStats {
 		identity = c.AuthState.Identity
 		identityURL = c.AuthState.IdentityURL
 	}
+	pubCounts := make([]PubCount, 0, len(c.pubCounts))
+	for topic, count := range c.pubCounts {
+		pubCounts = append(pubCounts, PubCount{
+			Topic: topic,
+			Count: count,
+		})
+	}
 	c.metaLock.RUnlock()
 	stats := ClientStats{
 		Version:         "V2",
@@ -232,6 +243,7 @@ func (c *clientV2) Stats() ClientStats {
 		Authed:          c.HasAuthorizations(),
 		AuthIdentity:    identity,
 		AuthIdentityURL: identityURL,
+		PubCounts:       pubCounts,
 	}
 	if stats.TLS {
 		p := prettyConnectionState{c.tlsConn.ConnectionState()}
@@ -241,6 +253,13 @@ func (c *clientV2) Stats() ClientStats {
 		stats.TLSNegotiatedProtocolIsMutual = p.NegotiatedProtocolIsMutual
 	}
 	return stats
+}
+
+func (c *clientV2) IsProducer() bool {
+	c.metaLock.RLock()
+	retval := len(c.pubCounts) > 0
+	c.metaLock.RUnlock()
+	return retval
 }
 
 // struct to convert from integers to the human readable strings
@@ -341,6 +360,12 @@ func (c *clientV2) Empty() {
 func (c *clientV2) SendingMessage() {
 	atomic.AddInt64(&c.InFlightCount, 1)
 	atomic.AddUint64(&c.MessageCount, 1)
+}
+
+func (c *clientV2) PublishedMessage(topic string, count uint64) {
+	c.metaLock.Lock()
+	c.pubCounts[topic] += count
+	c.metaLock.Unlock()
 }
 
 func (c *clientV2) TimedOutMessage() {
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -486,24 +486,19 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 	startTime := s.ctx.nsqd.GetStartTime()
 	uptime := time.Since(startTime)
 
-	// If we WERE given a topic-name, remove stats for all the other topics:
+	// filter by topic (if specified)
 	if len(topicName) > 0 {
-		// Find the desired-topic-index:
 		for _, topicStats := range stats {
 			if topicStats.TopicName == topicName {
-				// If we WERE given a channel-name, remove stats for all the other channels:
+				// filter by channel (if specified)
 				if len(channelName) > 0 {
-					// Find the desired-channel:
 					for _, channelStats := range topicStats.Channels {
 						if channelStats.ChannelName == channelName {
 							topicStats.Channels = []ChannelStats{channelStats}
-							// We've got the channel we were looking for:
 							break
 						}
 					}
 				}
-
-				// We've got the topic we were looking for:
 				stats = []TopicStats{topicStats}
 				break
 			}
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -481,6 +481,8 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 	channelName, _ := reqParams.Get("channel")
 	jsonFormat := formatString == "json"
 
+	producerStats := s.ctx.nsqd.GetProducerStats()
+
 	stats := s.ctx.nsqd.GetStats(topicName, channelName)
 	health := s.ctx.nsqd.GetHealth()
 	startTime := s.ctx.nsqd.GetStartTime()
@@ -503,23 +505,46 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 				break
 			}
 		}
+
+		filteredProducerStats := make([]ClientStats, 0)
+		for _, clientStat := range producerStats {
+			var found bool
+			var count uint64
+			for _, v := range clientStat.PubCounts {
+				if v.Topic == topicName {
+					count = v.Count
+					found = true
+					break
+				}
+			}
+			if !found {
+				continue
+			}
+			clientStat.PubCounts = []PubCount{PubCount{
+				Topic: topicName,
+				Count: count,
+			}}
+			filteredProducerStats = append(filteredProducerStats, clientStat)
+		}
+		producerStats = filteredProducerStats
 	}
 
 	ms := getMemStats()
 	if !jsonFormat {
-		return s.printStats(stats, ms, health, startTime, uptime), nil
+		return s.printStats(stats, producerStats, ms, health, startTime, uptime), nil
 	}
 
 	return struct {
-		Version   string       `json:"version"`
-		Health    string       `json:"health"`
-		StartTime int64        `json:"start_time"`
-		Topics    []TopicStats `json:"topics"`
-		Memory    memStats     `json:"memory"`
-	}{version.Binary, health, startTime.Unix(), stats, ms}, nil
+		Version   string        `json:"version"`
+		Health    string        `json:"health"`
+		StartTime int64         `json:"start_time"`
+		Topics    []TopicStats  `json:"topics"`
+		Memory    memStats      `json:"memory"`
+		Producers []ClientStats `json:"producers"`
+	}{version.Binary, health, startTime.Unix(), stats, ms, producerStats}, nil
 }
 
-func (s *httpServer) printStats(stats []TopicStats, ms memStats, health string, startTime time.Time, uptime time.Duration) []byte {
+func (s *httpServer) printStats(stats []TopicStats, producerStats []ClientStats, ms memStats, health string, startTime time.Time, uptime time.Duration) []byte {
 	var buf bytes.Buffer
 	w := &buf
 
@@ -599,6 +624,34 @@ func (s *httpServer) printStats(stats []TopicStats, ms memStats, health string, 
 			}
 		}
 	}
+
+	if len(producerStats) == 0 {
+		fmt.Fprintf(w, "\nProducers: None\n")
+	} else {
+		fmt.Fprintf(w, "\nProducers:")
+		for _, client := range producerStats {
+			connectTime := time.Unix(client.ConnectTime, 0)
+			// truncate to the second
+			duration := time.Duration(int64(now.Sub(connectTime).Seconds())) * time.Second
+			var totalPubCount uint64
+			for _, v := range client.PubCounts {
+				totalPubCount += v.Count
+			}
+			fmt.Fprintf(w, "\n   [%s %-21s] msgs: %-8d connected: %s\n",
+				client.Version,
+				client.ClientID,
+				totalPubCount,
+				duration,
+			)
+			for _, v := range client.PubCounts {
+				fmt.Fprintf(w, "      [%-15s] msgs: %-8d\n",
+					v.Topic,
+					v.Count,
+				)
+			}
+		}
+	}
+
 	return buf.Bytes()
 }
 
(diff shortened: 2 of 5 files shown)


ploxiln:
I think the official client libraries either only produce or only consume on a single connection to nsqd, so if that requirement was formalized, it would remove the need to add another counter to the client struct ("messages" could be re-used).

It is completely valid, and not uncommon, to publish to multiple different topics from a single producer tcp client connection to nsqd. But often only a single topic will be published to by a process / connection. It would be a useful enhancement to associate a producer client tcp connection with the topic(s) it publishes to, so one can see producer stats for different topics differentiated. Perhaps the most practical way is to limit this association to one topic, and if more than one, then just consider it "multiple".

sparklxb:
@ploxiln I did consider what you had mentioned. If producer client publishs more than one topic, there is more work to do, such as providing separate client struct for topics published by one producer or changing client struct (but make it more dirty or complicated). So just wait for owners' comment.

mreiferson:
> I think the official client libraries either only produce or only consume on a single connection to nsqd, so if that requirement was formalized, it would remove the need to add another counter to the client struct ("messages" could be re-used).

Yea, but it's not a _real_ restriction so I don't think we can reasonably do that.

> It is completely valid, and not uncommon, to publish to multiple different topics from a single producer tcp client connection to nsqd.

Feels like we can defer tracking stats at that granularity for this first pass.

nsqd/client_v2.go
```diff
@@ -346,6 +349,10 @@ func (c *clientV2) SendingMessage() {
 	atomic.AddUint64(&c.MessageCount, 1)
 }
 
+func (c *clientV2) PublishMessage(msgNum uint64) {
```
mreiferson: `s/Publish/Published`

nsqd/http.go
```diff
@@ -52,6 +52,7 @@ func newHTTPServer(ctx *context, tlsEnabled bool, tlsRequired bool) *httpServer
 	router.Handle("POST", "/pub", http_api.Decorate(s.doPUB, http_api.V1))
 	router.Handle("POST", "/mpub", http_api.Decorate(s.doMPUB, http_api.V1))
 	router.Handle("GET", "/stats", http_api.Decorate(s.doStats, log, http_api.V1))
+	router.Handle("GET", "/producer_stats", http_api.Decorate(s.doProducerStats, log, http_api.V1))
```
mreiferson: I'm still _strongly_ in favor of keeping this all in `/stats`, perhaps with a flag to enable.

nsqd/nsqd.go
```diff
@@ -51,7 +55,8 @@ type NSQD struct {
 	errValue  atomic.Value
 	startTime time.Time
 
-	topicMap map[string]*Topic
+	topicMap  map[string]*Topic
+	producers map[int64]Producer
```
mreiferson: Calling this variable `producers` seems misleading because we can't really know or categorize for sure. I'd rather call this `clients`.

nsqd/protocol_v2.go
```diff
@@ -796,6 +797,9 @@ func (p *protocolV2) PUB(client *clientV2, params [][]byte) ([]byte, error) {
 		return nil, protocol.NewFatalClientErr(err, "E_PUB_FAILED", "PUB failed "+err.Error())
 	}
 
+	p.ctx.nsqd.AddClient(client.ID, client)
```
mreiferson: why do we add the client every time here, can't we do this once when _any_ client connects? This relates to my comments above about variable naming.

nsqd/nsqd.go
```diff
@@ -206,6 +212,30 @@ func (n *NSQD) GetStartTime() time.Time {
 	return n.startTime
 }
 
+// AddClient adds a client to the producers map
+func (n *NSQD) AddClient(clientID int64, client Producer) {
+	n.Lock()
+	defer n.Unlock()
```
ploxiln: if we're really going to have to call `AddClient()` on every single `PUB`, we can make it a bit more efficient by not using `defer`

ploxiln: ... but if there was another client property, from which it could be determined if the client connection had already published anything, `AddClient()` could usually be skipped

ploxiln:
Just to be clear, @mreiferson, I proposed not tracking multiple topics per publisher, just one. Are you agreeing with that, or saying we should never track any topic for publisher stats?

mreiferson:
@ploxiln I'm suggesting that, for now, perhaps it's easiest to land something that doesn't track the destination topic at all (but iterate on that after).

But, as I'm writing this, I'm realizing that really wouldn't be all that helpful for the use case here.

nsqd/http.go
sparklxb: A flag to return one or all?

nsqd/protocol_v2.go
sparklxb: We can't make sure if a client is a producer, until it's first PUB or MPUB.

nsqd/nsqd.go
sparklxb: I think we can check if PublishCount property I added is greater than 0.

nsqd/http.go
ploxiln: I think he means "all", e.g. a flag/param like "with_producers=true" or "with_producers=1" or "producers=1" (brevity to taste ;).

If there was (minimal) channel information associated with the producers, then this new flag could combine with the existing "topic" flag in the logical way.

nsqd/protocol_v2.go
ploxiln: I think what mreiferson is suggesting is to keep a set of all clients (added to the set when they connect). Then, when generating publisher statistics, you can iterate over all clients and include them only if they have published any messages.

nsqd/client_v2.go
```diff
@@ -346,6 +353,10 @@ func (c *clientV2) SendingMessage() {
 	atomic.AddUint64(&c.MessageCount, 1)
 }
 
+func (c *clientV2) MessagesPublished(msgNum uint64) {
```
mreiferson: Let's call this `PublishedMessage`

nsqd/http.go
```diff
@@ -464,8 +464,14 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 	formatString, _ := reqParams.Get("format")
 	topicName, _ := reqParams.Get("topic")
 	channelName, _ := reqParams.Get("channel")
+	withProducers, _ := reqParams.Get("with_producers")
 	jsonFormat := formatString == "json"
 
+	var producersStats []ClientStats
+	if withProducers == "1" {
```
mreiferson: `1` seems like an overly specific value... how about if it's a non-empty string instead?

ploxiln: It's sometimes confusing if "0" and "no" and "false" are considered "true", and awkward if you need to override a parameter with a blank string e.g. in your own helper functions. So I humbly suggest checking against "1", "yes", and "true", perhaps case-insensitive.

mreiferson: that's a fair point, 👍 to `1`, `yes`, and `true` case-insensitive

mreiferson: However, with that, I would argue that the name of the param should be `producers`, not `with_producers`.

sparklxb:
Ready for review and follow-up for nsqadmin. By the way, certificates under the nsqd/test directory has expired, so some test cases fail. please make a update.

nsqd/nsqd.go
```diff
@@ -206,6 +213,26 @@ func (n *NSQD) GetStartTime() time.Time {
...
+	n.Unlock()
+}
+
+// RemoveClient removes a client from the producers map
+func (n *NSQD) RemoveClient(clientID int64) {
+	n.Lock()
+
+	_, ok := n.clients[clientID]
+	if !ok {
+		return
```
ploxiln: Unlock also needed here

sparklxb: Thanks.

nsqd/nsqd.go
```diff
@@ -206,6 +213,27 @@ func (n *NSQD) GetStartTime() time.Time {
 	return n.startTime
 }
 
+// AddClient adds a client to the producers map
+func (n *NSQD) AddClient(clientID int64, client Client) {
+	n.Lock()
+
```
mreiferson: whitespace 👮

nsqd/nsqd.go
```diff
@@ -206,6 +213,27 @@ func (n *NSQD) GetStartTime() time.Time {
...
+	n.clients[clientID] = client
+	n.Unlock()
+}
+
+// RemoveClient removes a client from the producers map
+func (n *NSQD) RemoveClient(clientID int64) {
+	n.Lock()
+
```
mreiferson: whitespace 👮

mreiferson:
This is looking good to me after those minor nitpicks.

ploxiln:
mreiferson updated the certificates. Rebasing on master will fix those test failures.

mreiferson:
I've rebased this, waiting for tests to pass.

mreiferson:
Given that we're marching towards an actual `1.0` release, I want to be really careful of the new interfaces we add prior to that since we won't have as much of a chance to work and play with them in production.

In that vein, after thinking more about this, it's certainly a good start but it seems like what one would really want is to understand _which topics_ a producer is publishing too, right?

mreiferson:
I've resurrected this PR, rebased, and implemented this from my last comment:

> In that vein, after thinking more about this, it's certainly a good start but it seems like what one would really want is to understand which topics a producer is publishing too, right?

WDYT @sparklxb @ploxiln @jehiah ?

mreiferson:
## `/stats` output

```
nsqd v1.1.0-rc1 (built w/go1.10.3)
start_time 2018-08-01T15:52:49-07:00
uptime 24.680560278s

Health: OK

Memory:
   heap_objects                 8393
   heap_idle_bytes              360448
   heap_in_use_bytes            2195456
   heap_released_bytes          0
   gc_pause_usec_100            0
   gc_pause_usec_99             0
   gc_pause_usec_95             0
   next_gc_bytes                4473924
   gc_total_runs                0

Topics:
   [test           ] depth: 4     be-depth: 4     msgs: 0        e2e%:

   [test2          ] depth: 2     be-depth: 1     msgs: 1        e2e%:

   [test3          ] depth: 2     be-depth: 1     msgs: 1        e2e%:

Producers:     
   [V2 PROSNAKES            ] msgs: 1        connected: 10s
      [test2          ] msgs: 1

   [V2 PROSNAKES            ] msgs: 1        connected: 8s
      [test3          ] msgs: 1
```

## `/stats?format=json` output

```
{
  "version": "1.1.0-rc1",
  "health": "OK",
  "start_time": 1533163969,
  "topics": [
    {
      "topic_name": "test",
      "channels": [],
      "depth": 4,
      "backend_depth": 4,
      "message_count": 0,
      "paused": false,
      "e2e_processing_latency": {
        "count": 0,
        "percentiles": null
      }
    },
    {
      "topic_name": "test2",
      "channels": [],
      "depth": 2,
      "backend_depth": 1,
      "message_count": 1,
      "paused": false,
      "e2e_processing_latency": {
        "count": 0,
        "percentiles": null
      }
    },
    {
      "topic_name": "test3",
      "channels": [],
      "depth": 2,
      "backend_depth": 1,
      "message_count": 1,
      "paused": false,
      "e2e_processing_latency": {
        "count": 0,
        "percentiles": null
      }
    }
  ],
  "memory": {
    "heap_objects": 9671,
    "heap_idle_bytes": 311296,
    "heap_in_use_bytes": 2244608,
    "heap_released_bytes": 0,
    "gc_pause_usec_100": 0,
    "gc_pause_usec_99": 0,
    "gc_pause_usec_95": 0,
    "next_gc_bytes": 4473924,
    "gc_total_runs": 0
  },
  "producers": [
    {
      "client_id": "PROSNAKES",
      "hostname": "PROSNAKES.local",
      "version": "V2",
      "remote_address": "127.0.0.1:61367",
      "state": 0,
      "ready_count": 0,
      "in_flight_count": 0,
      "message_count": 0,
      "finish_count": 0,
      "requeue_count": 0,
      "connect_ts": 1533163986,
      "sample_rate": 0,
      "deflate": false,
      "snappy": false,
      "user_agent": "to_nsq/1.1.0-rc1 go-nsq/1.0.6",
      "pub_counts": [
        {
          "topic": "test3",
          "count": 1
        }
      ],
      "tls": false,
      "tls_cipher_suite": "",
      "tls_version": "",
      "tls_negotiated_protocol": "",
      "tls_negotiated_protocol_is_mutual": false
    }
  ]
}
```

nsqd/client_v2.go
```diff
@@ -343,6 +362,17 @@ func (c *clientV2) SendingMessage() {
 	atomic.AddUint64(&c.MessageCount, 1)
 }
 
+func (c *clientV2) PublishedMessage(topic string, count uint64) {
+	c.metaLock.Lock()
+	counter, ok := c.pubCounts[topic]
+	if ok {
```
ploxiln: looking up a missing key in a golang map returns the zero value ... it would work, but would it be bad style, to just:

```go
c.pubCounts[topic] += count
```

mreiferson: SGTM

ploxiln:
overall looks great

mreiferson:
I'll squash this down, but I don't want to land this until we stamp stable `1.1.0-rc1`.

ploxiln:
you mean "stable 1.1.0" right :)
sounds good

ploxiln:
there seems to be a possible data race

```
WARNING: DATA RACE
Read at 0x00c420320330 by goroutine 85:
  github.com/nsqio/nsq/nsqd.(*Channel).flush()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/channel.go:212 +0xb50
  github.com/nsqio/nsq/nsqd.(*Channel).exit()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/channel.go:182 +0x3dc
  github.com/nsqio/nsq/nsqd.(*Channel).Close()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/channel.go:147 +0x3d
  github.com/nsqio/nsq/nsqd.(*Topic).exit()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/topic.go:373 +0x4dd
  github.com/nsqio/nsq/nsqd.(*Topic).Close()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/topic.go:335 +0x3d
  github.com/nsqio/nsq/nsqd.(*NSQD).Exit()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/nsqd.go:497 +0x298
  github.com/nsqio/nsq/nsqd.TestTouch()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/protocol_v2_test.go:645 +0xdb5
  testing.tRunner()
      /home/travis/.gimme/versions/go1.8.7.linux.amd64/src/testing/testing.go:657 +0x107
Previous write at 0x00c420320330 by goroutine 49:
  runtime.mapdelete()
      /home/travis/.gimme/versions/go1.8.7.linux.amd64/src/runtime/hashmap.go:598 +0x0
  github.com/nsqio/nsq/nsqd.(*Channel).popInFlightMessage()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/channel.go:468 +0x202
  github.com/nsqio/nsq/nsqd.(*Channel).FinishMessage()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/channel.go:350 +0x50
  github.com/nsqio/nsq/nsqd.(*protocolV2).FIN()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/protocol_v2.go:687 +0x457
  github.com/nsqio/nsq/nsqd.(*protocolV2).Exec()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/protocol_v2.go:177 +0x13d2
  github.com/nsqio/nsq/nsqd.(*protocolV2).IOLoop()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/protocol_v2.go:84 +0x634
  github.com/nsqio/nsq/nsqd.(*tcpServer).Handle()
      /home/travis/gopath/src/github.com/nsqio/nsq/nsqd/tcp.go:43 +0x86f
```

mreiferson:
It does look like there is indeed a race, but it's in an entirely unrelated code path. It's interesting that we've never observed it before, and that does indicate that this change might have made it more likely, but then again it did happen on an `i386` and `go 1.8.3` build.

IMO we can fix it separately if it continues to rear it's ugly head.

ploxiln:
agreed, we can fix separately if we ever see it again

Issue #798, opened 2016-10-14
nsqd: a way to configure logging verbosity


jchauncey:
I have a lot of messages like this - `[nsqd] 2016/10/14 06:12:02.995304 200 GET /stats?format=json (10.2.78.72:34186) 32.394µs` and they are not particularly useful to me.

mreiferson:
Unfortunately no, there are no logging verbosity options.

freeseacher:
+100500 for that. 
My logs full of mostly debug log output. 
i am experimenting with such solution
```
# cat /etc/systemd/system/nsqd.service
[Unit]
Description=Realtime distributed messaging (nsqd)
After=network.target

[Service]
ExecStart=/bin/sh -c '/bin/nsqd -config /etc/nsq/nsqd.conf 2>&1| egrep -v "sending heartbeat|200 GET"'
Restart=always
User=nsq
Group=nsq

[Install]
WantedBy=multi-user.target
```
It works exactly like looks.

mreiferson:
will be fixed in #892, closing

PR #892, opened 2017-05-13
*: replace verbose with log-level

antihax:
Removed verbose option and added a log-level option so users can customize what they receive in logs.
Should resolve issue #798

### DIFF
--- a/apps/nsqadmin/main.go
+++ b/apps/nsqadmin/main.go
@@ -21,7 +21,10 @@ var (
 
 	config      = flagSet.String("config", "", "path to config file")
 	showVersion = flagSet.Bool("version", false, "print version string")
-	logPrefix   = flagSet.String("log-prefix", "[nsqadmin] ", "log message prefix")
+
+	logLevel  = flagSet.String("log-level", "info", "set log verbosity: debug, info, warn, error, or fatal)")
+	logPrefix = flagSet.String("log-prefix", "[nsqadmin] ", "log message prefix")
+	verbose   = flagSet.Bool("verbose", false, "deprecated in favor of log-level")
 
 	httpAddress = flagSet.String("http-address", "0.0.0.0:4171", "<addr>:<port> to listen on for HTTP clients")
 
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -78,9 +78,12 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 
 	// basic options
 	flagSet.Bool("version", false, "print version string")
-	flagSet.Bool("verbose", false, "enable verbose logging")
 	flagSet.String("config", "", "path to config file")
+
+	flagSet.String("log-level", "info", "set log verbosity: debug, info, warn, error, or fatal)")
 	flagSet.String("log-prefix", "[nsqd] ", "log message prefix")
+	flagSet.Bool("verbose", false, "deprecated in favor of log-level")
+
 	flagSet.Int64("node-id", opts.ID, "unique part for message IDs, (int) in range [0,1024) (default is hash of hostname)")
 	flagSet.Bool("worker-id", false, "do NOT use this, use --node-id")
 
--- a/apps/nsqlookupd/nsqlookupd.go
+++ b/apps/nsqlookupd/nsqlookupd.go
@@ -20,8 +20,10 @@ func nsqlookupdFlagSet(opts *nsqlookupd.Options) *flag.FlagSet {
 
 	flagSet.String("config", "", "path to config file")
 	flagSet.Bool("version", false, "print version string")
-	flagSet.Bool("verbose", false, "enable verbose logging")
+
+	flagSet.String("log-level", "info", "set log verbosity: debug, info, warn, error, or fatal)")
 	flagSet.String("log-prefix", "[nsqlookupd] ", "log message prefix")
+	flagSet.Bool("verbose", false, "deprecated in favor of log-level")
 
 	flagSet.String("tcp-address", opts.TCPAddress, "<addr>:<port> to listen on for TCP clients")
 	flagSet.String("http-address", opts.HTTPAddress, "<addr>:<port> to listen on for HTTP clients")
--- a/contrib/nsqd.cfg.example
+++ b/contrib/nsqd.cfg.example
@@ -1,5 +1,5 @@
-## enable verbose logging
-verbose = false
+## log verbosity level: debug, info, warn, error, or fatal
+log-level = "info"
 
 ## unique identifier (int) for this worker (will default to a hash of hostname)
 # id = 5150
--- a/nsqadmin/http.go
+++ b/nsqadmin/http.go
@@ -196,10 +196,10 @@ func (s *httpServer) topicsHandler(w http.ResponseWriter, req *http.Request, ps 
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get topics - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get topics - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 
@@ -242,20 +242,20 @@ func (s *httpServer) topicHandler(w http.ResponseWriter, req *http.Request, ps h
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get topic producers - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get topic producers - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 	topicStats, _, err := s.ci.GetNSQDStats(producers, topicName)
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get topic metadata - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get topic metadata - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 
@@ -282,20 +282,20 @@ func (s *httpServer) channelHandler(w http.ResponseWriter, req *http.Request, ps
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get topic producers - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get topic producers - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 	_, allChannelStats, err := s.ci.GetNSQDStats(producers, topicName)
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get channel metadata - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get channel metadata - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 
@@ -312,10 +312,10 @@ func (s *httpServer) nodesHandler(w http.ResponseWriter, req *http.Request, ps h
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get nodes - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get nodes - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 
@@ -334,10 +334,10 @@ func (s *httpServer) nodeHandler(w http.ResponseWriter, req *http.Request, ps ht
 	if err != nil {
 		pe, ok := err.(clusterinfo.PartialErr)
 		if !ok {
-			s.ctx.nsqadmin.logf("ERROR: failed to get producers - %s", err)
+			s.ctx.nsqadmin.logf(LOG_ERROR, "failed to get producers - %s", err)
 			return nil, http_api.Err{502, fmt.Sprintf("UPSTREAM_ERROR: %s", err)}
 		}
-		s.ctx.nsqadmin.logf("WARNING: %s", err)
+		s.ctx.nsqadmin.logf(LOG_WARN, "%s", err)
 		messages = append(messages, pe.Error())
 	}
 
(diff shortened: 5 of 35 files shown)


ploxiln:
I think this incremental approach to enhancing nsq logging is not a bad one. Some comments though:

  * We might want to keep backwards-compatibility with the `--verbose` flag (just have it set log-level to debug). Or maybe not, we're still just before 1.0
  * There's some redundancy in `logf(LOG_ERROR, "ERROR: failed ...` that could be avoided by having the logger add the level text prefix
  * Is there some way to avoid having a separate copy of `logging.go` for each app? (Also, each app has its own `logger.go` with just a minimal `Logger` interface, I wonder if that can be consolidated.)

@judwhite probably has opinions on how nsq logging evolves ...

antihax:
- Agree with the --verbose compatibility, should be easy enough to handle.
- The redundancy could be easily removed. I did not want to go through altering messages too much without a discussion.
- `logger.go` and `logging.go` could be consolidated, but the `logf` functions are subtly different between nsqd and nsqadmin. It looks possible to make them similar, but opts would still be different between the three packages.

nsqd/http.go
```diff
@@ -612,6 +612,15 @@ func (s *httpServer) doConfig(w http.ResponseWriter, req *http.Request, ps httpr
...
 				return nil, http_api.Err{400, "INVALID_VALUE"}
 			}
+		case "verbose":
+			var verbose bool
+			err := json.Unmarshal(body, &verbose)
+			if err != nil {
+				return nil, http_api.Err{400, "INVALID_VALUE"}
+			}
+			if verbose {
+				opts.LogLevel = LOG_INFO
```
ploxiln: should this be `LOG_DEBUG` for verbose?

ploxiln: ... or should it actually change `opts.Verbose`?

antihax: It sure should! :D

nsqd/nsqd.go
```diff
@@ -94,25 +94,25 @@ func New(opts *Options) *NSQD {
...
+		n.logf(LOG_FATAL, "--node-id must be [0,1024)")
 		os.Exit(1)
 	}
 
 	if opts.StatsdPrefix != "" {
 		var port string
 		_, port, err = net.SplitHostPort(opts.HTTPAddress)
 		if err != nil {
-			n.logf("ERROR: failed to parse HTTP address (%s) - %s", opts.HTTPAddress, err)
+			n.logf(LOG_ERROR, "failed to parse HTTP address (%s) - %s", opts.HTTPAddress, err)
```
ploxiln: Pre-existing bug: looks like this should be `LOG_FATAL`

ploxiln:
This is looking good to me.

I'm still interested in consolidating the common parts between nsqd / nsqlookupd / nsqadmin, but that can be done in a later pull request. I might do that part myself :)
I think what you have here is a good improvement.

apps/nsqd/nsqd.go
```diff
@@ -15,7 +15,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/judwhite/go-svc/svc"
-	"github.com/mreiferson/go-options"
+	options "github.com/mreiferson/go-options"
```
mreiferson: can we drop this change in all places?

apps/nsqadmin/main.go
```diff
@@ -21,7 +21,10 @@ var (
 
 	config      = flagSet.String("config", "", "path to config file")
 	showVersion = flagSet.Bool("version", false, "print version string")
-	logPrefix   = flagSet.String("log-prefix", "[nsqadmin] ", "log message prefix")
+
+	logLevel  = flagSet.Int("log-level", nsqadmin.LOG_INFO, "set log verbosity: 1 through 5 (Debug, Info, Warn, Error, Fatal)")
+	logPrefix = flagSet.String("log-prefix", "[nsqadmin] ", "log message prefix")
+	verbose   = flagSet.Bool("verbose", false, "depreceated in favor of log-level")
```
mreiferson: minor typo: should be `deprecated`

apps/nsqlookupd/nsqlookupd.go
```diff
@@ -20,8 +20,10 @@ func nsqlookupdFlagSet(opts *nsqlookupd.Options) *flag.FlagSet {
 
 	flagSet.String("config", "", "path to config file")
 	flagSet.Bool("version", false, "print version string")
-	flagSet.Bool("verbose", false, "enable verbose logging")
+
+	flagSet.Int("log-level", nsqlookupd.LOG_WARN, "set log verbosity: 1 through 5 (Debug, Info, Warn, Error, Fatal)")
```
mreiferson: Is it a better UX for these to be strings rather than ints?

apps/nsqd/nsqd.go
antihax: Sure thing.

apps/nsqlookupd/nsqlookupd.go
antihax: It would be. Could process the string into an int once the flag parsing is done, deal with errors, etc.

Is it worth the extra effort?

antihax:
Odd error on the go 1.7 386 node, twice in a row:
```
Bad response status from coveralls: 422
{"message":"Couldn't find a repository matching this job.","error":true}
The command "./coverage.sh --coveralls" exited with 1.
````

File: apps/nsqlookupd/nsqlookupd.go (hunk 5 above)
Comment (ploxiln) [inline apps/nsqlookupd/nsqlookupd.go]: Probably yes. Command-line arguments are a sort of API in which changes should be avoided, so better to polish it now.

**Comment by ploxiln** [comment]
Don't worry about random coveralls failures, we won't complain about them :)

File: apps/nsqlookupd/nsqlookupd.go (hunk 5 above)
Comment (antihax) [inline apps/nsqlookupd/nsqlookupd.go]: Let me take a crack at it then.

**Comment by antihax** [comment]
There may be a more elegant way of handling this, but it is escaping me at this time.

**Comment by ploxiln** [comment]
I think `LogLevel` does have to have its default set in `NewOptions()` for each app, even though that's redundant with the flags ... unfortunate but true ...

(this comment was originally added to a commit accidentally, then deleted and copied here)

**Comment by antihax** [comment]
I will throw a default in there tomorrow night.

File: contrib/nsqd.cfg.example [hunk 6]
```diff
@@ -1,5 +1,10 @@
-## enable verbose logging
-verbose = false
+## set log verbosity level
+# 1 = Debug
+# 2 = Info
+# 3 = Warning
+# 4 = Error
+# 5 = Fatal
+log-level = 1
```
ploxiln: config file value should be strings too

antihax:
I wrote a test to check the http configuration side of it and discovered it was racing on logLevel defined in the NSQD struct; I moved it to `opts` but not left unexported.

Similar issue appears with nsqlookupd as `opts` is not `atomic.Value` there.

Check it out and give me some feedback on how you folks feel about this.

ploxiln:
I think only nsqadmin and nsqd have an http interface for configuration changes after startup, so nsqlookupd shouldn't need `atomic.Value`. I also think it's fine if the integer logLevel is "eventually consistent", such that other threads see the new integer logLevel a few seconds after it has changed. I don't know if there's a way to make the race detector just ignore this detail...

nsqadmin/logger.go
```diff
@@ -1,5 +1,58 @@
...
+
+const (
+	LOG_DEBUG = 1
+	LOG_INFO  = 2
+	LOG_WARN  = 3
+	LOG_ERROR = 4
+	LOG_FATAL = 5
+)
+
+func (n *NSQAdmin) logLevelFromString(level string) int {
```
ploxiln: This doesn't need to be a method of the app, it doesn't use the app struct at all.

nsqadmin/logger.go
```diff
@@ -1,5 +1,58 @@
...
+	case "info":
+		return LOG_INFO
+	case "warn":
+		return LOG_WARN
+	case "error":
+		return LOG_ERROR
+	case "fatal":
+		return LOG_FATAL
+	case "":
+		return LOG_INFO
```
ploxiln: I don't think you need to handle the blank case, this can be considered invalid.

I also don't think you need `ToLower()`, you can just require the lowercase strings. This is not a big deal though.

nsqadmin/logger.go
```diff
@@ -1,5 +1,58 @@
...
+	case LOG_WARN:
+		levelString = "WARNING"
+	case LOG_ERROR:
+		levelString = "ERROR"
+	case LOG_FATAL:
+		levelString = "FATAL"
+	}
+
+	if level >= n.getOpts().logLevel || n.getOpts().Verbose {
+		n.getOpts().Logger.Output(2, fmt.Sprintf(levelString+": "+f, args...))
```
ploxiln: How about you do `opts = n.GetOpts()` just once above, so the `atomic.Value` is fetched and cast just once.

nsqadmin/nsqadmin.go
```diff
@@ -38,35 +38,55 @@ func New(opts *Options) *NSQAdmin {
 	n := &NSQAdmin{
 		notifications: make(chan *AdminAction),
 	}
+
+	// check log-level is valid and translate to int
+	switch strings.ToLower(opts.LogLevel) {
+	case "debug":
+		opts.logLevel = LOG_DEBUG
```
ploxiln: should use `logLevelFromString()` here

nsqadmin/logger.go
mreiferson: I had a similar thought, but additionally was going to suggest that we make levels a type that can be shared across applications. We can provide a function that takes a string and returns an appropriately typed value and also implement the `Stringer` interface on the type to satisfy the purpose here.

I _think_ it should also be possible to cast an int directly to this type if its signature is `type LogLevel int`?

ploxiln: Agreed about sharing across applications, but was thinking it would be OK if @antihax did not want to do that refactoring in this pull request. We could have the cmdline flags and http interfaces finalized in this PR, and I could do internal refactoring in a follow-up (and we could "bikeshed" as necessary).

I normally don't entertain promises to clean up in a later pull-request, but since it's me ¯\\_(ツ)_/¯

Or, instead of directly merging this, I could include and then build on it in a separate PR.

mreiferson: OK with deferring if you're going to finish this up @ploxiln

nsqadmin/logger.go
antihax: I can change these easily, concern would be if log-level was missing from the .cfg; a blank case would silently load a default rather than fail.

Likewise if a user had `Debug` in a config instead of `debug`, may take them a moment to notice case matters.

nsqadmin/logger.go
antihax: Fixed in last commit.

nsqadmin/nsqadmin.go
antihax: Fixed.

nsqadmin/logger.go
antihax: I've been mulling over moving it to a separate package.

Concerned that it will end up looking like `nsdq.logf(logger.LOG_WARN, "something")` and a little baffled how to handle the config side. Normally could write a custom marshaling functions to deal with the mapping from string to int but i do not see how to support with go-options.

I am ok with you folks refactoring it later; I am still learning atm.

nsqadmin/logger.go
ploxiln: Right, I was proposing erroring/exiting for anything but the expected lowercase strings. Blank not "silently loading a default" like you have here, but instead failing as invalid.

(I think that if log-level is missing in the config file you'll get the default "info" from `NewOptions()`.)

nsqadmin/logger.go
ploxiln: My first idea was to move this stuff into `internal/lg/` and rename the constants without the `LOG_` prefix, so you would end up typing `nsqd.logf(lg.WARN, "something")`

ploxiln: (I didn't want to use `internal/log/` and possibly conflict with the stdlib in files that import it, but maybe that's not really a problem in practice)

nsqadmin/logger.go
antihax: Yes, you are correct, i see now where the defaults would be loaded.

nsqlookupd/nsqlookupd.go
```diff
@@ -30,20 +31,36 @@ func New(opts *Options) *NSQLookupd {
...
+		n.logLevel = LOG_DEBUG
+	case "info":
+		n.logLevel = LOG_INFO
+	case "warn":
+		n.logLevel = LOG_WARN
+	case "error":
+		n.logLevel = LOG_ERROR
+	case "fatal":
+		n.logLevel = LOG_FATAL
+	case "":
```
ploxiln: another blank

ploxiln: but actually, should use `logLevelFromString()`

nsqadmin/logger.go
mreiferson: @antihax 👍 SGTM, we can follow up. I appreciate your work on this!

nsqadmin/http.go
```diff
@@ -732,6 +732,21 @@ func (s *httpServer) doConfig(w http.ResponseWriter, req *http.Request, ps httpr
 			if err != nil {
 				return nil, http_api.Err{400, "INVALID_VALUE"}
 			}
+		case "log_level":
+			var logLevelStr string // temporary variable in case the value is invalid
+			err := json.Unmarshal(body, &logLevelStr)
```
ploxiln: I was testing this PR locally and ran into this - I don't think this should be json, it should just be a plain string.

Treating the body as json representation of a single string means you need to include the quotes e.g. `"info"` instead of `info` and that seems a bit silly.

(Same comment applies for nsqd.)

ploxiln:
Interesting complication: `internal/http_api/` and `internal/clusterinfo/` use the `Logger` from the app, but have their own `Log()` and `logf()` functions (and http_api is relatively complicated). This could be something I work on during the refactoring...

nsqadmin/http.go
ploxiln: To justify this opinion a bit more: "nsqlookupd_tcp_addresses" is a list so json makes sense, and "verbose" is a boolean which has non-quoted representation in json so that's convenient. But adding quotes around a single word, particularly for a manual tool like curl where you end up running `curl --data-binary '"info"' ...`, is a bit unexpected.

ploxiln:
This seems to me to be close to ready for merge.
We prefer for the commits to be squashed together before merge, I hope that's not too much trouble :)

nsqd/options.go
```diff
@@ -90,6 +92,8 @@ func NewOptions() *Options {
 	return &Options{
 		ID:        defaultID,
 		LogPrefix: "[nsqd] ",
+		LogLevel:  "info",
+		logLevel:  2,
```
ploxiln: I think putting the `logLevel` int default in `nsqd.NewOptions()` isn't needed, because it's set from the string in `nsqd.New()`. (The same goes for nsqadmin, but you didn't set the logLevel int default in `nsqadmin.NewOptions()`.)

But this doesn't hurt (except a tiny bit of code redundancy), and I think it may be fine to wait to fix it in my planned follow-up PR.

ploxiln:
I think this is ready to merge. @mreiferson any objections?

nsqadmin/logger.go
```diff
@@ -1,5 +1,57 @@
...
+	case "info":
+		return LOG_INFO
+	case "warn":
+		return LOG_WARN
+	case "error":
+		return LOG_ERROR
+	case "fatal":
+		return LOG_FATAL
+	default:
+		return -1
```
mreiferson: (This can be fixed in a follow up).

This feels like a C idiom, not Go. If this function has error conditions, it should return `(int, error)` and this should be a case that returns an error.

mreiferson:
I'm good, nice work @antihax!

Cryptophobia:
Is this actually working. I set log-level to "fatal" in the config and still getting lots of these messages:

```
[nsqd] 2018/05/16 16:48:14.474918 200 GET /stats?format=json (172.17.0.7:34458) 28.505µs
[nsqd] 2018/05/16 16:48:15.438242 200 GET /stats?format=json (172.17.0.7:34458) 81.649µs
[nsqd] 2018/05/16 16:48:16.055924 200 GET /ping (172.17.0.1:49356) 1.394µs
[nsqd] 2018/05/16 16:48:16.502575 200 GET /stats?format=json (172.17.0.7:34458) 81.476µs
[nsqd] 2018/05/16 16:48:17.469571 200 GET /ping (172.17.0.1:49362) 1.533µs
[nsqd] 2018/05/16 16:48:17.628870 200 GET /stats?format=json (172.17.0.7:34458) 98.366µs
[nsqd] 2018/05/16 16:48:18.592711 200 GET /stats?format=json (172.17.0.7:34458) 28.124µs
[nsqd] 2018/05/16 16:48:20.095872 200 GET /stats?format=json (172.17.0.7:34458) 87.2µs
[nsqd] 2018/05/16 16:48:21.096870 200 GET /stats?format=json (172.17.0.7:34458) 23.456µs
[nsqd] 2018/05/16 16:48:22.147546 200 GET /stats?format=json (172.17.0.7:34458) 47.282µs
[nsqd] 2018/05/16 16:48:22.665315 200 GET /stats?format=json (172.17.0.7:34458) 24.41µs
[nsqd] 2018/05/16 16:48:23.936802 200 GET /stats?format=json (172.17.0.7:34458) 35.987µs
[nsqd] 2018/05/16 16:48:25.298011 200 GET /stats?format=json (172.17.0.7:34458) 36.1µs
[nsqd] 2018/05/16 16:48:26.055191 200 GET /ping (172.17.0.1:49418) 834ns
[nsqd] 2018/05/16 16:48:26.287143 200 GET /stats?format=json (172.17.0.7:34458) 34.374µs
[nsqd] 2018/05/16 16:48:26.368759 200 GET /stats?format=json (172.17.0.7:34458) 26.843µs
```

ploxiln:
This was merged _after_ version "1.0.0-compat" was released. We need to cut another release.

Cryptophobia:
Oh nooo! I did not notice that this was merged after the last release! I assumed it was in the last release.

Well my birthday is coming up... and it has been almost a year since this was merged in, can we get a new release soon?

Cryptophobia:
@mreiferson : nsqd is a dependency of ours on the hephy project (deis workflow) fork and we would like to limit log level in order to remove noise in the logs. Any idea of when we can get a new release?

tmedford:
@mreiferson I need this too. Not being able to limit logs levels is creating extra cost to our logging solution.

sacus:
@mreiferson @ploxiln  I need this too. Not being able to limit logs levels is creating extra cost to our logging solution.

tmedford:
Can we get a release?

Cryptophobia:
@mreiferson : We really would like to limit log-level in our Hephy Workflow nsqd component. A new release with this feature would be really nice. Is there any way we can assist and help out with the release?

We can fork and create a release but I would rather someone who knows where exactly to tag even Travis builds seem to be passing!

mreiferson:
Yep, sorry for the delay 🙏.

Cryptophobia:
@mreiferson , don't get side-tracked and distracted now ... You have almost finished making this release.  :dollar: :moneybag:

tmedford:
@mreiferson Bump release please.

ploxiln:
There was just a [v1.1.0-rc1](https://github.com/nsqio/nsq/releases/tag/v1.1.0-rc1) "release candidate" which you can try :)

PR #898, opened 2017-05-20
*: internal logging package refactor

ploxiln:
follow-up to #892 

In-progress. So far just nsqd converted to use the internal package, while we decide on design and naming. Also haven't converted the internal `api_response` and `clusterinfo` packages to use log levels and such yet (planning to do that after the main parts of nsqadmin and nsqlookupd, by passing a function similar to NSQD.logf() instead of a Logger to them).

Critique and better ideas are welcome @mreiferson


internal/lg/lg.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	DEBUG = 1
+	INFO  = 2
+	WARN  = 3
+	ERROR = 4
+	FATAL = 5
+)
+
+func ParseLogLevel(levelstr string, verbose bool) (int, error) {
```
mreiferson: Here's what I had in mind: https://play.golang.org/p/bUb1ETf4zY

What do you think?

ploxiln: Yes, I can use a type, but I don't think it helps much.

```go
logger.Output(3, fmt.Sprintf("%s: "+f, msgLevel, args...))
```

doesn't work, go says

```
../../go/src/github.com/nsqio/nsq/internal/lg/lg.go:66: too many arguments in call to fmt.Sprintf
	have (string, LogLevel, []interface {}...)
	want (string, ...interface {})
```

Thus the uglier thing I pushed up.

nsqadmin/options.go
```diff
@@ -6,9 +6,10 @@ import (
 
 type Options struct {
 	LogLevel  string `flag:"log-level"`
-	logLevel  int
 	LogPrefix string `flag:"log-prefix"`
 	Verbose   bool   `flag:"verbose"` // for backwards compatibility
+	Logger    Logger
+	logLevel  int // private, not really an option
```
mreiferson: should be typed?

nsqlookupd/options.go
```diff
@@ -10,15 +10,15 @@ type Options struct {
 	LogLevel  string `flag:"log-level"`
 	LogPrefix string `flag:"log-prefix"`
 	Verbose   bool   `flag:"verbose"` // for backwards compatibility
+	Logger    Logger
+	logLevel  int // private, not really an option
```
mreiferson: same, should be typed

mreiferson:
@ploxiln this is looking good. It's pretty annoying that we're continuing to support `-v` now, and would love to think about ways we could drop that deprecated flag.

nsqadmin/options.go
ploxiln: Maybe a bit confusing: I did some moving-around to all three, but then used the new internal shared logging only for nsqd, while experimenting.

Since you approve of this style I'll roll it out to the rest (and we'll see some nice line-count savings).

ploxiln:
nsqadmin and nsqlookupd converted to `internal/lg/`, logging tests consolidated.

There's more to do for `internal/http_api/`, but you can see the strategy taken for `internal/clusterinfo/`.

mreiferson:
👍

mreiferson:
@ploxiln how're you feeling about this?

ploxiln:
I think I'm done with changes for this round. Need to squash this down, obviously :)

One remaining piece I can find, which has not yet been converted, is the now-separate disk-queue. I'm not sure of the best way to handle that, but it can be in a separate PR IMHO.

ploxiln:
Tastefully squashed, sanity-tested a bit with a light random mix of nsqd/nsqlookupd/nsqadmin/nsq_tail usage.

Coveralls ratings are annoyingly jittery - probably due to the asynchronous parts of the applications randomly hitting some cases depending on runtime scheduling.

mreiferson:
Thanks, nice work.

PR #903, opened 2017-06-03
Fix segment.com logo URL in README.md

0xfeeddeadbeef:
Instead of non-existent segmentio_logo.png it should be segment_logo.png
See: [segment_logo.png](https://github.com/nsqio/nsqio.github.io/blob/master/static/img/segment_logo.png)


mreiferson:
thanks!

PR #905, opened 2017-06-07
typo fixed

melotusme:
correct flock to lock


mreiferson:
Thanks for the contribution!

In this case, I actually intended this to mirror the function being called. Going to close.

Issue #906, opened 2017-06-09
*:struct field tag `json:'version'` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (vet)


zouyee:
when using  `go vet`,I got much errors.such as
`struct field tag `json:'version'` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (vet)`

gocs:
use double quotes in the word version? because the linter exempts the warning.

```go
type Foo struct {
        Version string `json:"version"`
}
```

ploxiln:
yes, that was done in #907

Logicbloke:
Get rid of that space between the colon and the double quotes.
`Version string `json:"version"``
instead of
Version string `json: "version"`

PR #907, opened 2017-06-09
fix vet error in clusterinfo/data.go

zouyee:
fix [#906](https://github.com/nsqio/nsq/issues/906)


mreiferson:
Thanks!

tati-z:
Where is the fix though?

ssravjee:
The issue in my opinion that Go compiler / interpreter is complaining about is there is  a "space" between colon and double quotes,
 `json: "version"`

Once you get rid of the space all is good, see below,
`json:"versoin"`

Issue #648, opened 2015-09-14
nsq_to_nsq: support multiple topics


ChengHuaUESTC:
When I  run "nsq_to_nsq -nsqd-tcp-address=127.0.0.1:4150 -topic=test1 -destination-nsqd-tcp-address=192.168.22.191:4150", I could in 192.168.22.191 receive message with topic test1.
Now, I want to send and receive multiple topics on only one nsq_to_nsq.
Does nsq_to_nsq support? If it could, how should I do?

mreiferson:
@ChengHuaUESTC `nsq_to_nsq` doesn't currently support multiple topics, however we recently made changes like this to `nsq_to_file` (see https://github.com/nsqio/nsq/pull/334).

If you'd like to make similar changes to `nsq_to_nsq` we would consider them.

Thanks!

dash1291:
@ChengHuaUESTC @mreiferson Is this being worked upon at the moment?

mreiferson:
no, are you interested?

dash1291:
Yeah, would like to give it a shot

elvarb:
Did the shot work?

zieckey:
I like this feature.

mreiferson:
resolved in #912 and #945

PR #912, opened 2017-06-27
nsq_to_nsq: support multiple topics

jlr52:
I added support for nsq_to_nsq multi topic listen capability that seems to address the issue from this ticket https://github.com/nsqio/nsq/issues/648.

I am currently writing unit test and documentation for the new feature in nsq_to_nsq, and will submit those pull request soon.

Let me know if there are any issues.

Description:

    Added the ability to listen to many topic and broadcast to one topic for nsq_to_nsq app

Testing:

    Using the following setup

        ./nsqlookupd --broadcast-address="127.0.0.1"

        ./nsqd --broadcast-address="127.0.0.1" --lookupd-tcp-address=127.0.0.1:4160

        ./nsqd --broadcast-address="127.0.0.1" --lookupd-tcp-address=127.0.0.1:4160 -tcp-address=0.0.0.0:4190 -data-path="bin/data4190"

        ./nsqd --broadcast-address="127.0.0.1"  -tcp-address=0.0.0.0:4182 -data-path="bin/data4182" -http-address=0.0.0.0:4

        ./nsq_to_nsq --topic=test --topic=test1 --destination-nsqd-tcp-address=0.0.0.0:4182  --lookupd-http-address=127.0.0.1:4161  --destination-topic=test

    Verify that

        curl -d 'hello world 2' 'http://127.0.0.1:4191/pub?topic=test'
        curl -d 'hello world 2' 'http://127.0.0.1:4191/pub?topic=test1'
        curl -d 'hello world 2' 'http://127.0.0.1:4151/pub?topic=test'
        curl -d 'hello world 2' 'http://127.0.0.1:4151/pub?topic=test1'

        will result in nsqd: -tcp-address=0.0.0.0:4182 receiving those broadcast.

### DIFF
--- a/apps/nsq_to_nsq/nsq_to_nsq.go
+++ b/apps/nsq_to_nsq/nsq_to_nsq.go
@@ -31,8 +31,6 @@ const (
 
 var (
 	showVersion = flag.Bool("version", false, "print version string")
-
-	topic       = flag.String("topic", "", "nsq topic")
 	channel     = flag.String("channel", "nsq_to_nsq", "nsq channel")
 	destTopic   = flag.String("destination-topic", "", "destination nsq topic")
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
@@ -44,6 +42,7 @@ var (
 	lookupdHTTPAddrs    = app.StringArray{}
 	destNsqdTCPAddrs    = app.StringArray{}
 	whitelistJSONFields = app.StringArray{}
+	topics              = app.StringArray{}
 
 	requireJSONField = flag.String("require-json-field", "", "for JSON messages: only pass messages that contain this field")
 	requireJSONValue = flag.String("require-json-value", "", "for JSON messages: only pass messages in which the required field has this value")
@@ -53,7 +52,7 @@ func init() {
 	flag.Var(&nsqdTCPAddrs, "nsqd-tcp-address", "nsqd TCP address (may be given multiple times)")
 	flag.Var(&destNsqdTCPAddrs, "destination-nsqd-tcp-address", "destination nsqd TCP address (may be given multiple times)")
 	flag.Var(&lookupdHTTPAddrs, "lookupd-http-address", "lookupd HTTP address (may be given multiple times)")
-
+	flag.Var(&topics, "topic", "nsq topic (may be given multiple times)")
 	flag.Var(&whitelistJSONFields, "whitelist-json-field", "for JSON messages: pass this field (may be given multiple times)")
 }
 
@@ -255,32 +254,28 @@ func hasArg(s string) bool {
 	return argExist
 }
 
-func main() {
-	var selectedMode int
+func commandLineValidation() {
 
-	cCfg := nsq.NewConfig()
-	pCfg := nsq.NewConfig()
-
-	flag.Var(&nsq.ConfigFlag{cCfg}, "consumer-opt", "option to passthrough to nsq.Consumer (may be given multiple times, see http://godoc.org/github.com/nsqio/go-nsq#Config)")
-	flag.Var(&nsq.ConfigFlag{pCfg}, "producer-opt", "option to passthrough to nsq.Producer (may be given multiple times, see http://godoc.org/github.com/nsqio/go-nsq#Config)")
-
-	flag.Parse()
-
-	if *showVersion {
-		fmt.Printf("nsq_to_nsq v%s\n", version.Binary)
-		return
-	}
-
-	if *topic == "" || *channel == "" {
+	if len(topics) == 0 || *channel == "" {
 		log.Fatal("--topic and --channel are required")
 	}
 
-	if *destTopic == "" {
-		*destTopic = *topic
+	if len(topics) == 1 {
+		if *destTopic == "" {
+			*destTopic = topics[0]
+		}
 	}
 
-	if !protocol.IsValidTopicName(*topic) {
-		log.Fatal("--topic is invalid")
+	if len(topics) > 1 {
+		if *destTopic == "" {
+			log.Fatal("--destination-topic must be specified when multi topic is enabled")
+		}
+	}
+
+	for _, topic := range topics {
+		if !protocol.IsValidTopicName(topic) {
+	        log.Fatal("--topic is invalid")
+		}
 	}
 
 	if !protocol.IsValidTopicName(*destTopic) {
@@ -301,6 +296,25 @@ func main() {
 	if len(destNsqdTCPAddrs) == 0 {
 		log.Fatal("--destination-nsqd-tcp-address required")
 	}
+}
+
+func main() {
+	var selectedMode int
+
+	cCfg := nsq.NewConfig()
+	pCfg := nsq.NewConfig()
+
+	flag.Var(&nsq.ConfigFlag{cCfg}, "consumer-opt", "option to passthrough to nsq.Consumer (may be given multiple times, see http://godoc.org/github.com/nsqio/go-nsq#Config)")
+	flag.Var(&nsq.ConfigFlag{pCfg}, "producer-opt", "option to passthrough to nsq.Producer (may be given multiple times, see http://godoc.org/github.com/nsqio/go-nsq#Config)")
+
+	flag.Parse()
+
+	if *showVersion {
+		fmt.Printf("nsq_to_nsq v%s\n", version.Binary)
+		return
+	}
+
+	commandLineValidation()
 
 	switch *mode {
 	case "round-robin":
@@ -317,9 +331,14 @@ func main() {
 	cCfg.UserAgent = defaultUA
 	cCfg.MaxInFlight = *maxInFlight
 
-	consumer, err := nsq.NewConsumer(*topic, *channel, cCfg)
-	if err != nil {
-		log.Fatal(err)
+	var consumerList []*nsq.Consumer
+
+	for _, topic := range topics {
+		consumer, err := nsq.NewConsumer(topic, *channel, cCfg)
+		if err != nil {
+			log.Fatal(err)
+		}
+		consumerList = append(consumerList, consumer)
 	}
 
 	pCfg.UserAgent = defaultUA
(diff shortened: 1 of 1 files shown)


apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -255,32 +254,28 @@ func hasArg(s string) bool {
...
+		if *destTopic == "" {
+			*destTopic = topics[0]
+		}
 	}
 
-	if *destTopic == "" {
-		*destTopic = *topic
+	if len(topics) > 1 {
+		if *destTopic == "" {
+			log.Fatal("--destination-topic must be specified when multi topic is enabled")
```
ploxiln: A common desired use-case will likely be to relay each topic to the same/corresponding topic name at the destination.

jlr52: That is reasonable, let me add that. I am probably going to have two options, many topic to 1 topic and many topic to its own corresponding topic like you mentioned.

jlr52: @ploxiln After thinking on the requirement a bit, I feel I need a bit clarification on the desired behavior.

Do we want to relay the topic only to nsq instances that is currently listening to that topic specifically?

For example consider the following

   Producer A (topic1) -> nsq_to_nsq (topic1) -> Consumer A (topic4)
   Producer A (topic1) -> nsq_to_nsq (topic1)-> Consumer B (topic1)

In our example above, Producer A produce topic 1, and nsq_to_nsq listen for topic1, however since Consumer A only listen to topic4, does Consumer A ever get relayed topic1 messages?

Right now nsq_to_nsq forces Consumer A to create a new topic (ie. topic 1) , is that desirable?

ploxiln: `nsq_to_nsq` relays to all nsqd specified with `--destination-nsqd-tcp-address`. The topic(s) involved do not matter for that part.

For receiving the messages, there are really two separate modes - nsqd specified with `--nsqd-tcp-address`, or using `--lookupd-http-address` to find all nsqd which have `--topic`. This multiple-topics feature only makes sense for the former, it doesn't go well with the lookupd mode.

ploxiln: ... ignore my second paragraph there, I was thinking of something else :) really just ignore that.

In summary, yes, creating new topics at the destination nsqd is OK and expected.

ploxiln: And for some useful clarification - I think `--destination-topic` can be optional. If specified, it combines all input topics to one output topic. If not specified, all input topics are passed through to identically named destination topics.

jlr52: Cool, I added the change

i. passing multiple listening topic to nsq_to_nsq will now publish the topic to a destination topic with the same name to every destination nsqd (ie. defined by the hostpool mode)

ii. user can elect to use single destination topic instead.

mreiferson:
Thanks for the PR.

re: functionality, I'm not personally a big fan of the multiple input topics to a single output topic, e.g. we don't do that for `nsq_to_http` nor has it ever been a requested feature. Less is more!

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -349,37 +402,45 @@ func main() {
...
+		}
 	}
 
-	err = consumer.ConnectToNSQLookupds(lookupdHTTPAddrs)
-	if err != nil {
-		log.Fatal(err)
+	aggregateStopChan := make(chan bool)
+
+	for _, consumer := range consumerList {
+		go func(consumer *nsq.Consumer, aggregateStopChan chan<- bool) {
```
mreiferson: I suggest looking into [`sync.WaitGroup`](https://golang.org/pkg/sync/#WaitGroup) to handle this cleanup instead of this loop/goroutine/channel approach.

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -31,10 +31,8 @@ const (
 
 var (
 	showVersion = flag.Bool("version", false, "print version string")
-
-	topic       = flag.String("topic", "", "nsq topic")
 	channel     = flag.String("channel", "nsq_to_nsq", "nsq channel")
-	destTopic   = flag.String("destination-topic", "", "destination nsq topic")
+	singleDestinationTopic   = flag.String("single-destination-topic", "", "activate single destination nsq topic Note: default destination topic is the topic itself")
```
ploxiln: I thought the "combine topics" functionality was a reasonable thing to support even though I probably wouldn't use it. In any case, I would not rename the `--destination-topic` flag, I think the existing name is fine, and it's better to not needlessly break compatibility for people who already have services configured to use it.

jlr52: @ploxiln @mreiferson  I am ok with either implementation, if we decide to keep it I can get rid of the rename of destTopic, I can also take out the single topic destination. 

Just note that the current multi topic support forces all destination tcp to accept all topic provided in the command line, I am not sure if this is ok with the original ask. It seems that user may want fine grain control of what nsqd accept what topic, but maybe not. I am open to discussion :)

I should also ask, what is testing criteria for pull request in this repository? I did some testing to make sure what I described in my commit is accurate, I am planning to do some load testing, is there any requirement I should be aware of?

apps/nsq_to_nsq/nsq_to_nsq.go
jlr52: hmm, the implementation only waits for the first consumer to be done and close the nsq_to_nsq when that happens, do we want to wait for all consumer to be done before closing nsq_to_nsq?

apps/nsq_to_nsq/nsq_to_nsq.go
ploxiln: If the user wants to send different topics to different nsqds, they should run multiple instances of nsq_to_nsq.

(That's also the already-possible way to achieve what this PR seeks to make more convenient - multiple instances.)

jlr52: @ploxiln Ok sounds good, I revert the naming for destination topic back to the original naming.

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -255,36 +255,20 @@ func hasArg(s string) bool {
...
 
-	if *destTopic == "" {
-		*destTopic = *topic
-	}
-
-	if !protocol.IsValidTopicName(*topic) {
-		log.Fatal("--topic is invalid")
+	for _, topic := range topics {
+		if !protocol.IsValidTopicName(topic) {
+	        log.Fatal("--topic is invalid")
```
ploxiln: needs gofmt

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -255,36 +255,20 @@ func hasArg(s string) bool {
...
+		}
 	}
 
-	if !protocol.IsValidTopicName(*destTopic) {
-		log.Fatal("--destination-topic is invalid")
+	if *destTopic != "" && !protocol.IsValidTopicName(*destTopic) {
+		log.Fatal("--single-destination-topic is invalid")
```
ploxiln: error message no longer accurate

apps/nsq_to_nsq/nsq_to_nsq.go
```diff
@@ -301,6 +285,81 @@ func main() {
...
+	flag.Var(&nsq.ConfigFlag{pCfg}, "producer-opt", "option to passthrough to nsq.Producer (may be given multiple times, see http://godoc.org/github.com/nsqio/go-nsq#Config)")
+
+	flag.Parse()
+
+	if *showVersion {
+		fmt.Printf("nsq_to_nsq v%s\n", version.Binary)
+		return
+	}
+
+	commandLineValidation()
```
ploxiln: moving some code into this new function above main() just increases the diff, without reducing code duplication or improving structure

apps/nsq_to_nsq/nsq_to_nsq.go
ploxiln: Right - you want to wait for all the consumers, but your current `aggregateStopChan` strategy only waits for one of them.

The previous code waited for either a message on `termChan`, _or_ for the consumer to stop, in case it happened "on its own", before any message on `termChan`. I think that for reasonably recent versions of go-nsq, the consumer won't stop before being told to stop explicitly, it will keep trying to reconnect to nsqlookupd or nsqd. So I think you can just wait for `termChan` first, then `close()` all consumers, then wait for them all to stop.

I think that if that is all true, and that strategy is followed, then the `sync.WaitGroup` is not really needed. I think the `sync.WaitGroup` would have just replaced the `aggregateStopChan`, enabling a convenient reliable way to wait for all consumers instead of just one. You would still need a goroutine for each consumer, waiting on the `StopChan` and then calling `wg.Done()`. But I think you can skip that goroutine-per-consumer, and just loop through all consumers again after calling `.Stop()` on all of them, making sure each is done.

apps/nsq_to_nsq/nsq_to_nsq.go
jlr52: good catch, I should be more careful.

ploxiln: np, it's what review is for :)

ploxiln:
Your current approach creates a `PublishHandler{}` for each topic. Each `PublishHandler{}` has its own hostPool, timermetrics, and set of responder() goroutines. I think we just want one of these things.

We could call it `PublishManager{}` - it would lack a destinationTopic member, and instead we might have minimal a `TopicHandler{}` for each topic, with just the destinationTopic and a reference to the shared PublishManager. It would call the PublishManager's HandleMessage() with the message and an additional topic argument. (This re-organization would also make the differences between `--destination-topic` and matching-topic mode even smaller.)

jlr52:
@ploxiln  

Really sorry for the late reply, I have some personal issue that I need to sort out that end up taking some time.

I was looking through the code review suggestions, and here are my thoughts

1. Is the end result of 
> The previous code waited for either a message on termChan, or for the consumer to stop, in case it happened "on its own", before any message on termChan. I think that for reasonably recent versions of go-nsq, the consumer won't stop before being told to stop explicitly, it will keep trying to reconnect to nsqlookupd or nsqd. So I think you can just wait for termChan first, then close() all consumers, then wait for them all to stop.

> I think that if that is all true, and that strategy is followed, then the sync.WaitGroup is not really needed. I think the sync.WaitGroup would have just replaced the aggregateStopChan, enabling a convenient reliable way to wait for all consumers instead of just one. You would still need a goroutine for each consumer, waiting on the StopChan and then calling wg.Done(). But I think you can skip that goroutine-per-consumer, and just loop through all consumers again after calling .Stop() on all of them, making sure each is done.


just deleting 


```
for _, consumer := range consumerList {
		go func(consumer *nsq.Consumer, aggregateStopChan chan<- bool) {
			for {
				select {
					case <-consumer.StopChan:
						aggregateStopChan<-true
						return
				}
			}

		}(consumer, aggregateStopChan)
	}
case <-aggregateStopChan:
    return
```
Since we already have 
```
case  <-termChan:
    for _, consumer := range consumerList {
        consumer.Stop()
    }
}
```


2. I don't know if we should implement the `PublishManager` approach, and the reason is each `consumer` might route to a different `topic`, so I need to keep track of the mapping between consumer and its destination topic either in a global variable that `HandleMessage` has access to or just simply save each mapping in each individual handler. Given that the signature of the handler is `func (ph *PublishHandler) HandleMessage(m *nsq.Message) error {` I feel I should only fetch the mapping from `PublishHandler` or from the parameter instead of from a global variable


Let me know what you think

mreiferson:
> 1. Is the end result of ...

Yes, _all_ of that new cleanup code would be replaced by what @ploxiln is suggesting, which is basically this:

```go
<-termChan
for _, consumer := range consumerList {
    consumer.Stop()
    <-consumer.StopChan
}
```

> 2. I don't know if we should implement the `PublishManager` approach ...

I second @ploxiln's recommendation that we only need one `PublishManager`, copy/pasting his original recommendation:

> We could call it PublishManager{} - it would lack a destinationTopic member, and instead we might have minimal a TopicHandler{} for each topic, with just the destinationTopic and a reference to the shared PublishManager. It would call the PublishManager's HandleMessage() with the message and an additional topic argument. (This re-organization would also make the differences between --destination-topic and matching-topic mode even smaller.)

jlr52:
@mreiferson ok let me fix it up

jlr52:
@mreiferson  @ploxiln 

Hi,

I finally got to made the changes suggested. Apologize for the delay.

The changes are following

1. Publish handler should now be shared across multiple topic handlers.
2. Error message should now be updated to reflect variable name
3. nsq_to_nsq no longer fail after first consumer stop, but will listen to termChan and stop all consumer after receiving from termChan.


Let me know if there are any problems.

Cheers.

mreiferson:
@jlr52 thanks for all your work on this ❤️ ! @ploxiln has taken these commits and landed these changes in #945!

PR #916, opened 2017-07-23
nsqd: diskqueue Logger using logf()

ploxiln:
Now the disk-queue integrates with log-levels (it was the last hold-out I think). This is a way that avoids changing the diskqueue.New() API. It does seem sub-optimal because everything will be logged at INFO, so errors will not be logged at ERROR. (Maybe we should log all diskqueue messages at WARN instead?)

### DIFF
--- a/internal/lg/lg.go
+++ b/internal/lg/lg.go
@@ -19,12 +19,12 @@ const (
 type AppLogFunc func(lvl LogLevel, f string, args ...interface{})
 
 type Logger interface {
-	Output(maxdepth int, s string) error
+	Output(calldepth int, s string) error
 }
 
 type NilLogger struct{}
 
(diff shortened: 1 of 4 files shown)


mreiferson:
@ploxiln given the tradeoff w/ logging levels, what's the argument _for_ doing this?

ploxiln:
With this, these remaining log messages can be silenced, by setting log level to warning.

It would be better to add `lg.AppLogFunc` support to the diskqueue  - I was just being lazy since that's a separate repo which I haven't cloned yet. Ill see if it's reasonably easy to do that. Maybe we don't need strict backwards compatibility there.

mreiferson:
Yea, let's do that, whenever you get a chance.

ploxiln:
replaced by #932

PR #925, opened 2017-08-08
nsqadmin/nsqd: optimize /stats

jinhailang:
* optimize: If "topic" was specified trims to only that topic, if "channel" was specified trims to only that channel.
* optimize: For the Topic page we just need the count of clients, we don't need all their info. For the Channel page, we do need all the client info, but not for all Channels, just for the one.

Nsqd and nsqadmin are compatible with older versions.


nsqadmin/static/js/lib/ajax_setup.js
```diff
@@ -7,6 +7,6 @@ $.ajaxPrefilter(function(options) {
         'X-UserAgent': USER_AGENT,
         'Accept': 'application/vnd.nsq; version=1.0'
     });
-    options['timeout'] = 20 * 1000;
+    options['timeout'] = 120 * 1000;
```
ploxiln: I don't think we want to change this timeout. We never want to wait this long for something to load.

(I understand you need to increase it in order for it to work for your setup. I guess we'd hope to make the stats requests just not take so long somehow.)

nsqd/stats.go
```diff
@@ -145,6 +145,37 @@ func (n *NSQD) GetStats() []TopicStats {
...
+	n.RLock()
+	var realTopic *Topic
+	for _, t := range n.topicMap {
+		if topic == t.name {
+			realTopic = t
+			break
+		}
+	}
+	n.RUnlock()
+	realTopic.RLock()
```
ploxiln: I think you have to handle the possibility that the topic does not exist.

nsqd/stats.go
```diff
@@ -145,6 +145,37 @@ func (n *NSQD) GetStats() []TopicStats {
...
+	for _, c := range realTopic.channelMap {
+		realChannels = append(realChannels, c)
+	}
+	realTopic.RUnlock()
+	sort.Sort(ChannelsByName{realChannels})
+	channels := make([]ChannelStats, 0, len(realChannels))
+	for _, c := range realChannels {
+		c.RLock()
+		clients := make([]ClientStats, 0, len(c.clients))
+		for _, client := range c.clients {
```
ploxiln: Another optimization that is possible here: For the Topic page we just need the count of clients, we don't need all their info. For the Channel page, we do need all the client info, but not for all Channels, just for the one.

ploxiln:
There's a lot of duplication between `GetStats()` and `GetTopicStats()` - it would be nice if most of the implementation were shared, e.g. `GetStats()` could just call the new function with a blank topic to mean "all topics".

nsqd/stats.go
ploxiln: Hmm ... we'd need to add a new param, to omit the client connection details from the stats response, to keep compatibility with anything out there that depends on the full stats responses. Luckily, the `/stats` endpoint will ignore params it is not expecting, so new nsqadmin will still work with old nsqd, it'll just get more info than it needs.

(The topic and channel params to `/stats` have existed for quite a while, since nsqd 0.3.6)

jinhailang:
@ploxiln I have done so, query a lot faster, thank you！
I tested it and was compatible.

nsqd/nsqd_test.go
```diff
@@ -387,7 +387,7 @@ func TestReconfigure(t *testing.T) {
 	nsqd.triggerOptsNotification()
 	test.Equal(t, 2, len(nsqd.getOpts().NSQLookupdTCPAddresses))
 
-	time.Sleep(200 * time.Millisecond)
+	time.Sleep(300 * time.Millisecond)
```
jinhailang: Here may be error, I guess sleep time is too short.

jinhailang:
I've completed. If there is time, then again to help me review the code, thank you!
@ploxiln @mreiferson

nsqadmin/http.go
```diff
@@ -359,7 +359,11 @@ func (s *httpServer) nodeHandler(w http.ResponseWriter, req *http.Request, ps ht
 	var totalMessages int64
 	for _, ts := range topicStats {
 		for _, cs := range ts.Channels {
-			totalClients += int64(len(cs.Clients))
+			if len(cs.Clients) > 0 {
+				totalClients += int64(len(cs.Clients))
+			} else {
+				totalClients += int64(cs.ClientCount)
+			}
```
mreiferson: rather than baking this logic into the client each time, we should just always return `client_count` as an int, separately from `clients` (the array).

internal/clusterinfo/data.go
```diff
@@ -530,7 +530,7 @@ func (c *ClusterInfo) GetNSQDTopicProducers(topic string, nsqdHTTPAddrs []string
 //
 // if selectedTopic is empty, this will return stats for *all* topic/channels
 // and the ChannelStats dict will be keyed by topic + ':' + channel
-func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic string) ([]*TopicStats, map[string]*ChannelStats, error) {
+func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic, selectChannel string) ([]*TopicStats, map[string]*ChannelStats, error) {
```
mreiferson: `s/selectChannel/selectedChannel/`

nsqd/http.go
```diff
@@ -494,7 +494,24 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 	channelName, _ := reqParams.Get("channel")
 	jsonFormat := formatString == "json"
 
-	stats := s.ctx.nsqd.GetStats()
+	var (
+		state TopicStats
+		stats []TopicStats
+	)
```
mreiferson: minor: we don't typically declare variables like this, we instead prefer `var` for each one.

nsqd/http.go
```diff
@@ -494,7 +494,24 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
 	channelName, _ := reqParams.Get("channel")
 	jsonFormat := formatString == "json"
 
-	stats := s.ctx.nsqd.GetStats()
+	var (
+		state TopicStats
+		stats []TopicStats
```
mreiferson: the `state` variable is awkwardly named, how about we call these `stats TopicStats` and `allStats []TopicStats`?

nsqd/http.go
```diff
@@ -494,7 +494,24 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
...
+	)
+	if topicName == "" {
+		stats = s.ctx.nsqd.GetStats()
+	} else if channelName == "" {
+		state, err = s.ctx.nsqd.GetTopicStats(topicName)
+	} else {
+		state, err = s.ctx.nsqd.GetChannelStats(topicName, channelName)
+	}
+	if nil != err {
```
mreiferson: minor: we typically prefer the variable in question to be on the _left_ side of the comparison, e.g. `if err != nil {`

nsqd/http.go
```diff
@@ -494,7 +494,24 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
...
+		s.ctx.nsqd.logf(LOG_ERROR, "failed to get stats - %s", err)
+		return nil, http_api.Err{404, "topic/channel is not found"}
+	} else if nil == stats {
```
mreiferson: This doesn't need to be an `else if`, because the block above returns. Also, same minor note applies here, please move the variable to the _left_ side of the comparison.

nsqd/http.go
```diff
@@ -494,7 +494,24 @@ func (s *httpServer) doStats(w http.ResponseWriter, req *http.Request, ps httpro
...
+		stats = []TopicStats{state}
+	}
+
 	health := s.ctx.nsqd.GetHealth()
 	startTime := s.ctx.nsqd.GetStartTime()
 	uptime := time.Since(startTime)
```
mreiferson: After these changes above, Is any of the topic/channel looping code necessary below this line?

nsqd/stats.go
```diff
@@ -145,6 +148,75 @@ func (n *NSQD) GetStats() []TopicStats {
 	return topics
 }
 
+func (n *NSQD) GetTopicStats(topic string) (TopicStats, error) {
+	n.RLock()
+	var realTopic *Topic
+	for _, t := range n.topicMap {
+		if topic == t.name {
```
mreiferson: `if t.name == topic {`

nsqd/stats.go
```diff
@@ -145,6 +148,75 @@ func (n *NSQD) GetStats() []TopicStats {
...
+			realTopic = t
+			break
+		}
+	}
+	n.RUnlock()
+
+	if nil == realTopic {
```
mreiferson: `if realTopic == nil {`

nsqd/stats.go
```diff
@@ -145,6 +148,75 @@ func (n *NSQD) GetStats() []TopicStats {
...
+		return TopicStats{}, errors.New("the topic is not found.")
+	} else {
```
mreiferson: since the above block returns, we can drop the `else` and un-indent this block

nsqd/stats.go
mreiferson: same applies to function below

nsqd/stats.go
mreiferson: same applies to function below

nsqd/stats.go
mreiferson: same applies to function below

andyxning:
@mreiferson @jinhailang @ploxiln I have not found this PR ahead. But, i also proposed a PR #929 to make stats query more efficient and some of the thoughts are the same with @jinhailang

mreiferson:
@jinhailang thanks for this, but going to close it in favor of #929!

jinhailang:
I'm really sorry that I plan to revise this weekend. I have modified it, but can not commit it, if necessary, it may be available as an alternative.
I have been on the production environment to run for some time, the effect is very good, thank you for your help! @ploxiln @mreiferson

ploxiln:
Just by the way, #929 does include the option to exclude all individual client stats from topic stats, which I think is what gave the big performance gain for you. Someone could still possibly do that, separately after #929 is merged.

andyxning:
Agreed with @ploxiln .

@jinhailang The object is to make stats api for efficient. :)

mreiferson:
Ahh, I missed that, good catch. Yes, we should do that too but let's do it as a follow up at this point.

PR #927, opened 2017-08-14
nsqd: fix deflate level handling

### LINKED ISSUE #926: deflate level specified from client lib in IDENTIFY not work
Greetings,
Upon identify with specified deflate level, nsqd does not accept it but using max deflate level in option.

in IDENTIFY(), identifyData.Deflate does not used to decide the deflate level, is it a bug here?
line #396, protocol_v2.go
`deflate := p.ctx.nsqd.getOpts().DeflateEnabled && identifyData.Deflate
	deflateLevel := 0
	if deflate {
		if identifyData.DeflateLevel <= 0 {
			deflateLevel = 6
		}
		deflateLevel = int(math.Min(float64(deflateLevel), float64(p.ctx.nsqd.getOpts().MaxDeflateLevel)))
	}`

besides, there is a typo in attribute of client connection in nsqadmin, when deflate apply.
image

jehiah:
This fixes #926 by properly using the client provided deflate level from identify

### DIFF
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"math/rand"
 	"net"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


ploxiln:
Can you add the deflate level to the log message "upgrading connection to deflate" below?

jehiah:
@ploxiln good idea. PTAL

nsqd/protocol_v2.go
```diff
@@ -399,7 +399,7 @@ func (p *protocolV2) IDENTIFY(client *clientV2, params [][]byte) ([]byte, error)
 		if identifyData.DeflateLevel <= 0 {
 			deflateLevel = 6
```
ploxiln: one more small thing - setting `deflateLevel` here has no effect, I think we want to set `identifyData.DeflateLevel` to the default of 6 here

mreiferson: Instead of that, we should drop the change on line 402 and update line 398 to:

```go
deflateLevel := identifyData.DeflateLevel
```

Otherwise the JSON response below would be broken.

mreiferson:
Want to fix the typo reported in #926 while you're in here??!?!?!

PR #929, opened 2017-08-16
nsqadmin/nsqd: refactor stats querying

andyxning:
As nsqd `stats` api supports `format`, `topic` and `channel` url parameters. When displaying info for a specific topic and channel, nsqadmin should query nsqd `stats` api with the specified `topic` and `channel` instead querying all channels under the specified `topic` and laterly filter out the specified channel stats. 

Query stats on demand can reduce the time for nsqadmin api to return and will reduce the possibility  that nsqadmin report timeout when to query a topic with many channels and each channel with many consumer clients.

/cc @mreiferson @ploxiln

### DIFF
--- a/internal/clusterinfo/data.go
+++ b/internal/clusterinfo/data.go
@@ -450,7 +450,7 @@ func (c *ClusterInfo) GetNSQDTopicProducers(topic string, nsqdHTTPAddrs []string
 		go func(addr string) {
 			defer wg.Done()
 
-			endpoint := fmt.Sprintf("http://%s/stats?format=json", addr)
+			endpoint := fmt.Sprintf("http://%s/stats?format=json&topic=%s", addr, topic)
 			c.logf("CI: querying nsqd %s", endpoint)
 
 			var statsResp statsRespType
@@ -528,9 +528,10 @@ func (c *ClusterInfo) GetNSQDTopicProducers(topic string, nsqdHTTPAddrs []string
 
 // GetNSQDStats returns aggregate topic and channel stats from the given Producers
 //
+// if selectedChannel is empty, this will return stats for topic/channel
 // if selectedTopic is empty, this will return stats for *all* topic/channels
 // and the ChannelStats dict will be keyed by topic + ':' + channel
-func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic string) ([]*TopicStats, map[string]*ChannelStats, error) {
+func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic string, selectedChannel string) ([]*TopicStats, map[string]*ChannelStats, error) {
 	var lock sync.Mutex
 	var wg sync.WaitGroup
 	var topicStatsList TopicStatsList
@@ -548,7 +549,15 @@ func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic string) ([
 			defer wg.Done()
 
 			addr := p.HTTPAddress()
+
 			endpoint := fmt.Sprintf("http://%s/stats?format=json", addr)
+			if selectedTopic != "" {
+				endpoint += "&topic=" + selectedTopic
+				if selectedChannel != "" {
+					endpoint += "&channel=" + selectedChannel
+				}
+			}
+
 			c.logf("CI: querying nsqd %s", endpoint)
 
 			var resp respType
--- a/nsqd/stats.go
+++ b/nsqd/stats.go
@@ -113,20 +113,36 @@ type ChannelsByName struct {
 
 func (c ChannelsByName) Less(i, j int) bool { return c.Channels[i].name < c.Channels[j].name }
 
-func (n *NSQD) GetStats() []TopicStats {
+func (n *NSQD) GetStats(topic string, channel string) []TopicStats {
 	n.RLock()
-	realTopics := make([]*Topic, 0, len(n.topicMap))
-	for _, t := range n.topicMap {
-		realTopics = append(realTopics, t)
+	var realTopics []*Topic
+	if topic == "" {
+		realTopics = make([]*Topic, 0, len(n.topicMap))
+		for _, t := range n.topicMap {
+			realTopics = append(realTopics, t)
+		}
+	} else if val, exists := n.topicMap[topic]; exists {
+		realTopics = []*Topic{val}
+	} else {
+		n.RUnlock()
+		return []TopicStats{}
 	}
 	n.RUnlock()
 	sort.Sort(TopicsByName{realTopics})
 	topics := make([]TopicStats, 0, len(realTopics))
 	for _, t := range realTopics {
 		t.RLock()
-		realChannels := make([]*Channel, 0, len(t.channelMap))
-		for _, c := range t.channelMap {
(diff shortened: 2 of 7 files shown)


internal/clusterinfo/data.go
```diff
@@ -548,7 +549,16 @@ func (c *ClusterInfo) GetNSQDStats(producers Producers, selectedTopic string) ([
 			defer wg.Done()
 
 			addr := p.HTTPAddress()
-			endpoint := fmt.Sprintf("http://%s/stats?format=json", addr)
+
+			var endpoint string
+			if len(selectedChannel) != 0 {
```
mreiferson: This isn't entirely correct, as it doesn't test for `selectedTopic`, let's just adopt the code in #925 wholesale for this block:

```go
if selectedTopic == "" {
	endpoint = fmt.Sprintf("http://%s/stats?format=json", addr)
} else if selectChannel == "" {
	endpoint = fmt.Sprintf("http://%s/stats?topic=%s&format=json", addr, selectedTopic)
} else {
	endpoint = fmt.Sprintf("http://%s/stats?topic=%s&channel=%s&format=json", addr, selectedTopic, selectChannel)
}
```

nsqd/stats.go
```diff
@@ -113,20 +113,32 @@ type ChannelsByName struct {
...
 func (c ChannelsByName) Less(i, j int) bool { return c.Channels[i].name < c.Channels[j].name }
 
-func (n *NSQD) GetStats() []TopicStats {
+func (n *NSQD) GetStats(topic string, channel string) []TopicStats {
 	n.RLock()
-	realTopics := make([]*Topic, 0, len(n.topicMap))
-	for _, t := range n.topicMap {
-		realTopics = append(realTopics, t)
+	var realTopics []*Topic
+	if len(topic) == 0 {
```
mreiferson: Minor, but In Go, this test should just be:

```go
if topic == "" {
```

nsqd/stats.go
```diff
@@ -113,20 +113,32 @@ type ChannelsByName struct {
...
+		realTopics = make([]*Topic, 0, len(n.topicMap))
+		for _, t := range n.topicMap {
+			realTopics = append(realTopics, t)
+		}
+	} else {
+		realTopics = make([]*Topic, 0, 1)
+		realTopics = append(realTopics, n.topicMap[topic])
```
mreiferson: These two lines can be simplified to:

```go
realTopics = []*Topic{n.topicMap[topic]}
```

nsqd/stats.go
```diff
@@ -113,20 +113,32 @@ type ChannelsByName struct {
...
 	n.RUnlock()
 	sort.Sort(TopicsByName{realTopics})
 	topics := make([]TopicStats, 0, len(realTopics))
 	for _, t := range realTopics {
 		t.RLock()
-		realChannels := make([]*Channel, 0, len(t.channelMap))
-		for _, c := range t.channelMap {
-			realChannels = append(realChannels, c)
+		var realChannels []*Channel
+		if len(channel) == 0 {
```
mreiferson: Same idiomatic improvement here:

```go
if channel == "" {
```

nsqd/stats.go
```diff
@@ -113,20 +113,32 @@ type ChannelsByName struct {
...
+			realChannels = make([]*Channel, 0, len(t.channelMap))
+			for _, c := range t.channelMap {
+				realChannels = append(realChannels, c)
+			}
+		} else {
+			realChannels = make([]*Channel, 0, 1)
+			realChannels = append(realChannels, t.channelMap[channel])
```
mreiferson: And same simplification here:

```go
realChannels = []*Channel{t.channelMap[channel]}
```

internal/clusterinfo/data.go
ploxiln: Let me humbly offer my own variant here:

```go
endpoint := fmt.Sprintf("http://%s/stats?format=json", addr)
if selectedTopic != "" {
        endpoint += "&topic=" + selectedTopic
        if selectedChannel != "" {
                endpoint += "&channel=" + selectedChannel
        }
}
```

mreiferson: 👍

nsqd/stats.go
andyxning: Will do.

nsqd/stats.go
andyxning: Will do.

nsqd/stats.go
andyxning: Will do.

nsqd/stats.go
andyxning: Will do.

internal/clusterinfo/data.go
andyxning: Thanks for your review and the more simplified and clearer refactor.

andyxning:
@mreiferson @ploxiln Comments addressed. PTAL.

nsqd/stats_test.go
```diff
@@ -32,7 +32,7 @@ func TestStats(t *testing.T) {
 	identify(t, conn, nil, frameTypeResponse)
 	sub(t, conn, topicName, "ch")
 
-	stats := nsqd.GetStats()
+	stats := nsqd.GetStats(topicName, "ch")
```
ploxiln: it's probably a good idea to expand this test a bit - put a message to one other topic and channel

ploxiln:
looking pretty good to me

nsqd/stats_test.go
andyxning: Sounds reasonable. Done.

andyxning:
@ploxiln Comments addressed. PTAL.

nsqd/stats.go
```diff
@@ -113,20 +113,30 @@ type ChannelsByName struct {
...
-	for _, t := range n.topicMap {
-		realTopics = append(realTopics, t)
+	var realTopics []*Topic
+	if topic == "" {
+		realTopics = make([]*Topic, 0, len(n.topicMap))
+		for _, t := range n.topicMap {
+			realTopics = append(realTopics, t)
+		}
+	} else {
+		realTopics = []*Topic{n.topicMap[topic]}
```
mreiferson: Just realized we're not checking if the topic exists in `topicMap`, can you add that check here?

nsqd/stats.go
```diff
@@ -113,20 +113,30 @@ type ChannelsByName struct {
...
-		for _, c := range t.channelMap {
-			realChannels = append(realChannels, c)
+		var realChannels []*Channel
+		if channel == "" {
+			realChannels = make([]*Channel, 0, len(t.channelMap))
+			for _, c := range t.channelMap {
+				realChannels = append(realChannels, c)
+			}
+		} else {
+			realChannels = []*Channel{t.channelMap[channel]}
```
mreiferson: same here, need to check if channel is in `channelMap` first

nsqd/stats.go
andyxning: @mreiferson Good catch! I also have enhanced the test for both of none exist topic and channel.

PTAL.

mreiferson:
@andyxning fixed up the commits, LGTM, thanks!

PR #931, opened 2017-08-19
Update alpine base image version

andyxning:
As 3.6 has been the latest release version for alpine. We should keep in sync.

/cc @mreiferson @ploxiln


mreiferson:
thanks, LGTM

PR #932, opened 2017-08-26
nsqd: diskqueue now takes AppLogFunc instead of Logger

ploxiln:
... so the diskqueue logging integrates with log levels.

The adapter function is needed because the LogLevel types are not defined in the same package, because `go-diskqueue` cannot import `nsq/internal/lg`. If levels were just plain ints this would be cleaner.

See https://github.com/nsqio/go-diskqueue/pull/2 for companion PR. If that is merged I'll update the Godeps in this PR.

cc @mreiferson


mreiferson:
LGTM after updates on the `go-diskqueue` side

ploxiln:
updated but still waiting to update Godeps

mreiferson:
Just waiting on `Godeps` update then we can land this...

ploxiln:
updated Godeps

PR #935, opened 2017-08-27
nsq_to_file: --topic-pattern causes connection leaks

jxskiss:
ready for review

### DIFF
--- a/apps/nsq_to_file/nsq_to_file.go
+++ b/apps/nsq_to_file/nsq_to_file.go
@@ -85,6 +85,7 @@ type ConsumerFileLogger struct {
 }
 
 type TopicDiscoverer struct {
+	clusterInfo *clusterinfo.ClusterInfo
 	topics   map[string]*ConsumerFileLogger
 	termChan chan os.Signal
 	hupChan  chan os.Signal
@@ -92,8 +93,9 @@ type TopicDiscoverer struct {
 	cfg      *nsq.Config
 }
 
-func newTopicDiscoverer(cfg *nsq.Config) *TopicDiscoverer {
+func newTopicDiscoverer(cfg *nsq.Config, connectTimeout time.Duration, requestTimeout time.Duration) *TopicDiscoverer {
 	return &TopicDiscoverer{
+		clusterInfo: clusterinfo.New(nil, http_api.NewClient(nil, connectTimeout, requestTimeout)),
 		topics:   make(map[string]*ConsumerFileLogger),
 		termChan: make(chan os.Signal),
 		hupChan:  make(chan os.Signal),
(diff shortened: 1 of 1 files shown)


jxskiss:
Improvement:
Establish connections to lookupd servers only when no --topic provided and --topic-pattern provided.

apps/nsq_to_file/nsq_to_file.go
```diff
@@ -514,8 +513,11 @@ func main() {
 			log.Fatal("use --topic to list at least one topic to subscribe to or specify at least one --lookupd-http-address to subscribe to all its topics")
 		}
 		topicsFromNSQLookupd = true
+		// initialize http client only when topicsFromNSQLookupd = true
+		// that is --topic not provided and --topic-pattern provided
+		discoverer.clusterInfo = clusterinfo.New(nil, http_api.NewClient(nil, connectTimeout, requestTimeout))
```
mreiferson: I don't think it is necessary to avoid initializing it, seems harmless and unnecessarily complex

jxskiss: In case of --topic provided, the connection is never used but never released, if one use nsq-to-file to read messages from many individual topics, then there will be many these connections.

mreiferson: You're suggesting that `clusterinfo.New` _makes an HTTP request_ just by initializing it?

jxskiss: Ah, may be I'm understanding http/transport wrong.
I'am not sure that will these codes in "nsq/internal/http_api/api_request.go" make a connection or the connection is only made when sending the exact http endpoint request?
Does the initialization of http client/transport require resource from OS?

```go
func NewDeadlineTransport(connectTimeout time.Duration, requestTimeout time.Duration) *http.Transport {
	transport := &http.Transport{
		Dial: func(netw, addr string) (net.Conn, error) {
			c, err := net.DialTimeout(netw, addr, connectTimeout)
			if err != nil {
				return nil, err
			}
			return &deadlinedConn{connectTimeout, c}, nil
		},
		ResponseHeaderTimeout: requestTimeout,
	}
	return transport
}

type Client struct {
	c *http.Client
}

func NewClient(tlsConfig *tls.Config, connectTimeout time.Duration, requestTimeout time.Duration) *Client {
	transport := NewDeadlineTransport(connectTimeout, requestTimeout)
	transport.TLSClientConfig = tlsConfig
	return &Client{
		c: &http.Client{
			Transport: transport,
			Timeout:   requestTimeout,
		},
	}
}
```

mreiferson: Connections should only be made when requests are made.

jxskiss: Thank your for the clarification, I will push a new commit soon later.

ploxiln: Just to clarify: the Transport and Client are just some structs with configuration (and function pointers etc), they just use a bit of memory (maybe a kilobyte or two). No file descriptors are created (or bound or connected) until a Dial or Request method is called.

jxskiss:
The lazy initializing of clusterInfo has been reset.

mreiferson:
LGTM

PR #937, opened 2017-08-29
nsqd: add client metrics

andyxning:
This PR will add client related metrics to nsqd. As for now, we only have topic and channel related metrics. When we want to know the consumer rate we need to summarize all the consumer rate for all consumer instances.

**Caution**: This uses client remote address as the key to identify each  connection. It will send as many metrics as connections, i.e., it is proportionable with the number of connections. 

/cc @ploxiln @mreiferson

### DIFF
--- a/nsqd/statsd.go
+++ b/nsqd/statsd.go
@@ -5,6 +5,8 @@ import (
 	"math"
 	"time"
 
+	"strings"
+
 	"github.com/nsqio/nsq/internal/statsd"
 )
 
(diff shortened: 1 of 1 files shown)


ploxiln:
`nsq.<nsqd_host>_<nsqd_port>.topic.<topic_name>.channel.<channel_name>.client.<client_host>_<client_port>.finished_count`

This isn't a great idea. I think it would have to be conditional if we included it.

The reason is because the number of keys can be practically un-bounded: with `client_port` in the key, it's a new key every time a client restarts or otherwise reconnects. If you're using the classic graphite/carbon/whisper stack for these stats it can take massive amounts of disk space, relative to the amount of stats actually collected. Maybe if you translate for something like prometheus or influxdb, something with "tags", it's OK. Maybe if it's just `client_host`, it's practical if you have a relatively slow-changing group of servers running nsq consumers (comparable to the group of servers running nsqd).

andyxning:
@ploxiln Agreed.

> Maybe if it's just client_host, it's practical if you have a relatively slow-changing group of servers running nsq consumers (comparable to the group of servers running nsqd).

I have thought about that. But for a connection, remote address is the key to uniquely identify it. Btw, we can aggregate the metrics for a single `client_host` and then use it in the metric actually. So, @mreiferson @ploxiln which one do you suggest?

mreiferson:
I'm -1 on this. I'm also concerned about the sheer volume of keys this could produce. Clients are typically short-lived relative to topics and channels, which is why I feel like they're a poor use case for statsd / looking at trends over time.

andyxning:
@mreiferson @ploxiln 

How about we record client metrics in nsqd instances dimensionality. I.e., 
`nsq.<nsqd_host>_<nsqd_port>.topic.<topic_name>.channel.<channel_name>.client.finished_count`

As the instance for nsqd should not change frequently. It should be well enough to do this. WDYT.

mreiferson:
@andyxning we already have `requeue_count`, and `finish_count` is derivable from the existing metrics

andyxning:
@mreiferson My fault for `requeue_count` metrics. :( 

As for `finish_count`, if the monitor infra is Graphite, it is able to be derived from `message_count`, `depth`, `in_flight_count`, `requeue_count` and `timeout_count` through Graphite API.

Actually, we use OpenTSDB as the monitor infra and we use a tool to convert nsq Graphite based metrics to OpenTSDB's tagKV based metrics. There is no such function like [`diffSeries`](http://graphite-api.readthedocs.io/en/latest/functions.html#graphite_api.functions.diffSeries) in OpenTSDB for us to do the same job. 

Thus, it is good to directly include a metric about `finish_count`.

ploxiln:
Your converter tool/service could derive it :)

Your use of OpenTSDB "tagkv" explains why you didn't think a few keys *per client connection* would be a bad thing. But you have to consider all the common use-cases of the graphite feature, and also the complexity of adding variations and options. There are better and worse places to push complexity. In your case, you already have a converter, it makes sense for any additional adaptations needed for OpenTSDB to be in that piece.

andyxning:
Sounds reasonable @ploxiln 

It is indeed not so good to add specific metrics just for our use cases. Learned some thing about how to maintain an open source. :)

PR #942, opened 2017-09-16
travis: update go versions

dvrkps:


andyxning:
@dvrkps I don't think we need to keep backward compatibility with Golang 1.6 anymore since 1.9 has been released. Keep backward compatibility for latest three major Golang versions should be fine.

@mreiferson @ploxiln WDYT.

andyxning:
BTW, seems the error in travisCI has nothing to do this change. @mreiferson @ploxiln PTAL.

mreiferson:
pushed a few commits here, tests should pass 🤞

mreiferson:
thanks @dvrkps

PR #945, opened 2017-09-25
nsq_to_nsq: multiple topics support

ploxiln:
my proposed changes to #912 

I squashed #912 down to one commit, then added my proposed changes, mostly code-style. See my commit messages. If we agree on these changes, I'll squash most of my changes (all but the last commit) into @jlr52's commit (which will keep his authorship).

The biggest change is to re-inline `commandLineValidation()` and `initConsumerAndHandler()` into `main()` - this is admittedly subjective, but the reasoning is that it reduces the diff against master to about half of what #912 was.

I'll do my own quick manual tests now (but I'm confident :). Take a quick look and let me know what you think @jlr52 @mreiferson

### DIFF
--- a/apps/nsq_to_nsq/nsq_to_nsq.go
+++ b/apps/nsq_to_nsq/nsq_to_nsq.go
@@ -31,10 +31,8 @@ const (
 
 var (
 	showVersion = flag.Bool("version", false, "print version string")
-
-	topic       = flag.String("topic", "", "nsq topic")
 	channel     = flag.String("channel", "nsq_to_nsq", "nsq channel")
-	destTopic   = flag.String("destination-topic", "", "destination nsq topic")
+	destTopic   = flag.String("destination-topic", "", "activate single destination nsq topic Note: default destination topic is the topic itself")
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 
 	statusEvery = flag.Int("status-every", 250, "the # of requests between logging status (per destination), 0 disables")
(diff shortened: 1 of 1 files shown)


ploxiln:
Tested locally with a few nsqd, topics, and made-up messages, both with and without `--destination-topic`. Seems to work :)

mreiferson:
LGTM, thanks

mreiferson:
FYI, GitHub added a feature where maintainers can push to PR branches, see https://github.com/blog/2247-improving-collaboration-with-forks.

I've been using this to cleanup commits for recently merged PRs from contributors and I _think_ it would have worked for you on #912 @ploxiln.

ploxiln:
Ah, I didn't think of that, will try that next time

mreiferson:
@ploxiln interested in applying the same multi-topic changes to `nsq_to_http`? It might benefit from the Go http client changes too.

ploxiln:
Not anytime soon, I just want to finish what I started (I haven't forgotten the "contrib modules" thing :)

`nsq_to_http` uses `http_api.NewDeadlineTransport()` so it'll get the http client transport updates when they're merged. I don't think we make custom Transports anywhere besides that helper function.

jlr52:
@ploxiln Thanks for cleaning up the diff, I definitely forgot to change that back. I will make sure I remember not to increase the diff next time.

@mreiferson  @ploxiln  Can I take a stab on supporting multi topic for nsq_to_http ?

mreiferson:
@jlr52 go for it!

PR #946, opened 2017-09-25
internal/http_api: update NewDeadlineTransport() with new go http features

ploxiln:
More timeouts and limits were added to http.Transport in go-1.7.
Some that were enabled in http.DefaultTransport ended up disabled
in nsq's with-custom-timeouts http.Transport.
IdleConnTimeout in particular is helpful.

see DefaultTransport at https://golang.org/pkg/net/http/#RoundTripper
Not yet field-tested :)

### DIFF
--- a/internal/http_api/api_request.go
+++ b/internal/http_api/api_request.go
@@ -13,30 +13,19 @@ import (
 	"time"
 )
 
-type deadlinedConn struct {
-	Timeout time.Duration
-	net.Conn
-}
-
-func (c *deadlinedConn) Read(b []byte) (n int, err error) {
-	return c.Conn.Read(b)
-}
-
-func (c *deadlinedConn) Write(b []byte) (n int, err error) {
-	return c.Conn.Write(b)
-}
-
 // A custom http.Transport with support for deadline timeouts
 func NewDeadlineTransport(connectTimeout time.Duration, requestTimeout time.Duration) *http.Transport {
+	// arbitrary values copied from http.DefaultTransport
 	transport := &http.Transport{
-		Dial: func(netw, addr string) (net.Conn, error) {
-			c, err := net.DialTimeout(netw, addr, connectTimeout)
-			if err != nil {
-				return nil, err
-			}
-			return &deadlinedConn{connectTimeout, c}, nil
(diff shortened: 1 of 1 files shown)


ploxiln:
When #942 was opened my thought was "why drop support for go-1.6 before having a good specific reason" - my C code works on compilers (and thus linux distros) from around the time Go was first released as beta. Anyway, here's a good reason I guess.

Reminder to self to update https://github.com/nsqio/nsqio.github.io/blob/master/_posts/2014-03-01-installing.md

internal/http_api/api_request.go
```diff
@@ -28,15 +28,17 @@ func (c *deadlinedConn) Write(b []byte) (n int, err error) {
...
 // A custom http.Transport with support for deadline timeouts
 func NewDeadlineTransport(connectTimeout time.Duration, requestTimeout time.Duration) *http.Transport {
+	// arbitrary values copied from http.DefaultTransport
 	transport := &http.Transport{
-		Dial: func(netw, addr string) (net.Conn, error) {
-			c, err := net.DialTimeout(netw, addr, connectTimeout)
-			if err != nil {
-				return nil, err
-			}
-			return &deadlinedConn{connectTimeout, c}, nil
```
mreiferson: can we drop the `deadlinedConn` struct/methods above?

ploxiln: nice catch, silly of me to leave that stuff

ploxiln:
I plan to roll this out to a few servers tomorrow, and see how it does over a week or so.

ploxiln:
(rebased just for testing purposes)

ploxiln:
Everything seems to work. I'd say go ahead and merge this now.

I don't have the patience to complete the fully scientific side-by-side test case where I go a full week without touching two nsqadmin instances, one with this update and one without. I had it set up but one of the nsqadmin instances was used just now (for a legitimate purpose by a co-worker). I can set up a better test with some more effort but it'll be another week to reproduce the original problem. I'll still probably make an attempt, but I think this is probably a good general update, which is at least now sanity-tested to not immediately break anything.

PR #954, opened 2017-10-29
dep; update travis, dist, test, coverage

judwhite:
Similar to https://github.com/nsqio/go-nsq/pull/218

Ready for review. @mreiferson @jehiah @ploxiln


dist.sh
```diff
@@ -18,10 +18,7 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 rm -rf   $DIR/dist/docker
 mkdir -p $DIR/dist/docker
-rm -rf   $DIR/.godeps
-mkdir -p $DIR/.godeps
-export GOPATH=$DIR/.godeps:$GOPATH
-GOPATH=$DIR/.godeps gpm install
```
mreiferson: I've always liked that `dist.sh` preserves the state of my `GOPATH`, should we maintain that property?

judwhite: @mreiferson With `dep` you don't have to change `GOPATH`, it uses `./vendor`. Maybe I misunderstood.

mreiferson: Perfect!

judwhite:
@mreiferson Squashed. `go1.9 GOARCH=386` failed on `TestReconfigure`, flaky test?

Tests also take a bit to execute, let me know if you want to drop `go1.7`.

ploxiln:
I re-started that one sub-test and it passed. (I intended to copy the failure details and make an issue for the flaky test, but wasn't being careful and lost the copy.)

mreiferson:
this is cool, thanks.

@judwhite last request, update the docs on http://nsq.io/deployment/installing.html#building-from-source?

PR #957, opened 2017-10-30
*: multi-topic nsq_tail; automated docker builds; etc

soar:
I've used NSQ for one of our projects - It is cool system, thanks!

But I had very annoying problem - We have more than 5 queues and for each I should start new container of nsq_tail:

```yaml
version: '3.2'
services:
  nsqtail:
    image: nsqio/nsq
    command: /nsq_tail --topic=mytopic1 --lookupd-http-address=nsqlookupd:4161
  nsqtail2:
    image: nsqio/nsq
    command: /nsq_tail --topic=mytopic2 --lookupd-http-address=nsqlookupd:4161
  nsqtail3:
    image: nsqio/nsq
    command: /nsq_tail --topic=mytopic3 --lookupd-http-address=nsqlookupd:4161
...
```

I know that Docker is "lightweight" system... But not lightweight-enough for running too many containers for simple tasks. So my idea was to make `nsq_tail` listen for multiple topics at once and write all messages from them. Now I can do simply:

```yaml
version: '3.2'
services:
  nsqtail:
    image: soarname/nsq
    command: /nsq_tail --topic=topic1 --topic=topic2 --topic=topic3 --lookupd-http-address=nsqlookupd:4161
```

And I see in logs:

```
nsqtail_1     | topic1 | test-message-1
nsqtail_1     | topic2 | test-message-2
nsqtail_1     | topic3 | test-message-3
```

Also I've fixed:
- **`Dockerfile`** - now builds for Docker ecosystem are fully automated - you can check it here: [soarname/nsq](https://hub.docker.com/r/soarname/nsq/)
- **`.travis.yml`** - import path for building forks (without this line - I've got an error "_use of internal package not allowed_")
- **`coverage.sh`** - fix path for `goveralls`
- **`dist.sh`** - fix calling `mktemp` without `XXX` in template

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   - GOARCH=amd64
   - GOARCH=386
 sudo: false
+go_import_path: github.com/nsqio/nsq
 before_install:
   - go get github.com/mattn/goveralls
 script:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,26 @@
+FROM golang:latest AS build
+
+RUN mkdir -p /go/src/github.com/nsqio/nsq
+COPY . /go/src/github.com/nsqio/nsq
+
+WORKDIR /go/src/github.com/nsqio/nsq
+
+RUN wget -O /bin/dep https://github.com/golang/dep/releases/download/v0.3.2/dep-linux-amd64 \
+ && chmod +x /bin/dep \
+ && /bin/dep ensure \
+ && ./test.sh \
+ && CGO_ENABLED=0 make DESTDIR=/opt PREFIX=/nsq GOFLAGS='-ldflags="-s -w"' install
+
 FROM alpine:3.6
 
 EXPOSE 4150 4151 4160 4161 4170 4171
 
-VOLUME /data
-VOLUME /etc/ssl/certs
+# Optional volumes
+# /data - used for persistent storage across reboots
+# VOLUME /data
+# /etc/ssl/certs - directory for SSL certificates
+# VOLUME /etc/ssl/certs
 
-COPY dist/docker/bin/ /usr/local/bin/
+COPY --from=build /opt/nsq/bin/ /usr/local/bin/
 RUN ln -s /usr/local/bin/*nsq* / \
-    && ln -s /usr/local/bin/*nsq* /bin/
+ && ln -s /usr/local/bin/*nsq* /bin/
--- a/apps/nsq_tail/nsq_tail.go
+++ b/apps/nsq_tail/nsq_tail.go
@@ -18,27 +18,42 @@ import (
 var (
 	showVersion = flag.Bool("version", false, "print version string")
 
-	topic         = flag.String("topic", "", "NSQ topic")
 	channel       = flag.String("channel", "", "NSQ channel")
 	maxInFlight   = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 	totalMessages = flag.Int("n", 0, "total messages to show (will wait if starved)")
+	printTopic    = flag.Bool("print-topic", false, "print topic name where message was received")
 
 	nsqdTCPAddrs     = app.StringArray{}
 	lookupdHTTPAddrs = app.StringArray{}
+	topics           = app.StringArray{}
 )
 
 func init() {
 	flag.Var(&nsqdTCPAddrs, "nsqd-tcp-address", "nsqd TCP address (may be given multiple times)")
 	flag.Var(&lookupdHTTPAddrs, "lookupd-http-address", "lookupd HTTP address (may be given multiple times)")
+	flag.Var(&topics, "topic", "NSQ topic (may be given multiple times)")
 }
 
 type TailHandler struct {
+	topicName string
 	totalMessages int
 	messagesShown int
 }
 
 func (th *TailHandler) HandleMessage(m *nsq.Message) error {
 	th.messagesShown++
+
+	if *printTopic {
+		_, err := os.Stdout.WriteString(th.topicName)
+		if err != nil {
+			log.Fatalf("ERROR: failed to write to os.Stdout - %s", err)
+		}
+		_, err = os.Stdout.WriteString(" | ")
+		if err != nil {
+			log.Fatalf("ERROR: failed to write to os.Stdout - %s", err)
+		}
+	}
+
 	_, err := os.Stdout.Write(m.Body)
 	if err != nil {
 		log.Fatalf("ERROR: failed to write to os.Stdout - %s", err)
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -7,6 +7,7 @@ env:
   - GOARCH=amd64
   - GOARCH=386
 sudo: false
+go_import_path: github.com/nsqio/nsq
```
mreiferson: Can you explain why this is needed?

Dockerfile
```diff
@@ -1,10 +1,28 @@
+FROM golang:latest AS build
```
mreiferson: How large is the resulting container image with this approach?

Dockerfile
```diff
@@ -1,10 +1,28 @@
...
+COPY . /go/src/github.com/nsqio/nsq
+
+WORKDIR /go/src/github.com/nsqio/nsq
+
+RUN go get github.com/mattn/goveralls \
+ && wget -O /bin/dep https://github.com/golang/dep/releases/download/v0.3.2/dep-linux-amd64 \
+ && chmod +x /bin/dep \
+ && /bin/dep ensure \
+ && ./test.sh \
+ && ./coverage.sh --coveralls \
```
mreiferson: I don't think we need to run code coverage as part of this build?

apps/nsq_tail/nsq_tail.go
```diff
@@ -18,28 +18,38 @@ import (
...
-	_, err := os.Stdout.Write(m.Body)
+	_, err := os.Stdout.WriteString(th.topicName)
+	if err != nil {
+		log.Fatalf("ERROR: failed to write to os.Stdout - %s", err)
+	}
+	_, err = os.Stdout.WriteString(" | ")
+	if err != nil {
+		log.Fatalf("ERROR: failed to write to os.Stdout - %s", err)
+	}
+	_, err = os.Stdout.Write(m.Body)
```
mreiferson: A side effect of this change is that the output can no longer be (easily) piped into subsequent processes, a common operation when using `nsq_tail`.

I suppose one option is to just multiplex the output without prefixing the topic name?

dist.sh
```diff
@@ -30,7 +30,7 @@ echo "... running tests"
 
 for os in linux darwin freebsd windows; do
     echo "... building v$version for $os/$arch"
-    BUILD=$(mktemp -d -t nsq)
+    BUILD=$(mktemp -d -t nsq-XXXXX)
```
mreiferson: Does this still work on OSX?

ploxiln:
This would be better as two separate pull requests IMHO

Dockerfile
ploxiln: The special part is the separate `FROM alpine:3.6` below - this first image is only used for building, then the result is copied from this image to the separate final image. This technique is newly possible in ~~docker~~ moby 17.05.0-ce

dist.sh
ploxiln: We've hit this before, not sure what happened 😝 
thread starts at https://github.com/nsqio/nsq/pull/718#issuecomment-187758182

Dockerfile
mreiferson: Yep, curious about the results.

dist.sh
mreiferson: Cool, let's update this to reflect @ploxiln's suggestion in https://github.com/nsqio/nsq/pull/718#issuecomment-187877348 ?

mreiferson:
> This would be better as two separate pull requests IMHO

Agreed, unless we can quickly come to a conclusion around `nsq_tail`.

apps/nsq_tail/nsq_tail.go
```diff
@@ -91,29 +100,42 @@ func main() {
...
+			log.Fatal(err)
+		}
+
+		consumers = append(consumers, *consumer)
 	}
 
-	for {
+	for _, consumer := range consumers {
 		select {
 		case <-consumer.StopChan:
```
ploxiln: This is not really useful: it will only wait for the first consumer to stop.

I don't think any consumers will stop before we tell them to though, they will try to reconnect. So this case could just be removed (and the loop as well). This just needs to wait for the sigChan and then tell all consumers to stop (as you have below). After that, to be most correct, it should then wait for all consumers to have stopped.

.travis.yml
soar: @mreiferson without this line - fork build fails, you can see example run here:

https://travis-ci.org/soar/nsq/jobs/294934597

```
package github.com/soar/nsq/apps/nsq_pubsub
    imports github.com/nsqio/nsq/internal/app: use of internal package not allowed
```

And here is Travis documentation for this line:

https://docs.travis-ci.com/user/languages/go/#Go-Import-Path

With this line any fork will work.

Dockerfile
soar: @mreiferson This is multistage-build feature from Docker 17.05+. First image is used for build, second - for distribution.

Dockerfile
soar: @mreiferson Why not? In my projects I always have testing/coveraging as part of build process. For example if one test fails - build will be interrupted and broken image will not be uploaded to Docker repository. I think this is good practice. Yes, it has some impact on build time, but it is not too drastical.

apps/nsq_tail/nsq_tail.go
soar: @ploxiln I've simply rewrited existing logic. Are you sure, that we need to have only waiting for `SigChan`?

Dockerfile
ploxiln: Runnings tests here can be useful because they can fail the build. But in this case the coverage results are just thrown out.

apps/nsq_tail/nsq_tail.go
ploxiln: It wasn't 100% correct before, but in the process of rewriting it you made it more nonsensical. The previous bare for loop is different in nature from a loop over all consumers.

Dockerfile
```diff
@@ -1,10 +1,27 @@
+FROM golang:latest AS build
+
+RUN mkdir -p /go/src/github.com/nsqio/nsq
+COPY . /go/src/github.com/nsqio/nsq
+
+WORKDIR /go/src/github.com/nsqio/nsq
+
+RUN go get github.com/mattn/goveralls \
```
ploxiln: goveralls isn't needed here since ./coverage.sh is not run here

ploxiln:
I found another little thing, but the updates look good, thanks.

Dockerfile
soar: You're right, yeah...

ploxiln:
This looks good to me. I didn't want to be a bother about squashing commits, so I went ahead and did that in #960 

I'll still wait a bit for @mreiferson's approval

ploxiln:
probably worth mentioning: I've tested the nsq_tail multi-topic functionality, but not the docker image build, I don't have a new-enough version of docker/moby set up at the moment

soar:
@ploxiln you don't need Docker to check that It builds successfully, here is automated build on Docker Hub: https://hub.docker.com/r/soarname/nsq/builds/

ploxiln:
Ah, yes, I can pull and inspect that image. It comes out to 57.36 MB uncompressed, and seems to work fine.

ploxiln:
(updated)

soar:
@ploxiln @mreiferson Thank you!

PR #960, opened 2017-11-05
nsq_tail multi-topic and build stuff

ploxiln:
I just squashed and re-titled the commits from #957

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   - GOARCH=amd64
   - GOARCH=386
 sudo: false
+go_import_path: github.com/nsqio/nsq
 before_install:
   - go get github.com/mattn/goveralls
 script:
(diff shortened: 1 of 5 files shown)


ploxiln:
@mreiferson I could also squash all the build-related commits together if you like

ploxiln:
I also added a tweak to the stopping, see the last commit @soar

mreiferson:
@ploxiln LGTM, you want to leave that last commit separate?

mreiferson:
also, I think this is another case where you could push to the original PR's branch instead?

soar:
How this PR got increased coverage with same changes as mine? :-D

ploxiln:
the coverage fluctuates randomly, one of these days I'll get around to figuring out why :)

ploxiln:
I'll squash my fix in. I didn't think of the fact that I might be able to push to the original branch because I'm an admin, I'll try it ...

ploxiln:
I was able to force-push this to #957

PR #975, opened 2017-12-18
avoid client connection reset by peer error

andyxning:
This PR fixes situation where the error response has not been sent completely and the socket is closed. To the  client, the client may return `connection reset by peer error` when trying to read the response with a bad protocol magic string.

With setting [`linger`](https://golang.org/pkg/net/#TCPConn.SetLinger), tcp/ip stack should promises that server has sent all the response before socket is really closed.

This is a race condition actually and is probably to happen when only used with a bad protocol magic.

/cc @mreiferson


judwhite:
@andyxning Isn't this the default behavior? https://golang.org/pkg/net/#TCPConn.SetLinger

PR #984, opened 2018-01-09
nsqd: add --profiling flag

shenlanse:
add flag to indicate whether support pprof or not


mreiferson:
Hi @shenlanse, thanks for this PR!

Have you observed any negative impact from having profiling "available" all the time? What's the use case for this change?

shenlanse:
@mreiferson Thanks for your reply. Haven't found any negative effect when enable profiling, I just want to make it configurable and flexible.

mreiferson:
@shenlanse thanks for following up, sorry for the delay.

@ploxiln's [response](https://github.com/nsqio/nsq/pull/985#issuecomment-356808189) mirrors how I feel about this change set. I don't think the configurability is ultimately worth it.

Closing.

PR #985, opened 2018-01-10
nsqadmin: add notifyClient field for NSQAdmin

shenlanse:
we should reuse httpclient instead of creating new one for every request

### DIFF
--- a/nsqadmin/nsqadmin.go
+++ b/nsqadmin/nsqadmin.go
@@ -28,6 +28,7 @@ type NSQAdmin struct {
 	notifications       chan *AdminAction
 	graphiteURL         *url.URL
 	httpClientTLSConfig *tls.Config
+	notifyClient        *http.Client
 }
 
 func New(opts *Options) *NSQAdmin {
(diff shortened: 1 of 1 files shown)


ploxiln:
This was considered "not performance critical" - there will usually be a relatively long time between administrative actions, so connections will usually not be re-used anyway. Did you experience a performance or operation/network problem due to these notification requests from nsqadmin?

shenlanse:
Have not found performance problem, but i still think we shouldn't create new http client for each request. If add new field to NSQAdmin is unnecessary, global var also works.

ploxiln:
A global var would be awkward because the http.Client's Transport is configured with options on the NSQAdmin.

I don't completely object to this, nor do I completely object to #984. I'd just like to explain: more efficiency is good, but fewer lines of code is also good, fewer options is good (and less complexity in general). We do add additional lines of code and additional options, but only when they bring "enough" benefit (which is subjective).

So in the case of #984, our assumption is that the pprof endpoints have zero runtime cost if you don't access them. So the cost of 5 lines of (actual non-whitespace non-comment etc) code is greater than the zero runtime cost of those endpoints existing but never being used. (And because the possibility of those endpoints is compiled-in either way, there is no binary-size savings.)

In the case of this PR, there's a similar low cost to your change - it moves a variable which is present in the scope of only one function, to be present in the shared state struct (or globally), and needs an extra line of code to manage that. That cost is admittedly low. But how much efficiency is it buying? We would argue that the efficiency gain is pretty close to zero because the notification requests are expected to be so infrequent.

But I suppose that, for some large clusters and large teams, there might regularly be multiple notifications per minute ... and maybe it's nicer to the notification service if nsqadmin re-uses a tcp/tls connection ... so I think the argument in favor of this PR is stronger than the argument in favor of #984. (Also the extra command-line flag in #984 "costs" a lot more in clutter/attention than the extra line in this PR.)

shenlanse:
@ploxiln Thanks for your patient reply. I agree that simple is the best.

mreiferson:
I'm inclined to also close this one as well, given that this isn't a performance critical section and I would rather err on the side of code that's already stable and working.

PR #987, opened 2018-01-19
Update assertions.go

omz-one:
internal/test/assertions.go:40: wrong number of args for format in Logf call: 2 needed but 3 args


ploxiln:
hehe - it makes sense we wouldn't notice this if tests pass. thanks!

Issue #899, opened 2017-05-23
docs: e2e_processing_latency_percentiles settings


JakeXu:
From the github, I found the "ndsq.cfg.example", it wrote like this:
## message processing time percentiles to keep track of (float)
e2e_processing_latency_percentiles = [
    100.0,
    99.0,
    95.0
]

But, on the official document, Command Line Options describe:
-e2e-processing-latency-percentile value
    message processing time percentiles (as float (0, 1.0]) to track (can be specified multiple times or comma separated '1.0,0.99,0.95', default none)

Which is right? Tks

ploxiln:
Both are correct. If you set this in the config file, you use an array. If you set this with a command-line flag, you use commas.

https://github.com/nsqio/nsq/blob/master/nsqd/options.go#L64
https://github.com/mreiferson/go-options/blob/master/options.go#L192

jehiah:
@ploxiln I think the question here is are the values in the range `0, 1.0` or `0, 100`

ploxiln:
Ah, that is a good question. I had to hunt all the way into https://github.com/bmizerany/perks/ to be sure. Each value needs to be in the range `(0,1.0)` (note that this is not inclusive of 1.0, there's a `/ (1 - q)` in bmizerany/perks).

I'm also not sure if the command-line flag actually works, it doesn't look like it copies the percentiles option to `opts`:

```go
	// End to end percentile flags
	e2eProcessingLatencyPercentiles := app.FloatArray{}
	flagSet.Var(&e2eProcessingLatencyPercentiles, "e2e-processing-latency-percentile", "message processing time percentiles (as float (0, 1.0]) to track (can be specified multiple times or comma separated '1.0,0.99,0.95', default none)")
	flagSet.Duration("e2e-processing-latency-window-time", opts.E2EProcessingLatencyWindowTime, "calculate end to end latency quantiles for this duration of time (ie: 60s would only show quantile calculations from the past 60 seconds)")
```

ploxiln:
I take back that bit about not being sure the command-line flag works, go-options gets the reference from the flagSet Var.

Anyway, a documentation/examples update should be sufficient to resolve this issue.

mreiferson:
fixed in #988

PR #988, opened 2018-01-19
nsqd: Validate E2E percentiles & fix the config example

protoss-player:
Passing values greater than 1 makes `quantile` library go nuts and consume lots of memory and CPU after a couple of minutes under ~1000 messages/second. I believe `nsqd` should exit with error instead of pretending that everything's OK in that case.

Also, example config contradicted the nsqd's CLI, which stated that percentile values should be from `(0.0; 1.0]`.

----

Addresses #899.

### DIFF
--- a/contrib/nsqd.cfg.example
+++ b/contrib/nsqd.cfg.example
@@ -87,9 +87,9 @@ statsd_mem_stats = true
 
 ## message processing time percentiles to keep track of (float)
 e2e_processing_latency_percentiles = [
-    100.0,
-    99.0,
-    95.0
+    1.0,
+    0.99,
+    0.95
 ]
 
 ## calculate end to end latency quantiles for this duration of time (time.Duration)
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -148,6 +148,13 @@ func New(opts *Options) *NSQD {
 	}
 	n.tlsConfig = tlsConfig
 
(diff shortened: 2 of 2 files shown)


apps/nsqd/nsqd.go
```diff
@@ -220,6 +220,13 @@ func (p *program) Start() error {
 	cfg.Validate()
 
 	options.Resolve(opts, flagSet, cfg)
+
+	for _, v := range opts.E2EProcessingLatencyPercentiles {
+		if v <= 0 || v > 1 {
```
ploxiln: this line:

https://github.com/bmizerany/perks/blob/master/quantile/stream.go#L62

makes me think that 1.0 might not be an acceptable value ... though I don't understand the algorithm and can't say for sure

protoss-player: It seems like 1.0 is legit, since the math you pointed at evaluates to `+Inf` in this case, and thus should not affect the `if f < m` logic a couple of lines below.

ploxiln: I had to run a little experiment of my own to be sure ... dividing by zero here indeed does result in +Inf instead of NaN or a trap or something.

One more thing: the rest of the argument checking is done in `NSQD.New()` (in `nsqd/nsqd.go`) so this should probably go there.

protoss-player: Done.

ploxiln:
Thanks!

PR #991, opened 2018-01-29
nsq_tail: mv stdout logging to stderr

jehiah:
This fixes a regression from #957 which added a `fmt.Printf` line which outputs to stdout. `stdout` is used exclusively by `nsq_tail` to output data, and logging should be on `stderr` via the `log` package.


mreiferson:
whooops

ploxiln:
thanks 👍

PR #1001, opened 2018-02-27
 Can I do this if I want to store push and consumption messages for messages? Are there any good suggestions?

LvBay:
Initially I thought about wrapping go-nsq code so I did not need to change nsqd and go-nsq, but I found that after the producer pushed the message, I could not get the message_id, so I could only change nsqd's code It's
Although my modified code works fine and meets my needs, I do not expect my code to go through the merge, just hope to get better advice.


ploxiln:
The same message ID can easily be generated for two different topics. It's also possible for the same message ID to be generated by two different nsqds, they could get the same node ID.  And we have considered different, possibly more efficient ways, of generating the message ID. We only want to guarantee that a message ID is unique within a particular channel on a particular nsqd instance. So if you depend on unique message IDs, we generally recommend that you generate it yourself, and put it in the message body.

LvBay:
thanks for the reply.
I did not really think about the message_id that would be repeated.
According to your suggestion, if I generate messageId by myself, I do not need to modify the code of nsqd. Publish_log and consume_log can be stored in mysql by simply wrapping go-nsq.
pretty good idea! @ploxiln

PR #1004, opened 2018-02-28
nsqlookupd: better API when used as a library

### LINKED ISSUE #1003: nsqlookupd: Main() has no way to expose its errors and calls os.Exit
Using NSQ as a library, including lookupd, is a little tricky to handle error situations, like if you have something else on the same port. It doesn't expose the error beyond a log, and then it calls os.Exit, so importers have no chance to handle the error themselves and it brings down the rest of the program. (This caused a problem for me in test code, where I typically need `defer`s to run to do some cleanup. :) )

I'm proposing a new top-level function that will perform most of what `Main` does now, but has a new name and returns errors instead of logging and exiting. That function could be named `MainErr`. `Main`, then, will call `MainErr` and then use the return value to recreate its current behavior.

stephens2424:
for package importers to get error info, and to allow importers avoid a call to os.Exit.

fixes #1003

### DIFF
--- a/apps/nsqlookupd/nsqlookupd.go
+++ b/apps/nsqlookupd/nsqlookupd.go
@@ -77,7 +77,11 @@ func (p *program) Start() error {
 	options.Resolve(opts, flagSet, cfg)
 	daemon := nsqlookupd.New(opts)
 
-	daemon.Main()
+	err := daemon.Main()
+	if err != nil {
+		log.Fatalf("ERROR: failed to start nsqlookupd: %v", err)
+	}
+
 	p.nsqlookupd = daemon
 	return nil
 }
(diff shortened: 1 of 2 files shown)


mreiferson:
I'm fine with cleaning this up, and I don't particularly care about backwards compatibility (which we haven't traditionally guaranteed when using NSQ as a library).

Instead I would suggest making `Main()` actually behave appropriately by returning an error and promote the logging and exiting to the `package main` code paths.

ploxiln:
looks like this was updated (it helps to comment after updating the PR branch, because sometimes github doesn't notify for that)

mreiferson:
yea, this LGTM

mreiferson:
ok, fixed up this commit msg

PR #1005, opened 2018-03-05
nsqadmin: fix delfate->deflate typo in attributes

arussellsaw:


ploxiln:
thanks!

ploxiln:
... maybe immediately was a mistake, because this fix won't actually take effect until the "bindata" is regenerated ... sorry @mreiferson

PR #1008, opened 2018-03-06
nsqd: nsqlookupd connect callback simpler, more robust

ploxiln:
Remove syncTopicChan and do the initial REGISTER commands in
the lookupPeer connectCallback(). This should be mostly
equivalent, just with less indirection.

Close() the lookupPeer in all error cases. It will attempt to
re-connect with the next Ping command.

This stuff was brought up in #993 - consider this an un-tested proposal. Please question/comment :)

### DIFF
--- a/nsqd/lookup.go
+++ b/nsqd/lookup.go
@@ -12,7 +12,7 @@ import (
 	"github.com/nsqio/nsq/internal/version"
 )
 
-func connectCallback(n *NSQD, hostname string, syncTopicChan chan *lookupPeer) func(*lookupPeer) {
+func connectCallback(n *NSQD, hostname string) func(*lookupPeer) {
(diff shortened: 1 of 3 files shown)


mreiferson:
I can't think of any reason why this would be a bad idea!

mreiferson:
... mostly this is because I can no longer remember why we wrote it with such indirection. My assumption is that it's just really old Go code of ours that had way too much use of goroutines and channels 😀

ploxiln:
Rebased and squashed in the little fixups. I'm going to deploy this out to my infrastructure to let it stew for a bit (but honestly my nsq cluster is not very big or stressed).

ploxiln:
I've been running this on a few servers for about a week, using this test-integration branch: https://github.com/ploxiln/nsq/commits/test_201804

PR #1011, opened 2018-03-19
nsqd: replace util.UniqRands

vearne:
The performance of util.UniqRands is not good, I re-implemented
ready for review

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -68,3 +68,7 @@
 [[constraint]]
   name = "golang.org/x/sys"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/vearne/randomchoice"
+  branch = "adc98c957c116a968aec4b0d525adcb386d03e03"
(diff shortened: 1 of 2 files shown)


mreiferson:
Hey @vearne, thanks for the PR!

While we appreciate the contribution, the code path that uses `util.UniqRands` isn't a performance bottleneck, so adding an external dependency to speed it up isn't something we're likely to merge.

If you're interested in performance optimization, I would encourage you to benchmark and provide that evidence as part of future pull requests.

ploxiln:
The external dependency is only about 25 lines of actual code. I recognize it as the card-deck-shuffle algorithm (forget the real name), and I do think it's better than the retry-until-unused-number algorithm currently used.

You should improve UniqRands instead of adding an import. And you should show a benchmark of UniqRands before and after (for around 20 channels). Like @mreiferson I don't expect that this will show any measurable performance impact to overall runtime of normal nsqd processes, but I do think it would be a small improvement to the codebase.

(Also, I think variable names `l` and `n` are less confusing than the names `count` and `choiceCount`)

mreiferson:
> You should improve UniqRands instead of adding an import.

Sure, I'd be fine with this (and even renaming it if that's appropriate).

PR #1020, opened 2018-04-02
nsqd: refactoring channel code

feifeiiiiiiiiiii:
can use defer replace? use defer affect performance?


ploxiln:
Defer does affect performance (though less than it used to). We do use defer sometimes, if the function is complicated, and handling in all exit paths would be messy. But these functions are very simple.

feifeiiiiiiiiiii:
@ploxiln thanks for you answer

judwhite:
@ploxiln Related, we should probably make `gofmt -l -s -d .` part of `test.sh`

PR #1024, opened 2018-04-13
remove locking around listeners for nsqd, nsqlookupd, nsqadmin

ploxiln:
not really needed if we create all listeners before starting servers
and run Main() functions to start these synchronously in tests

(I'm not 100% confident though, I could be wrong ...)


mreiferson:
this also LGTM, can't think of any reason why the locks would be necessary after this rewrite

do you want to merge this into the branch you're validating for #1008, enable race detection, and assuming we see no reports we can land?

ploxiln:
I've been running this on a few servers for about a week, using this test-integration branch: https://github.com/ploxiln/nsq/commits/test_201804

PR #1029, opened 2018-04-28
Add Fandom logo

ljagiello:
Fandom proudly used NSQ in production services.

(connected with https://github.com/nsqio/nsqio.github.io/pull/37).


mreiferson:
Thanks!

Issue #1031, opened 2018-05-09
nsqd: message loss during topic/channel bootstrap from metadata


michaelyou:
Hi:

I found the code start the `t.messagePump` goroutine before all the channels of the topic are created. If there are messages in the diskqueue, some of channels created later may not receive the message.

I verified it by create 10 channels and 2 messages in the diskqueue
![image](https://user-images.githubusercontent.com/8854914/39817414-aa38e682-53d0-11e8-90dc-ea9bf6419327.png)

mreiferson:
see #1032

PR #1032, opened 2018-05-10
nsqd: message loss during topic/channel bootstrap from metadata

michaelyou:
#1031 

I am not sure will this introduce other bugs~

This is what the pr do：

Initialize topic with `pause=1`, so when start `t.messagePump` goroutine，it won't read msg out. 
In the method `GetTopic`，after `NewTopic` return, check `n.isLoading` to see if it is loading from metadata file(restart). If not, unpause it immediately, else, unpause it when all channels are created in `LoadMetadata `. 

`NewTopic` is only called by `GetTopic`， so the change should not affect other code.

------

image

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -358,9 +358,6 @@ func (n *NSQD) LoadMetadata() error {
 			continue
 		}
 		topic := n.GetTopic(t.Name)
-		if t.Paused {
-			topic.Pause()
-		}
 
 		for _, c := range t.Channels {
 			if !protocol.IsValidChannelName(c.Name) {
@@ -371,6 +368,12 @@ func (n *NSQD) LoadMetadata() error {
 			if c.Paused {
 				channel.Pause()
 			}
+		}
+
+		// this logic is reversed, and done _after_ channel creation to ensure that all channels
+		// are created before messages begin flowing
+		if !t.Paused {
+			topic.UnPause()
 		}
 	}
 	return nil
@@ -514,6 +517,13 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	}
 	t = NewTopic(topicName, &context{n}, deleteCallback)
 	n.topicMap[topicName] = t
+
+	// if we're creating a new topic in this process while we're loading from metadata, the topic may already
+	// have message data persisted on disk. To ensure that all channels are created before message flow
+	// begins, we pause the topic.
+	if atomic.LoadInt32(&n.isLoading) == 1 {
+		t.Pause()
+	}
 
 	n.logf(LOG_INFO, "TOPIC(%s): created", t.name)
 


nsqd/nsqd.go
```diff
@@ -372,6 +372,7 @@ func (n *NSQD) LoadMetadata() error {
 				channel.Pause()
 			}
 		}
+		topic.UnPause()
```
ploxiln: Above, you'll see that paused state is stored in the topic metadata, so it does:

```go
 if t.Paused {
 	topic.Pause()
}
```

So you can remove that now, but here only un-pause if the topic was not originally paused.

nsqd/nsqd.go
```diff
@@ -515,6 +516,10 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	t = NewTopic(topicName, &context{n}, deleteCallback)
 	n.topicMap[topicName] = t
 
+	if atomic.LoadInt32(&n.isLoading) == 0 {
+		t.UnPause()
```
ploxiln: I wonder if we can move this below the channels block, as an alternative to taking the topic lock to block PutMessages() while getting channels.

nsqd/nsqd.go
michaelyou: Thanks, i will correct it

nsqd/nsqd.go
michaelyou: The `channelMap` in `getOrCreateChannel` needs the topic's lock . I am not sure if we can do this~

```go
// this expects the caller to handle locking
func (t *Topic) getOrCreateChannel(channelName string) (*Channel, bool) {
    channel, ok := t.channelMap[channelName]
```

mreiferson:
I wonder if we should just create a `NewTopicWithChannels` function that allows this to be handled in the correct order before goroutines are launched.

nsqd/nsqd.go
ploxiln: Right - you'd call `t.GetChannel()` instead. But this is still tentative supposition :)

michaelyou:
@mreiferson Creating a `NewTopicWithChannels ` looks like a big change. I modified my commit and does it now look more natural~

mreiferson:
It will be a slightly bigger change, but I feel like this code is a _little_ too clever and surprising. We could comment why the ordering is what it is, but I feel like I'd rather have slightly more code that's clear in its intent?

What do you think @ploxiln?

ploxiln:
I think the `LoadMetadata()` unpause-after style is reasonably elegant, and maybe we could improve the locking/responsiveness of `GetTopic()` with that style. A `NewTopicWithChannels()` would not work for `GetTopic()` (because the nsqd lock would have to be held while querying lookupd). But I don't have a strong opinion - just adding a `NewTopicWithChannels()` just for `LoadMetadata()` would work fine too.

ploxiln:
It looks like this changed since I last viewed it - instead of initializing a Topic with `paused: true` in `NewTopic()`, it calls `Topic.Pause()` just before releasing the nsqd lock in `GetTopic()` iff `nsqd.isLoading`. It seems to me like that works too.

Anyway, I think this is a fine minimal fix-up, and further refactoring can be proposed in a separate PR. @mreiferson?

nsqd/nsqd.go
```diff
@@ -515,6 +515,10 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	t = NewTopic(topicName, &context{n}, deleteCallback)
 	n.topicMap[topicName] = t
 
+	if atomic.LoadInt32(&n.isLoading) == 1 {
```
mreiferson: Let's add a comment above this line that reads:

```
// if we're creating a new topic in this process while we're loading from metadata, the topic may already
// have message data persisted on disk. To ensure that all channels are created before message flow
// begins, we pause the topic.
```

nsqd/nsqd.go
```diff
@@ -372,6 +369,9 @@ func (n *NSQD) LoadMetadata() error {
 				channel.Pause()
 			}
 		}
+		if !t.Paused {
```
mreiferson: And a comment above this line:

```
// this logic is reversed, and done _after_ channel creation to ensure that all channels
// are created before messages begin flowing
```

michaelyou:
@mreiferson done

ploxiln:
Thanks for squashing down to one commit. The final little issue is that the commit title is not descriptive of the overall change. I suggest:

> nsqd: pause topics while loading metadata

michaelyou:
@ploxiln OK, done

ploxiln:
Thanks!

penghuazhou:
@ploxiln, why the pr code not in master now?

penghuazhou:
@michaelyou, why the pr code not in master now?

ploxiln:
This PR is from 2018, likely the code around this area changed in the following years.

penghuazhou:
thanks

PR #1036, opened 2018-05-20
nsqlookupd: change Producers type to map

andyxning:
This PR will change the value type for `registrationMap` from slice to map. This is mainly used to accelerate the lookup performance in case we have more nsqd producers and more topics/channels in a cluster. 

In our case, once the nsqd producers in a cluster is more than 600 and the topics/channels number exceeds 5000, the execution time for `LookupRegistrations` will be more than 100ms which is too long as it holds the read lock while running. This will block nsqd producers from registering themselves for they can not acquire the write lock quickly.

### DIFF
--- a/nsqlookupd/registration_db.go
+++ b/nsqlookupd/registration_db.go
@@ -9,7 +9,7 @@ import (
 
 type RegistrationDB struct {
 	sync.RWMutex
-	registrationMap map[Registration]Producers
+	registrationMap map[Registration]ProducerMap
 }
 
 type Registration struct {
@@ -37,6 +37,7 @@ type Producer struct {
 }
 
 type Producers []*Producer
+type ProducerMap map[string]*Producer
 
 func (p *Producer) String() string {
 	return fmt.Sprintf("%s [%d, %d]", p.peerInfo.BroadcastAddress, p.peerInfo.TCPPort, p.peerInfo.HTTPPort)
@@ -53,7 +54,7 @@ func (p *Producer) IsTombstoned(lifetime time.Duration) bool {
 
 func NewRegistrationDB() *RegistrationDB {
 	return &RegistrationDB{
-		registrationMap: make(map[Registration]Producers),
+		registrationMap: make(map[Registration]ProducerMap),
 	}
 }
 
@@ -63,7 +64,7 @@ func (r *RegistrationDB) AddRegistration(k Registration) {
 	defer r.Unlock()
 	_, ok := r.registrationMap[k]
 	if !ok {
-		r.registrationMap[k] = Producers{}
+		r.registrationMap[k] = make(map[string]*Producer)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ploxiln:
I'm still digesting this ... but, you could add a small benchmark to demonstrate the performance difference at a few different counts of producers and topics.

andyxning:
> you could add a small benchmark to demonstrate the performance difference at a few different counts of producers and topics.

Yeah. But it need some extra time to do a bench. Theoretically slice is O(N) and map is O(1). But it is true that map will consume more memory. :)

andyxning:
@mreiferson I am thinking how to prepare it after @ploxiln suggests. Will give more benchmark data later. :)

andyxning:
@ploxiln @mreiferson 
I have added two types of benchmark. One for map producers and the other, `FakeRegistrationDB`, for original slice producers.
The benchmark output is running on my laptop(Darwin MBP 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64)
> go test -run=BenchmarkLookupRegistrations -bench=LookupRegistrations -benchtime=5s
goos: darwin
goarch: amd64
pkg: github.com/nsqio/nsq/nsqlookupd
BenchmarkLookupRegistrations64Registration64Producer-4           	 1000000	     16202 ns/op
BenchmarkLookupRegistrations128Registration128Producer-4         	  200000	     50131 ns/op
BenchmarkLookupRegistrations256Registration256Producer-4         	   50000	    119318 ns/op
BenchmarkLookupRegistrations512Registration512Producer-4         	   20000	    337405 ns/op
BenchmarkLookupRegistrations1024Registration1024Producer-4       	   10000	    964775 ns/op
BenchmarkLookupRegistrations2048Registration2048Producer-4       	    5000	   3384047 ns/op
BenchmarkFakeLookupRegistrations64Registration64Producer-4       	  500000	     16967 ns/op
BenchmarkFakeLookupRegistrations128Registration128Producer-4     	  200000	     44824 ns/op
BenchmarkFakeLookupRegistrations256Registration256Producer-4     	   50000	    131509 ns/op
BenchmarkFakeLookupRegistrations512Registration512Producer-4     	   20000	    281686 ns/op
BenchmarkFakeLookupRegistrations1024Registration1024Producer-4   	   10000	    692863 ns/op
BenchmarkFakeLookupRegistrations2048Registration2048Producer-4   	    2000	   2633470 ns/op
PASS
ok  	github.com/nsqio/nsq/nsqlookupd	172.154s

It is obviously that map producers have about 1x performance enhance compared with original slice producers when registrations and producers are both 2048

ploxiln:
Thanks!

I'm going to try to tweak this, so the benchmark addition is in a separate commit, and can work with current master (instead of having a separate mock-implementation to benchmark against).

andyxning:
Any updates? @ploxiln @mreiferson :)

mreiferson:
Wrapping this up in #1038, thanks @andyxning!

andyxning:
@ploxiln @mreiferson Thanks for reviewing and merging this PR. 👏

PR #1038, opened 2018-06-02
nsqlookupd: use map for producers - with benchmarks

ploxiln:
I refactored and added a bit to @andyxning's benchmark from #1036 

Results:

```
benchmark                                  old ns/op     new ns/op     delta
BenchmarkLookupRegistrations8x8-4          1065          1088          +2.16%
BenchmarkLookupRegistrations8x64-4         2264          1165          -48.54%
BenchmarkLookupRegistrations64x64-4        18669         8364          -55.20%
BenchmarkLookupRegistrations64x512-4       134794        10933         -91.89%
BenchmarkLookupRegistrations512x512-4      1698650       150939        -91.11%
BenchmarkLookupRegistrations512x2048-4     5751659       177923        -96.91%
BenchmarkFindProducers8x8-4                244           619           +153.69%
BenchmarkFindProducers8x64-4               243           2259          +829.63%
BenchmarkFindProducers64x64-4              244           2437          +898.77%
BenchmarkFindProducers64x512-4             245           16447         +6613.06%
BenchmarkFindProducers512x512-4            277           16825         +5974.01%
BenchmarkFindProducers512x2048-4           277           83689         +30112.64%
```

take a look @mreiferson

### DIFF
--- a/nsqlookupd/registration_db.go
+++ b/nsqlookupd/registration_db.go
@@ -9,7 +9,7 @@ import (
 
 type RegistrationDB struct {
 	sync.RWMutex
-	registrationMap map[Registration]Producers
+	registrationMap map[Registration]ProducerMap
 }
 
 type Registration struct {
@@ -37,6 +37,7 @@ type Producer struct {
 }
 
 type Producers []*Producer
+type ProducerMap map[string]*Producer
 
 func (p *Producer) String() string {
 	return fmt.Sprintf("%s [%d, %d]", p.peerInfo.BroadcastAddress, p.peerInfo.TCPPort, p.peerInfo.HTTPPort)
@@ -53,7 +54,7 @@ func (p *Producer) IsTombstoned(lifetime time.Duration) bool {
 
 func NewRegistrationDB() *RegistrationDB {
 	return &RegistrationDB{
-		registrationMap: make(map[Registration]Producers),
+		registrationMap: make(map[Registration]ProducerMap),
 	}
 }
 
@@ -63,7 +64,7 @@ func (r *RegistrationDB) AddRegistration(k Registration) {
 	defer r.Unlock()
 	_, ok := r.registrationMap[k]
 	if !ok {
-		r.registrationMap[k] = Producers{}
+		r.registrationMap[k] = make(map[string]*Producer)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ploxiln:
After that, I realized: we usually look-up topics, not channels, which my benchmark was looking up producers for. And it looks like there's some "filtering" code that kicks in sometimes. I should probably benchmark looking up producers of topics.

So I split out topic-lookup benchmarks. The results were pretty much the same as for channel lookup.

But I realized that the typical `/lookup` operation of nsqlookups involved `FindRegistrations()`, and other stuff, so now I figure I should make a benchmark that more closely matches what `doLookup()` actually does ...

ploxiln:
and the results:

```
$ benchcmp old.txt new.txt 
benchmark                                  old ns/op     new ns/op     delta
BenchmarkLookupRegistrations8x8-4          2969          2923          -1.55%
BenchmarkLookupRegistrations8x64-4         6334          3298          -47.93%
BenchmarkLookupRegistrations64x64-4        55507         30026         -45.91%
BenchmarkLookupRegistrations64x512-4       396369        39583         -90.01%
BenchmarkLookupRegistrations512x512-4      5135966       515863        -89.96%
BenchmarkLookupRegistrations512x2048-4     16354360      595718        -96.36%
BenchmarkDoLookup8x8-4                     1477          1885          +27.62%
BenchmarkDoLookup8x64-4                    1462          3584          +145.14%
BenchmarkDoLookup64x64-4                   6189          8770          +41.70%
BenchmarkDoLookup64x512-4                  6387          23065         +261.12%
BenchmarkDoLookup512x512-4                 45920         65297         +42.20%
BenchmarkDoLookup512x2048-4                45398         119196        +162.56%
```

It's probably a good idea to add benchmarks for typical lookup-protocol IDENTIFY and REGISTER operations (mostly `AddProducer()`). `LookupRegistrations()` is only called when a producer disconnects a tcp-lookup-protocol connection. (But it can still be pretty bad for it to lock the RegistrationDB for a long time.)

ploxiln:
Results for the new REGISTER bench:

```
benchmark                                  old ns/op       new ns/op      delta
BenchmarkRegister8x8-4                     80615           79249          -1.69%
BenchmarkRegister8x64-4                    733022          741142         +1.11%
BenchmarkRegister64x64-4                   6035122         6548519        +8.51%
BenchmarkRegister64x512-4                  179271032       52564415       -70.68%
BenchmarkRegister512x512-4                 1505092668      416921240      -72.30%
BenchmarkRegister512x2048-4                17299945707     1770267744     -89.77%
```

andyxning:
@ploxiln Thanks for the more detailed benchmark. The new scheme's decreased performance may be due to the map allocation and adjustment.

But the lookup is called periodically and the register is only called on the start of nsqd and new consumer connections.

mreiferson:
I think I agree, since lookups can also happen in parallel.

Thanks for this — do you want to clean up these commits before we merge?

ploxiln:
squashed

andyxning:
@ploxiln @mreiferson Thanks for reviewing and merging this PR. 👏

PR #1043, opened 2018-06-13
Update Dep to 0.4.1

vitaliytv:
update to stable dep version https://golang.github.io/dep/


ploxiln:
It's a bit random. You can just ignore it if it's obvious your change should not have affected coverage.

vitaliytv:
Thank you

PR #1045, opened 2018-06-14
Update alpine in Dockerfile to latest version (3.7)

vitaliytv:


mreiferson:
@vitaliytv have you confirmed the container builds and runs with this change?

vitaliytv:
Tested on localhost

PR #1047, opened 2018-06-29
nsqd: make statsd SpreadWriter use exit chan

ploxiln:
to avoid potential long delay when exiting nsqd

not sure if we can come up with a more elegant fix
cc @mreiferson


mreiferson:
LGTM

PR #1049, opened 2018-07-04
nsqd: avoid error logs while loading metadata on startup

ploxiln:
lookupd connections are not yet established while adding topics
from the metadata file during start-up - this is OK

Example of the error log messages being avoided, in context:

```
[nsqd] 2018/07/04 13:54:11.811747 INFO: nsqd v1.0.1-alpha (built w/go1.10.3)
[nsqd] 2018/07/04 13:54:11.811879 INFO: ID: 762
[nsqd] 2018/07/04 13:54:11.812360 INFO: TOPIC(tt1): created
[nsqd] 2018/07/04 13:54:11.812378 ERROR: no available nsqlookupd to query for channels to pre-create for topic tt1
[nsqd] 2018/07/04 13:54:11.812523 INFO: TOPIC(tt1): new channel(tc1)
[nsqd] 2018/07/04 13:54:11.812658 INFO: TOPIC(tt2): created
[nsqd] 2018/07/04 13:54:11.812667 ERROR: no available nsqlookupd to query for channels to pre-create for topic tt2
[nsqd] 2018/07/04 13:54:11.812685 INFO: NSQ: persisting topic/channel metadata to nsqd.dat
[nsqd] 2018/07/04 13:54:11.813647 INFO: LOOKUP(127.0.0.1:4160): adding peer
[nsqd] 2018/07/04 13:54:11.813662 INFO: LOOKUP connecting to 127.0.0.1:4160
[nsqd] 2018/07/04 13:54:11.813670 INFO: HTTP: listening on [::]:4151
[nsqd] 2018/07/04 13:54:11.814317 INFO: LOOKUPD(127.0.0.1:4160): peer info {TCPPort:4160 HTTPPort:4161 Version:1.0.1-alpha BroadcastAddress:plo-pro.lan}
[nsqd] 2018/07/04 13:54:11.814342 INFO: LOOKUPD(127.0.0.1:4160): REGISTER tt1 tc1
[nsqd] 2018/07/04 13:54:11.814502 INFO: LOOKUPD(127.0.0.1:4160): REGISTER tt2
[nsqd] 2018/07/04 13:54:11.814661 INFO: LOOKUPD(127.0.0.1:4160): topic REGISTER tt1
[nsqd] 2018/07/04 13:54:11.814765 INFO: LOOKUPD(127.0.0.1:4160): channel REGISTER tt1 tc1
[...]

PR #1050, opened 2018-07-04
nsqd: create Topic paused

ploxiln:
Enables simplifications. Experimental, not fully tested.

Alternative to #1049. cc @mreiferson

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -504,19 +504,15 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	t = NewTopic(topicName, &context{n}, deleteCallback)
 	n.topicMap[topicName] = t
 
-	// if we're creating a new topic in this process while we're loading from metadata, the topic may already
-	// have message data persisted on disk. To ensure that all channels are created before message flow
-	// begins, we pause the topic.
-	if atomic.LoadInt32(&n.isLoading) == 1 {
-		t.Pause()
-	}
+	n.Unlock()
 
 	n.logf(LOG_INFO, "TOPIC(%s): created", t.name)
+	// topic is created paused, only un-pause after adding channels
 
-	// release our global nsqd lock, and switch to a more granular topic lock while we init our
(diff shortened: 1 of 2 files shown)


mreiferson:
Yea, this is definitely better, thanks. It's _still_ pretty gnarly, but I'm not sure how much it's worth messing with it.

ploxiln:
I should probably do some real testing :P

ploxiln:
Initial testing looks good, as expected.

I wonder if there might be some potential for confusion due to this change ... maybe if some script creates a topic then immediately pauses it, that could happen before the automatic un-pause and the the topic could unexpectedly end up unpaused. Or something like that. I don't want to be paranoid about this, just something to keep in mind.

On the other hand, I like how this reduces how long topic locks are held. Though I guess it's a one-time thing.

As a further digression, I know you thought may be a good idea to make a `NewTopic()` that takes channels to pre-create. While that would work for loading metadata, I think that wouldn't work for the runtime-created topics. We would still need some kind of lock for multiple clients (http or tcp) touching the same new topic around the same time. We probably would not want to hold the nsqd lock while querying nsqlookupds, so we'd add a sub-lock ... so we might as well create the topic and use it for coordination ;)

mreiferson:
> I wonder if there might be some potential for confusion due to this change ... maybe if some script creates a topic then immediately pauses it, that could happen before the automatic un-pause and the the topic could unexpectedly end up unpaused. Or something like that. I don't want to be paranoid about this, just something to keep in mind.

That's an interesting edge case. It probably means we're abusing `paused` state when what we really need is tighter control over internals of topic creation, e.g. if we moved starting of `topic.messagePump` outside of `NewTopic` (into some `Start()` func) then we can control when that goro launches and not need to use pausing as a hack?

ploxiln:
Yeah I think that's a good idea - instead of a special first "UnPause", have a separate "Start".

ploxiln:
The tricky bit there is that `Topic.Pause()` and `Topic.GetChannel()` currently block until `messagePump()` consumes the notification (or until the exitChan is closed).

PR #1051, opened 2018-07-07
nsqd: return error if pause state is not changed

ploxiln:
so the caller of pause or unpause can know if the topic or channel
was already paused or unpaused

also avoids a little bit of work if no change

a humble idea, just to solicit opinions, could probably use better error strings

### DIFF
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -253,38 +253,33 @@ func (c *Channel) Depth() int64 {
 	return int64(len(c.memoryMsgChan)) + c.backend.Depth()
 }
 
-func (c *Channel) Pause() error {
+func (c *Channel) Pause() bool {
 	return c.doPause(true)
 }
 
-func (c *Channel) UnPause() error {
+func (c *Channel) UnPause() bool {
 	return c.doPause(false)
 }
 
-func (c *Channel) doPause(pause bool) error {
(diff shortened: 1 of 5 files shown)


mreiferson:
I guess it's useful to propagate this information?

ploxiln:
Yeah I though it could be useful info to get in the response - sort of inspired by the corner-case of maybe having a script retry pause a few times until it takes effect. Maybe not worth it.

There's also the interesting bit where Pause() / UnPause() return type error, but (until this change) it is _always_ nil. They could return the previous state as a bool instead. Or nothing. Dunno if any software out there really uses nsqd guts as a library, probably not, and we're breaking "api behavior" with the topic pause-or-start changes anyway ...

mreiferson:
Ready to land this?

ploxiln:
hang on, I can add a proper test pretty easily

ploxiln:
I'm thinking about maybe returning 200 but with a body that says e.g. "already paused" somehow. May be less disruptive to existing scripts. Which may or may not even exist. I dunno.

Anyway I'd like to defer this PR until after #1053 is resolved.

ploxiln:
I pushed up a "non-error" variant. But I'm thinking it's best to just close this until someone asks for this feature. (I don't really need it myself.)

PR #1053, opened 2018-07-07
nsqd: start Topic messagePump separately from create

ploxiln:
with new Topic.Start()

and make Topic.Pause() and Topic.GetChannel() no longer block
on the notification channels for Topic.messagePump()

and simplify some calls to WaitGroupWrapper.Wrap()

follow-up to #1050 cc @mreiferson 
possibly better, but does change the behavior of Pause() and GetChannel() slightly ...

### DIFF
--- a/nsqadmin/nsqadmin.go
+++ b/nsqadmin/nsqadmin.go
@@ -177,7 +177,7 @@ func (n *NSQAdmin) Main() {
 	n.waitGroup.Wrap(func() {
 		http_api.Serve(n.httpListener, http_api.CompressHandler(httpServer), "HTTP", n.logf)
 	})
-	n.waitGroup.Wrap(func() { n.handleAdminActions() })
+	n.waitGroup.Wrap(n.handleAdminActions)
 }
 
 func (n *NSQAdmin) Exit() {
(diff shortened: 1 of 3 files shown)


mreiferson:
> and make Topic.Pause() and Topic.GetChannel() no longer block on the notification channels for Topic.messagePump()

I was wondering if this would cause random timing-related test failures, looks like it hasn't. However, I'm a _little_ concerned about the subtlety of this change.

For `Pause()`, I'm thinking that for topic creation code paths, `GetTopic` and `NewTopic` should really take `paused` as a parameter in their signatures. This way we wouldn't have to call `Paused` immediately after to set appropriate state (code would look a _little_cleaner too, I think). I realize we'll have to change 1000 places in tests we call GetTopic, but I _think_ I'd prefer that over changing the behavior of the notification.

I don't have any great ideas for `GetChannel` though, what do you think?

ploxiln:
I don't like the idea of adding a `paused` parameter to `GetTopic()` (only to be used if it creates a new topic?)

However, I came up with another design. `Topic.messagePump()` is started by `NewTopic()` as before, but now it has a pre-start-loop to recv on `pauseChan` and `channelUpdateChan`, and which is ended via a new `startChan`.

mreiferson:
Nice, I like it. LGTM, squash?

ploxiln:
Will squash and do some local testing ...

ploxiln:
updated

ploxiln:
thanks for the review!

PR #1055, opened 2018-07-19
nsqd: remove support for old metadata filename

ploxiln:
I recall we wanted to remove this sometime after the "1.0.0-compat" release. I don't feel strongly about removing it now vs leaving it in for one more release.

cc @mreiferson @jehiah


mreiferson:
I think we should keep this in one more release.

PR #1057, opened 2018-07-19
update dist.sh docker build

ploxiln:
no longer uses local cross-built binaries
(now built in separate docker container)

"-f" flag for "docker tag" is obsolete

also misc checklist/instructions updates
(homebrew update not urgent IMHO, someone else will probably take care of it in a week anyway)

cc @mreiferson @jehiah


mreiferson:
LGTM

PR #1062, opened 2018-07-31
nsqadmin: update JS dependencies

jehiah:
This ticket is to update the nsqadmin JS build dependencies


jehiah:
new jquery; that's probably of most note here; everything seems to still work so this should be RFM when CI finishes

mreiferson:
we write such high-quality javascript, that's gotta be the reason

PR #753, opened 2018-07-31
nsqadmin: missing channel from topic list

jehiah:
the api endpoint `/api/topics/:topic` seems to be missing a channel that i see in `/stats` output from nsqd's. I'm not sure what the bug is yet, but this causes it to be missing from the nsqadmin UI at `/topics/:topic` even though it shows properly if you manually navigate to `/topics/:topic/:channel`.


jehiah:
RFR @mreiferson @ploxiln 

I was able to reproduce this bug with the following setup

Node A: topic: T channels a,b,c
Node B: topic: T channels a.  

With that setup the count was previously 3 on both nodes, and depending on the order `/stats` responses completed you would see all three channels listed on the topics page, or just one.

internal/clusterinfo/types.go
```diff
@@ -115,8 +115,8 @@ func (t *TopicStats) Add(a *TopicStats) {
 	if a.Paused {
 		t.Paused = a.Paused
 	}
-	found := false
 	for _, aChannelStats := range a.Channels {
+		found := false
```
ploxiln: nice fix 👍

nsqadmin/static/js/views/topic.hbs
```diff
@@ -81,7 +81,7 @@
             {{#if ../graph_active}}
                 <td class="bold rate" target="{{rate "topic" node topic_name ""}}"></td>
             {{/if}}
-            <td>{{commafy ../channels.length}}</td>
+            <td>{{commafy this/channels.length}}</td>
```
ploxiln: this is greek to me :)
I wonder if "this/" is not needed and just "channels.length" would suffice?

jehiah: I'm not sure... I confused myself by working in the wrong part of the template for a while, but there is a global `channels` var so this is at least more explicit.

internal/clusterinfo/types.go
jehiah: it only took me about a hundred printf's to find it =)

nsqadmin/static/js/views/topic.hbs
ploxiln: fair enough :) no objection

PR #1065, opened 2018-07-31
nsqd: reset conn deadline after TLS upgrade

### LINKED ISSUE #1027: nsqd: connection EOF after TLS handshake
greetings,

Producer connections closed with EOF, after connection upgrades to TLS 。
Problem disappeared after I commented out `tlsConn.SetDeadline(time.Now().Add(5 * time.Second))` in `Client_v2`.

My guess is, read deadline is reset to heartbeat *3 in IOLoop however write deadline specified in `Client_v2.UpgradeTLS` still in effect, and connection returns EOF after that deadline

Maybe a reset of writedeadline after UpgradeTLS, or in IOLoop?

mreiferson:
fixes #1027

### DIFF
--- a/nsqd/client_v2.go
+++ b/nsqd/client_v2.go
@@ -473,6 +473,9 @@ func (c *clientV2) UpgradeTLS() error {
 	}
 	c.tlsConn = tlsConn
 
+	var zeroTime time.Time
+	tlsConn.SetDeadline(zeroTime)
+
 	c.Reader = bufio.NewReaderSize(c.tlsConn, defaultBufferSize)
 	c.Writer = bufio.NewWriterSize(c.tlsConn, c.OutputBufferSize)
 


nsqd/client_v2.go
```diff
@@ -473,6 +473,9 @@ func (c *clientV2) UpgradeTLS() error {
 	}
 	c.tlsConn = tlsConn
 
+	var zeroTime time.Time
+	tlsConn.SetDeadline(zeroTime)
```
ploxiln: I think you can also do `tlsConn.SetDeadline(time.Time{})` (but what you have is probably more clear)

mreiferson:
I _think_ this is correct, but I decided to try to write a failing test and I'm having trouble reproducing the initial report.

@DoraALin do you have some reproducible configuration that we can test out?

mreiferson:
Digging through the code, I don't think the change here is actually necessary.

The read deadline is reset after a TLS upgrade [here](https://github.com/nsqio/nsq/blob/master/nsqd/protocol_v2.go#L54-L58) and the write deadline is reset [here](https://github.com/nsqio/nsq/blob/master/nsqd/protocol_v2.go#L143-L148) as a result of sending the `OK` response from the upgrade.

This would seem to indicate that the report in #1027 was actually some sort of TLS negotiation latency that hit the arbitrary 5s deadline. Note that @DoraALin's solution was to _comment out the deadline_, not reset it after TLS negotiation.

Perhaps the real fix is to make the TLS deadline configurable.

PR #1068, opened 2018-08-03
nsqadmin: html-escaping large_graph url breaks inside javascript

ploxiln:
ends up like `/render?colorList&#x3D;green&amp;height&#x3D;450&amp;...`
`=` was not escaped to `&#x3D;` before
this starts the url hash which is not sent to the server

I'm still validating that this change actually fixes it
cc @jehiah


ploxiln:
alrighty, it works now. I'll squash down (all but the `.eslintrc`) if you approve

ploxiln:
side note, I currently have this mini script I run as `./go-bindata`:

```bash
#!/bin/sh
set -eu
go-bindata --modtime=1533016989 --pkg=nsqadmin "$@" --prefix=static/build static/build/...
gofmt -w bindata.go
```

mreiferson:
cool

ploxiln:
squashed a bit

PR #1071, opened 2018-08-15
nsq_to_file: new flag -rotate-check-interval

### LINKED ISSUE #1070: nsq_to_file: make the interval at which we check for log rotation configurable
Currently `nsq_to_file` checks whether to rotate the output file every 30s, a hard-coded interval.  It would be convenient if this quantity were customizable.

alpaker:
This PR adds a new CLI option `rotate-check-interval` (default 30s) to `nsq_to_file` that controls the frequency at which FileLogger checks whether a file should be rotated.  Fixes #1070.

### DIFF
--- a/apps/nsq_to_file/nsq_to_file.go
+++ b/apps/nsq_to_file/nsq_to_file.go
@@ -34,6 +34,7 @@ var (
 
 	rotateSize     = flag.Int64("rotate-size", 0, "rotate the file when it grows bigger than `rotate-size` bytes")
 	rotateInterval = flag.Duration("rotate-interval", 0*time.Second, "rotate the file every duration")
+	rotateCheckInterval = flag.Duration("rotate-check-interval", 30*time.Second, "check whether the file should be rotated every duration")
 
 	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
 	httpRequestTimeout = flag.Duration("http-client-request-timeout", 5*time.Second, "timeout for HTTP request")
--- a/apps/nsq_to_file/file_logger.go
+++ b/apps/nsq_to_file/file_logger.go
@@ -82,7 +82,7 @@ func (f *FileLogger) router(r *nsq.Consumer) {
 	pos := 0
 	output := make([]*nsq.Message, *maxInFlight)
 	sync := false
-	ticker := time.NewTicker(time.Duration(30) * time.Second)
+	ticker := time.NewTicker(*rotateCheckInterval)
 	closing := false
 	closeFile := false
 	exit := false


alpaker:
@mreiferson

ploxiln:
It's worth noting that `nsq_to_file` already checks if the file should be rotated just before writing each message:

```go
 		case <-ticker.C:
 			if f.needsFileRotate() {
 				...
 		case m := <-f.logChan:
 			if f.needsFileRotate() {
 				...
```

So the ticker, and this change, just affects how quickly the "last" log file is closed when there are no new messages, and the generation of empty files.

mreiferson:
thanks @alpaker!

@ploxiln for some context, we ran into some issues with extremely low volume topics and the files not being flushed/closed before being accessed.

mreiferson:
I should probably stamp that stable release before landing this?

ploxiln:
Ah, so this is for the `sync = true` effect.

Yeah, I think we can stamp the stable release, and then merge this and the other PRs you've revived recently.

mreiferson:
thanks again @alpaker!

apps/nsq_to_file/nsq_to_file.go
```diff
@@ -34,6 +34,7 @@ var (
 
 	rotateSize     = flag.Int64("rotate-size", 0, "rotate the file when it grows bigger than `rotate-size` bytes")
 	rotateInterval = flag.Duration("rotate-interval", 0*time.Second, "rotate the file every duration")
+	rotateCheckInterval = flag.Duration("rotate-check-interval", 30*time.Second, "check whether the file should be rotated every duration")
```
ploxiln: I think name and description is not really right - the check whether the file should be rotated is done on every single message, this really just affects how often the files (including empty files I guess) are synced to disk

mreiferson: Hmmmm, what do you have in mind?

mreiferson: Yea, I agree, `--sync-interval` makes more sense. I can whip that up or do you wanna do the honors @alpaker?

alpaker: Sure.  Regarding the description: If we're going to foreground the flush interval as what's controlled by this option, then I suggest omitting mention of rotation altogether.  That the need for rotation is checked on msg receipt isn't explicitly documented, which makes the significance of "check if file should should have been rotated since the last message received" unclear.  How about: 
```
syncInterval = flag.Duration("sync-interval", 30*time.Second, "flush messages to disk every duration")
```

mreiferson: Sounds good, except `s/flush/sync` to match the flag name?

ploxiln: all looks good to me, I was just trying to appease the reasoning behind the original naming :)

PR #1072, opened 2018-08-19
nsq_to_http: add custom header flag

alwindoss:
Signed-off-by: Alwin Doss <alwindoss84@gmail.com>

### DIFF
--- a/apps/nsq_to_http/nsq_to_http.go
+++ b/apps/nsq_to_http/nsq_to_http.go
@@ -49,14 +49,17 @@ var (
 	statusEvery        = flag.Int("status-every", 250, "the # of requests between logging status (per handler), 0 disables")
 	contentType        = flag.String("content-type", "application/octet-stream", "the Content-Type used for POST requests")
 
-	getAddrs         = app.StringArray{}
-	postAddrs        = app.StringArray{}
-	nsqdTCPAddrs     = app.StringArray{}
-	lookupdHTTPAddrs = app.StringArray{}
+	getAddrs           = app.StringArray{}
+	postAddrs          = app.StringArray{}
+	customHeaders      = app.StringArray{}
+	nsqdTCPAddrs       = app.StringArray{}
+	lookupdHTTPAddrs   = app.StringArray{}
+	validCustomHeaders map[string]string
 )
 
 func init() {
 	flag.Var(&postAddrs, "post", "HTTP address to make a POST request to.  data will be in the body (may be given multiple times)")
+	flag.Var(&customHeaders, "header", "Custom header for HTTP requests (may be given multiple times)")
 	flag.Var(&getAddrs, "get", "HTTP address to make a GET request to. '%s' will be printf replaced with data (may be given multiple times)")
 	flag.Var(&nsqdTCPAddrs, "nsqd-tcp-address", "nsqd TCP address (may be given multiple times)")
 	flag.Var(&lookupdHTTPAddrs, "lookupd-http-address", "lookupd HTTP address (may be given multiple times)")
@@ -180,6 +183,14 @@ func main() {
 		return
 	}
 
+	if len(customHeaders) > 0 {
+		var err error
+		validCustomHeaders, err = parseCustomHeaders(customHeaders)
+		if err != nil {
+			log.Fatal("--header value format should be 'key=value'")
+		}
+	}
+
 	if *topic == "" || *channel == "" {
 		log.Fatal("--topic and --channel are required")
 	}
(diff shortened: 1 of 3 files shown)


apps/nsq_to_http/nsq_to_http.go
```diff
@@ -49,14 +50,17 @@ var (
...
+	postAddrs          = app.StringArray{}
+	customHeaders      = app.StringArray{}
+	nsqdTCPAddrs       = app.StringArray{}
+	lookupdHTTPAddrs   = app.StringArray{}
+	validCustomHeaders map[string]string
 )
 
 func init() {
 	flag.Var(&postAddrs, "post", "HTTP address to make a POST request to.  data will be in the body (may be given multiple times)")
+	flag.Var(&customHeaders, "custom-header", "Custom header for HTTP requests (may be given multiple times)")
```
ploxiln: I suggest making the flag shorter, just `header` (it would also match `curl`'s `--header` flag)

mreiferson:
Thanks for the contribution @alwindoss, this is coming out nice so far — looks like @ploxiln's got the code review 😁

apps/nsq_to_http/nsq_to_http.go
alwindoss: Agreed, this would enable people who are comfortable with curl to use this option

alwindoss: Fixed

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -289,3 +301,19 @@ func main() {
 		}
 	}
 }
+
+func parseCustomHeaders(strs []string) (map[string]string, error) {
+	parsedHeaders := make(map[string]string)
+	for _, s := range strs {
+		sp := strings.Split(s, "=")
```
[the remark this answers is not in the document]
alwindoss: Fixed this

alwindoss:
Added `test` and `build/` to `.gitignore`

.gitignore
```diff
@@ -1,6 +1,8 @@
 .godeps
 vendor
 ./build
+build/
```
ploxiln: the line right above this should already work to ignore the build folder

ploxiln: (just the build folder at the very top of the repo directory tree)

.gitignore
```diff
@@ -1,6 +1,8 @@
...
+test
```
ploxiln: this will also ignore `internal/test/` which has checked-in source code we don't want ignored

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -289,3 +301,21 @@ func main() {
...
+	parsedHeaders := make(map[string]string)
+	for _, s := range strs {
+		sp := strings.SplitN(s, ":", 2)
+		if len(sp) != 2 {
+			return nil, errors.New("Not meeting the criteria")
+		}
+		key := strings.TrimSpace(sp[0])
+		val := strings.TrimSpace(sp[1])
+		if key == "" || val == "" {
+			return nil, errors.New("Not meeting the criteria")
```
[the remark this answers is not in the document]
alwindoss: Moved the `log.Printf("Specified custom headers: %v", validCustomHeaders)` to a move appropriate place i.e., after the error check. Also, cleaned up the other logs

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -180,6 +183,14 @@ func main() {
 		return
 	}
 
+	if len(customHeaders) > 0 {
+		validCustomHeaders, err := parseCustomHeaders(customHeaders)
+		if err != nil {
+			log.Fatal("--header value format should be 'key=value'")
+		}
+		log.Printf("Specified custom headers: %v", validCustomHeaders)
```
mreiferson: Can we drop this log line?

alwindoss: Yes it does seem redundant. But technically I need help here. For some reason I keep getting an error that I have an unused variable called `validCustomHeaders`
I could use a lil help here

mreiferson: Ahhh, an annoying quirk of Go known as "variable shadowing". What you want on line 187 is this:

```go
var err error
validCustomHeaders, err = parseCustomHeaders(customHeaders)
...
```

(note the lack of `:=`)

alwindoss: I know Go does not allow you to have unused variables but "variable shadowing" I am hearing for the first time. I have fixed it.

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -180,6 +183,14 @@ func main() {
 		return
 	}
 
+	if len(customHeaders) > 0 {
+		var err error
+		validCustomHeaders, err = parseCustomHeaders(customHeaders)
+		if err != nil {
+			log.Fatal("--header value format should be 'key=value'")
```
ploxiln: you could include the err string in the fatal message

ploxiln:
Looks good to me. Can you squash these commits down to a single commit?

PR #1076, opened 2018-08-23
*: use net/http constants for status codes

chainhelen:
Use httpstatuscode in offical http pacakge instead of hardcode


apps/nsq_to_http/nsq_to_http.go
```diff
@@ -130,7 +130,7 @@ func (p *PostPublisher) Publish(addr string, msg []byte) error {
 	io.Copy(ioutil.Discard, resp.Body)
 	resp.Body.Close()
 
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
```
ploxiln: 👎 on this idea

The http status codes are much more recognizable as small numbers, and less familiar as long names - who knows what it means for a StatusCode to be greater than StatusMultipleChoices?

I think this takes the generally good idea of separately defining "magic constants" too far. In this case the numbers _are_ the generally recognized names.

chainhelen: @ploxiln In net/http/[status.go](https://golang.org/src/net/http/status.go), the numbers are the generally recognized names.  
It is clear , so I think should  reuse this and don't repeat meaningless things.

image

jehiah: @chainhelen I think the point @ploxiln is making is that **this** code block is about the numeric ranges (i.e. 1xx + 4xx, 5xx) not a specific singular status code (`StatusOK`). As such this should use the numeric values not a named alias for a specific numeric value.

chainhelen:
@mreiferson  Thx for your reply.
I reset  the code that used for the numeric ranges. And rebase the two commit into one.  

(I don't have a greater idea to hanld the numeric ranges so I reset those part)
I can close this pr if you disagree with those changes . Thx.

heat:
:+1:

apps/nsq_to_http/nsq_to_http.go
```diff
@@ -39,7 +39,7 @@ var (
 
 	topic       = flag.String("topic", "", "nsq topic")
 	channel     = flag.String("channel", "nsq_to_http", "nsq channel")
-	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
+	maxInFlight = flag.Int("max-in-flight", http.StatusOK, "max number of messages to allow in flight")
```
judwhite: 😅

PR #1092, opened 2018-10-19
Bump Travis versions

HaraldNordgren:


ploxiln:
We've historically trimmed the list to just the last 3 minor (major?) versions, to reduce test time. But, I like supporting slightly older versions of Go, when it's easy to. I think 4 versions is OK. I also think it would be OK to drop 1.9 instead of 1.8, and assume that if 1.8 works and 1.10 works then 1.9 is probably fine.

mreiferson:
Yea, perhaps keeping the "minimum" Go release as an anchor plus the most recent three releases seems like a good approach going forward.

PR #1098, opened 2018-11-08
add Bytedance

andyxning:
This PR adds [Bytedance](https://bytedance.com/) to the `In Production` section.

We largely use NSQ in our production environment to support more and more important services. The QPS from one of our largest clusters is about 300 million at the peak time. :)


ploxiln:
👍

andyxning:
@ploxiln Thanks~

PR #1109, opened 2018-11-30
nsqd: fix topic.messageCount on error

SwanSpouse:
As the topic says, topic.messageCount may be less than actually when using topic.PutMessages to write message to backend occurs an error.

### DIFF
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -194,9 +194,10 @@ func (t *Topic) PutMessages(msgs []*Message) error {
 	if atomic.LoadInt32(&t.exitFlag) == 1 {
 		return errors.New("exiting")
 	}
-	for _, m := range msgs {
+	for i, m := range msgs {
 		err := t.put(m)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


nsqd/topic.go
```diff
@@ -199,8 +199,8 @@ func (t *Topic) PutMessages(msgs []*Message) error {
 		if err != nil {
 			return err
 		}
+		atomic.AddUint64(&t.messageCount, 1)
```
mreiferson: instead of single increments, let's maintain a local counter and then perform a single atomic add below on line 203?

mreiferson:
@SwanSpouse thanks for this PR, just a minor comment!

ploxiln:
Yup, good find!

nsqd/topic.go
mreiferson: ahh right, I completely missed the fact that it was returning, too. thx 🙏

SwanSpouse: I have modified my codes according to @ploxiln 's advice. Please take a look 🙂

PR #1110, opened 2018-12-05
nsq_to_file: support for separate working dir for in-progress files

mccutchen:
### The problem

There are a couple of edge cases we currently need to guard against when shipping the archive files generated by `nsq_to_file` to remote storage (e.g. S3).

 1. **Skipping files for the current hour to which nsq_to_file is actively writing incoming messages.**  This requires our sync scripts to know how nsq_to_file will be bucketing files by time, so they can manually exclude the current bucket.
 2. **Skipping corrupted gzip files that can arise from the combination of using `nsq_to_file -gzip` and an unclean exit (e.g. `kill -9`).**  We don't currently have a good workaround for this issue, which causes issues in downstream systems and requires manual intervention to remediate.

### Proposed change

If `nsq_to_file` used a separate "working dir" for the files to which it is actively writing and only moved files into the final output dir once they were cleanly finished/closed, the problems above would be much easier to address:

 1. We no longer need special logic to skip "active" files, because files that appear in the output dir are guaranteed to be finished and ready to upload
 2. While corrupted gzip files are still possible, they will never show up in the output dir.  A separate process could watch the working dir for orphaned files, allowing for alerts, simpler manual remediation, and even automated recovery.

### Implementation details

If `-work-dir` is given, `nsq_to_file` will create new files under that directory and only move them to `-output-dir` upon close, ensuring that only fully finished and synced files appear in the output dir.

By default, the work dir will match the output dir in order to preserve existing behavior out of the box.

### DIFF
--- a/apps/nsq_to_file/file_logger.go
+++ b/apps/nsq_to_file/file_logger.go
@@ -35,9 +35,9 @@ type FileLogger struct {
 }
 
 func NewFileLogger(gzipEnabled bool, compressionLevel int, filenameFormat, topic string) (*FileLogger, error) {
-	if gzipEnabled || *rotateSize > 0 || *rotateInterval > 0 {
+	if gzipEnabled || *rotateSize > 0 || *rotateInterval > 0 || *workDir != *outputDir {
 		if strings.Index(filenameFormat, "<REV>") == -1 {
-			return nil, errors.New("missing <REV> in --filename-format when gzip or rotation enabled")
+			return nil, errors.New("missing <REV> in --filename-format when gzip, rotation, or work dir enabled")
 		}
 	} else {
 		// remove <REV> as we don't need it
@@ -164,6 +164,45 @@ func (f *FileLogger) Close() {
 			f.gzipWriter.Close()
 		}
 		f.out.Close()
+
+		// Move file from work dir to output dir if necessary, taking care not
+		// to overwrite existing files
+		if *workDir != *outputDir {
+			src := f.out.Name()
+			dst := makeOutputPath(src, *workDir, *outputDir)
+
+			// Optimistic rename
+			log.Printf("INFO: moving finished file %s to %s", src, dst)
+			err := atomicRename(src, dst)
+			if err == nil {
+				return
+			} else if !os.IsExist(err) {
+				log.Fatalf("ERROR: unable to move file from %s to %s: %s", src, dst, err)
+				return
+			}
+
+			// Optimistic rename failed, so we need to generate a new
+			// destination file name by bumping the revision number.
+			_, filenameTmpl := filepath.Split(f.lastFilename)
+			dstDir, _ := filepath.Split(dst)
+			dstTmpl := filepath.Join(dstDir, filenameTmpl)
+
+			for i := f.rev + 1; ; i++ {
+				log.Printf("INFO: destination file already exists: %s", dst)
+				dst := strings.Replace(dstTmpl, "<REV>", fmt.Sprintf("-%06d", i), -1)
+				err := atomicRename(src, dst)
+				if err != nil {
+					if os.IsExist(err) {
+						continue // next rev
+					}
+					log.Fatalf("ERROR: unable to rename file from %s to %s: %s", src, dst, err)
+					return
+				}
+				log.Printf("INFO: renamed finished file %s to %s to avoid overwrite", src, dst)
+				break
+			}
+		}
+
 		f.out = nil
 	}
 }
@@ -227,14 +266,8 @@ func (f *FileLogger) updateFile() {
 	f.lastFilename = filename
 	f.lastOpenTime = time.Now()
 
-	fullPath := path.Join(*outputDir, filename)
-	dir, _ := filepath.Split(fullPath)
-	if dir != "" {
-		err := os.MkdirAll(dir, 0770)
-		if err != nil {
-			log.Fatalf("ERROR: %s Unable to create %s", err, dir)
-		}
-	}
+	fullPath := path.Join(*workDir, filename)
+	makeOutputDir(fullPath)
 
 	f.Close()
 
@@ -242,6 +275,24 @@ func (f *FileLogger) updateFile() {
 	var fi os.FileInfo
 	for ; ; f.rev++ {
 		absFilename := strings.Replace(fullPath, "<REV>", fmt.Sprintf("-%06d", f.rev), -1)
+
+		// If we're using a working directory for in-progress files,
+		// proactively check for duplicate file names in the output dir to
+		// prevent conflicts on rename in the normal case
+		if *workDir != *outputDir {
+			outputFileName := makeOutputPath(absFilename, *workDir, *outputDir)
+			makeOutputDir(outputFileName)
+
+			_, err := os.Stat(outputFileName)
+			if err == nil {
+				log.Printf("INFO: output file already exists: %s", outputFileName)
+				continue // next rev
+			} else if !os.IsNotExist(err) {
+				log.Fatalf("ERROR: unable to stat output file %s: %s", outputFileName, err)
+				return
+			}
+		}
+
 		openFlag := os.O_WRONLY | os.O_CREATE
 		if f.gzipEnabled {
 			openFlag |= os.O_EXCL
@@ -251,7 +302,7 @@ func (f *FileLogger) updateFile() {
 		f.out, err = os.OpenFile(absFilename, openFlag, 0666)
 		if err != nil {
 			if os.IsExist(err) {
-				log.Printf("INFO: file already exists: %s", absFilename)
+				log.Printf("INFO: working file already exists: %s", absFilename)
 				continue
 			}
 			log.Fatalf("ERROR: %s Unable to open %s", err, absFilename)
@@ -277,4 +328,36 @@ func (f *FileLogger) updateFile() {
 	} else {
 		f.writer = f
 	}
+}
+
+// makeOutputFilename translates a file path from work dir to output dir
+func makeOutputPath(workPath string, workDir string, outputDir string) string {
+	if workDir == outputDir {
+		return workPath
+	}
+	return filepath.Join(outputDir, strings.TrimPrefix(workPath, workDir))
+}
+
+func makeOutputDir(path string) {
+	dir, _ := filepath.Split(path)
+	if dir != "" {
+		err := os.MkdirAll(dir, 0770)
+		if err != nil {
+			log.Fatalf("ERROR: %s Unable to create %s", err, dir)
+		}
+	}
+}
+
+func atomicRename(src, dst string) error {
+	err := os.Link(src, dst)
+	if err != nil {
+		return err
+	}
+
+	err = os.Remove(src)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/apps/nsq_to_file/nsq_to_file.go
+++ b/apps/nsq_to_file/nsq_to_file.go
@@ -23,6 +23,7 @@ var (
 	maxInFlight = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 
 	outputDir      = flag.String("output-dir", "/tmp", "directory to write output files to")
+	workDir        = flag.String("work-dir", "", "directory for in-progress files before moving to output-dir")
 	datetimeFormat = flag.String("datetime-format", "%Y-%m-%d_%H", "strftime compatible format for <DATETIME> in filename format")
 	filenameFormat = flag.String("filename-format", "<TOPIC>.<HOST><REV>.<DATETIME>.log", "output filename format (<TOPIC>, <HOST>, <PID>, <DATETIME>, <REV> are replaced. <REV> is increased when file already exists)")
 	hostIdentifier = flag.String("host-identifier", "", "value to output in log filename in place of hostname. <SHORT_HOST> and <HOSTNAME> are valid replacement tokens")
@@ -34,7 +35,7 @@ var (
 
 	rotateSize     = flag.Int64("rotate-size", 0, "rotate the file when it grows bigger than `rotate-size` bytes")
 	rotateInterval = flag.Duration("rotate-interval", 0*time.Second, "rotate the file every duration")
-	syncInterval = flag.Duration("sync-interval", 30*time.Second, "sync file to disk every duration")
+	syncInterval   = flag.Duration("sync-interval", 30*time.Second, "sync file to disk every duration")
 
 	httpConnectTimeout = flag.Duration("http-client-connect-timeout", 2*time.Second, "timeout for HTTP connect")
 	httpRequestTimeout = flag.Duration("http-client-request-timeout", 5*time.Second, "timeout for HTTP request")
@@ -100,6 +101,10 @@ func main() {
 
 	if len(topics) == 0 && len(lookupdHTTPAddrs) == 0 {
 		log.Fatal("--lookupd-http-address must be specified when no --topic specified")
+	}
+
+	if *workDir == "" {
+		*workDir = *outputDir
 	}
 
 	cfg.UserAgent = fmt.Sprintf("nsq_to_file/%s go-nsq/%s", version.Binary, nsq.VERSION)


jehiah:
👍 in general (haven't looked at the code yet of course). I've run into these same pain points to varying degrees.

A separate working directory would make it easy to monitor for any old files in the working directory and manually recover them (remove corrupt gzip blocks with data that will end up in other files)

mccutchen:
> A separate working directory would make it easy to monitor for any old files in the working directory and manually recover them (remove corrupt gzip blocks with data that will end up in other files)

Yep, that's exactly what we're thinking!

apps/nsq_to_file/file_logger.go
```diff
@@ -164,6 +164,19 @@ func (f *FileLogger) Close() {
 			f.gzipWriter.Close()
 		}
 		f.out.Close()
+
+		// Move file from work dir to output dir if necessary
+		if *workDir != *outputDir {
+			src := f.out.Name()
+			dst := strings.Replace(src, *workDir, *outputDir, 1)
```
jehiah: nit, if your working directory is `a` this replace won't be so pretty. Can we use `filepath.Split` on `f.out.Name` here?

mccutchen: Ah, very good point!

That was actually the first approach I took here, but I found that it breaks down for the case where `-filename-format` itself contains a directory, like

```
-filename-format="<TOPIC>/<TOPIC>.<DATETIME>+0000.<HOST><REV>.log"
```

In that case, a naive use of `filepath.Split()` drops the topic directory from the final output destination path.

Would an approach along these lines be good enough?

```go
dst := filepath.Join(*outputDir, strings.TrimPrefix(src, *workDir))
```

mccutchen: (Updated w/ that new approach)

jehiah: 👍 good catch that filename-format can have directories; it's not "just" a filename. `TrimPrefix` sounds like it will work.

mccutchen:
Ahhhhh dang I just realized this change is unsafe as-is and should not be merged!  (I'd dismiss the approval if I could.)

If an `nsq_to_file` process is cleanly restarted, it will move its current file from the work dir to the output dir and then re-open a new file in the workdir with the exact same name because the logic for bumping `<REV>` won't see the finished files in the output dir.  Then, once the time bucket rolls over, the new working file will overwrite the existing file with the same name.

I'm not sure if that explanation makes sense as written, but here's a sort of illustration:

```bash
/tmp/nsqout
└── foo
    ├── foo.2018-12-05_22-41+0000.rig-000000.log.gz
    ├── foo.2018-12-05_22-42+0000.rig-000000.log.gz
    ├── foo.2018-12-05_22-43+0000.rig-000000.log.gz
    └── foo.2018-12-05_22-44+0000.rig-000000.log.gz  # <-- moved to output dir on clean restart
/tmp/nsqwork
└── foo
    └── foo.2018-12-05_22-44+0000.rig-000000.log.gz  # <-- new working file w/ same name, will overwrite file above
```

Note: The current minute is included in these filenames to speed up the testing process; full command line is:

```
~/nsq_to_file -topic=foo -gzip -output-dir=/tmp/nsqout -work-dir=/tmp/nsqwork -filename-format="<TOPIC>/<TOPIC>.<DATETIME>+0000.<HOST><REV>.log" -nsqd-tcp-address=:4150 -datetime-format="%Y-%m-%d_%H-%M"
```

mreiferson:
Good catch, feel like we should perform a similar check at rename time as we do here https://github.com/nsqio/nsq/pull/1110/files#diff-375625c60c85a1884dbff6cf36306703R260 (and cycle through revision numbers?

ploxiln:
Great idea, and great catch.

I don't think we can use the exact same filename conflict logic, since this bit is not opening a new file to write to, it's moving to a new name:

```go
 		if f.gzipEnabled {
 			openFlag |= os.O_EXCL
 		} else {
 			openFlag |= os.O_APPEND
 		}
 		f.out, err = os.OpenFile(absFilename, openFlag, 0666)
 		if err != nil {
if os.IsExist(err) {
```

I think there's no `O_EXCL` like flag for `Rename()` (in Go or Posix or Linux). So it'll be slightly different, we'll have to just stat the dst filepath.

ploxiln:
ah, very cool 👍

jehiah:
I think i dislike the "rename on move" approach (or at least relying on that in entirety). I think we should pick initial name based on unused names between both directories. (even though it introduces some race condition issues if you were to run more than one nsq_to_file against the same directory, it should be safe outside of that). 

I like the property that a filename is "unique" and constant once it's opened. That means that if you kill -9 nsq_to_file you still have a uniquely named file in the working directory to recover and don't have to do extra work to figure out if it was a duplicate file or not where that filename might already appear elsewhere.

In practice I think this means a`os.Stat` on output-dir before the `os.Create` against work-dir when iterating to find an available filename.

mreiferson:
After thinking this through a bit, I agree that optimistically checking both dirs is a useful addition.

However, I'd like `nsq_to_file` to be confidently race free and operator-friendly in avoiding data loss, so I think it's useful to preserve the behavior @mccutchen is introducing to perform a conflict-free rename. In other words, it will ensure that `nsq_to_file` will _never_ overwrite a file in the output dir.

The question is then, what should `nsq_to_file`'s behavior be if it does detect a conflict? @jehiah's suggesting that it shouldn't automatically rename. I'm fine with keeping things simple, which I suppose would be to "do nothing" by logging the conflict and leaving the file in the working dir?

WDYT @mccutchen @jehiah?

jehiah:
If we check both directories to pick a unique filename up front in the normal case, i'm ok with the possibility of a rename on move to guarantee a path to resolve conflicts at that time.

mccutchen:
👍 cool, I'll take a pass at this

mccutchen:
Okay, with the latest commits, I think I've implemented the approach described above. Please take another look!

There's a bit of awkwardness/duplication in the code (e.g. safely munging file paths, duplicated log messages, not-quite-the-same overlap in functionality when iterating through revs) so all suggestions for improvements are welcome.

I'm gonna bang on this locally a bit more to try to ensure I'm not missing any edge cases.

mccutchen:
Meh, sorry for the noise, but I have spotted a bug when choosing a new file name when there's a conflict in the output dir.  Note that the minute changed from `17` to `18` in the file name listed below:

```
2018/12/07 22:18:00 INFO: renamed finished file /tmp/nsqwork/foo/foo.2018-12-07_22-17+0000.rig-000001.log.gz to /tmp/nsqout/foo/foo.2018-12-07_22-18+0000.rig-000001.log.gz to avoid overwrite
```

mccutchen:
Okay, I think this change is functionally complete and working as expected, though the code itself is not very pretty.  Please take another look when y'all get a chance @jehiah @mreiferson @ploxiln.

mccutchen:
Also, to give myself a bit more confidence in the new behavior here, I spent the weekend over-engineering this jepsen-lite test program to exercise these changes:
https://gist.github.com/mccutchen/174bf5681264871644dfb5b8ec079cd9

An example run:

<details>

```
$ go run nsq_to_file_jepsen.go 
2018/12/11 16:33:45 =============================================================
2018/12/11 16:33:45 setup
2018/12/11 16:33:45 =============================================================
2018/12/11 16:33:45 removing working dir /var/folders/4m/qzsqvqqj2p5gcn_ygxdpv8q00000gn/T/nsq-to-file-test157817500 (if it exists)
2018/12/11 16:33:45 creating working dir /var/folders/4m/qzsqvqqj2p5gcn_ygxdpv8q00000gn/T/nsq-to-file-test157817500
2018/12/11 16:33:45 nsqd and nsq_to_file logs combined in /var/folders/4m/qzsqvqqj2p5gcn_ygxdpv8q00000gn/T/nsq-to-file-test157817500/logs.txt
2018/12/11 16:33:45 running nsqd
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:46 running nsq_to_file for topic "foo", restarting every 10s
2018/12/11 16:33:47 =============================================================
2018/12/11 16:33:47 generating messages
2018/12/11 16:33:47 =============================================================
2018/12/11 16:33:47 publishing 100000 messages over 3m0s
2018/12/11 16:33:47 nsq_to_file stats: 1 messages; 0 channel depth, 1 in flight
2018/12/11 16:33:52 nsq_to_file stats: 3640 messages; 0 channel depth, 257 in flight
2018/12/11 16:33:57 nsq_to_file stats: 7296 messages; 0 channel depth, 645 in flight
2018/12/11 16:34:02 nsq_to_file stats: 11051 messages; 0 channel depth, 1663 in flight
2018/12/11 16:34:07 nsq_to_file stats: 14681 messages; 0 channel depth, 528 in flight
2018/12/11 16:34:12 nsq_to_file stats: 18376 messages; 0 channel depth, 856 in flight
2018/12/11 16:34:17 nsq_to_file stats: 22006 messages; 0 channel depth, 429 in flight
2018/12/11 16:34:22 nsq_to_file stats: 25661 messages; 0 channel depth, 705 in flight
2018/12/11 16:34:27 nsq_to_file stats: 29293 messages; 0 channel depth, 303 in flight
2018/12/11 16:34:32 nsq_to_file stats: 32960 messages; 0 channel depth, 603 in flight
2018/12/11 16:34:37 nsq_to_file stats: 36597 messages; 0 channel depth, 182 in flight
2018/12/11 16:34:42 nsq_to_file stats: 40241 messages; 0 channel depth, 438 in flight
2018/12/11 16:34:47 nsq_to_file stats: 43988 messages; 0 channel depth, 84 in flight
2018/12/11 16:34:52 nsq_to_file stats: 47663 messages; 0 channel depth, 384 in flight
2018/12/11 16:34:57 nsq_to_file stats: 51359 messages; 0 channel depth, 98 in flight
2018/12/11 16:35:02 nsq_to_file stats: 55059 messages; 0 channel depth, 1607 in flight
2018/12/11 16:35:07 nsq_to_file stats: 58707 messages; 0 channel depth, 339 in flight
2018/12/11 16:35:12 nsq_to_file stats: 62149 messages; 0 channel depth, 1489 in flight
2018/12/11 16:35:17 nsq_to_file stats: 65710 messages; 0 channel depth, 1511 in flight
2018/12/11 16:35:22 nsq_to_file stats: 69226 messages; 0 channel depth, 1376 in flight
2018/12/11 16:35:27 nsq_to_file stats: 72891 messages; 0 channel depth, 325 in flight
2018/12/11 16:35:32 nsq_to_file stats: 76565 messages; 0 channel depth, 1500 in flight
2018/12/11 16:35:37 nsq_to_file stats: 80217 messages; 0 channel depth, 268 in flight
2018/12/11 16:35:42 nsq_to_file stats: 83899 messages; 0 channel depth, 1342 in flight
2018/12/11 16:35:47 nsq_to_file stats: 87559 messages; 0 channel depth, 685 in flight
2018/12/11 16:35:52 nsq_to_file stats: 91165 messages; 0 channel depth, 1395 in flight
2018/12/11 16:35:57 nsq_to_file stats: 94816 messages; 0 channel depth, 1509 in flight
2018/12/11 16:36:02 nsq_to_file stats: 98285 messages; 0 channel depth, 1601 in flight
2018/12/11 16:36:07 nsq_to_file stats: 100000 messages; 0 channel depth, 1293 in flight
2018/12/11 16:36:12 nsq_to_file stats: 100000 messages; 0 channel depth, 0 in flight
2018/12/11 16:36:12 nsqd fully drained
2018/12/11 16:36:12 stopping nsqd
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 stopping nsq_to_file
2018/12/11 16:36:12 =============================================================
2018/12/11 16:36:12 report
2018/12/11 16:36:12 =============================================================
2018/12/11 16:36:12 success: found exactly 100000 messages in 170 output files
2018/12/11 16:36:12 output dir: /var/folders/4m/qzsqvqqj2p5gcn_ygxdpv8q00000gn/T/nsq-to-file-test157817500/output
```

</details>

apps/nsq_to_file/file_logger.go
```diff
@@ -227,21 +266,33 @@ func (f *FileLogger) updateFile() {
...
+
+		// If we're using a working directory for in-progress files,
+		// proactively check for duplicate file names in the output dir to
+		// prevent conflicts on rename in the normal case
+		if *workDir != *outputDir {
+			outputFileName := makeOutputPath(absFilename, *workDir, *outputDir)
+			makeOutputDir(outputFileName)
+
+			_, err := os.Stat(outputFileName)
+			if err == nil || os.IsExist(err) {
```
ploxiln: I would guess that you should never get an error here (including an Exist error). But this seems very safe, and clear.

apps/nsq_to_file/file_logger.go
```diff
@@ -278,3 +329,35 @@ func (f *FileLogger) updateFile() {
...
+	dir, _ := filepath.Split(path)
+	if dir != "" {
+		err := os.MkdirAll(dir, 0770)
+		if err != nil {
+			log.Fatalf("ERROR: %s Unable to create %s", err, dir)
+		}
+	}
+}
+
+func atomicRename(src, dst string) error {
```
ploxiln: I might have called this `exclusiveRename()` because the file does exist in two places for a moment, and because it emulates the `open(..., O_EXCL)`, but this is also fine.

apps/nsq_to_file/file_logger.go
mccutchen: Yeah, during normal operation this conditional should never be true.

The `os.IsExist` case covers an unlikely but possible scenario (I think!) where some entirely other process adds a file to the final output dir that has the same name as an in-progress file in the working dir, in a sequence of events along these lines:
 - Two (or more) nsq_to_file processes are sharing the same working and output dirs
 - One process is `kill -9`'d, orphaning a file in the working dir
 - A separate cleanup process scans the working dir for orphaned files and moves the file into a tmp dir for cleanup/processing
 - The killed nsq_to_file process starts back up and picks the same file name as the orphaned file (because it's no longer present in either work dir or output dir)
 - The cleanup process finishes processing and naively (or _buggily_) sticks the orphaned file into the output dir under the same name
- The restarted nsq_to_file process now has to safely pick a new filename at the point where it's trying to move the file from working dir to output dir

I think the sequence above is quite unlikely, but, as you said, I think guarding against it isn't too onerous!

ploxiln: I just mean, if the file exists, I suspect the stat will never return an error, it will be the `err == nil` case.

mccutchen:
Hahaha oh, yeah, that sounds about right. 🤦‍♂️

mccutchen:
Squashed and rebased, but the latter step seems to have angered the code coverage tests.  Clicking through to [the failing build](https://coveralls.io/builds/20873354), I'm not sure the failure has anything to do with these changes.

mreiferson:
it can be noisy, nbd

thanks!

yellottyellott:
We're finally making moves to use this feature. @mreiferson could we cut a new nsq release to get this included sometime soon? I think the latest release is 1.1.0 on Aug 19, 2018

mreiferson:
You’re saying you want me to actually touch a computer?

yellottyellott:
Will send brisket in exchange. Or the frozen turkeys @mccutchen loves so much.

Issue #880, opened 2017-04-08
nsq_to_file: log less (or support a "quiet mode")


dominicbarnes:
Currently, `nsq_to_tail` is logging each time it flushes to disk, which ends up being a lot especially when the nsq topic has a non-trivial volume of data. It would be really great if there were ways to tell `nsq_to_file` to log less frequently. (like, errors only)

I know the internals are just using the golang `log` package, so adding levels like "debug" and such aren't available as a feature there, but it would be great if something similar could be implemented.

jehiah:
@dominicbarnes can you give some context as to your nsq_to_file configuration and volume? How often is the `syncing xxx` logging happening in practice for you?

For some context, I log many high volume datastreams, but i run my nsq_to_file with `--max-in-flight=50000` to flush my gzip stream to disk less often. I end up with this log frequency as a result:

```
2017/04/08 00:45:44 syncing 42405 records to disk
2017/04/08 00:46:01 syncing 51935 records to disk
```

dominicbarnes:
I sampled a single second on a single machine, and I saw 50+ logs/second on average. Each had <100 records being flushed to disk. My relevant configuration looks like:

```
-max-in-flight 300
-rotate-interval 5m
-rotate-size 10485760 # 10MB
```

What config in general impacts how often flushing to disk happens?

jehiah:
There are two aspects. One is `-max-in-flight 300` combined with the number of connections. It's a little bit of a simplification, but when a single connection reaches it's limit (max-in-flight / number of connections) messages get flushed to avoid issues in situation where the source hosts are imbalanced. 

The second is that nsq_to_file will always flush at least once every 30 seconds.

mreiferson:
Looks like #892 will land soon and we can apply the same to the other utility applications.

mreiferson:
see #1117

PR #1117, opened 2019-01-03
nsq_to_file: cleanup

mreiferson:
I've long been frustrated with `nsq_to_file`'s codebase, and didn't want to schlep it off onto @mccutchen in #1110, so here goes.


mreiferson:
(individual commits are probably easier to review than the combined diff)

mreiferson:
Fixes #880 in 8b738efad1dadf52e774374416680e294ab150de

mreiferson:
I think I'm done here, still passes @mccutchen's nice ["jepsen lite" test program](https://github.com/nsqio/nsq/pull/1110#issuecomment-446419249).

apps/nsq_to_file/topic_discoverer.go
```diff
@@ -10,89 +9,92 @@ import (
...
+	t.updateTopics(t.opts.Topics)
 forloop:
 	for {
 		select {
 		case <-ticker:
-			newTopics, err := t.ci.GetLookupdTopics(addrs)
+			newTopics, err := t.ci.GetLookupdTopics(t.opts.NSQLookupdHTTPAddrs)
 			if err != nil {
-				log.Printf("ERROR: could not retrieve topic list: %s", err)
+				t.opts.logf(lg.ERROR, "could not retrieve topic list: %s", err)
```
ploxiln: if logf was a member or method of `TopicDiscoverer` and `FileLogger` that would make all the log lines a bit more concise (removing the need for `.opts` in each line)

ploxiln:
All looks pretty good to me.

I guess you probably plan to squash all commits down before merge, but otherwise I see some runs of commits that could be squashed together, leaving 5 or so commits in the sequence that could be merged.

apps/nsq_to_file/topic_discoverer.go
mreiferson: Yea, I'm definitely abusing `opts` being passed around everywhere... I'll fix.

mreiferson:
> I guess you probably plan to squash all commits down before merge

Hah, I wasn't. So I'm all ears to how you think this should be squashed.

mreiferson:
@jehiah you might be particularly interested in deb7c330894d87803da74ac8fc2be4f5d8a8d45b

apps/nsq_to_file/topic_discoverer.go
mreiferson: fixed

ploxiln:
If you want my 2c on how to squash, sure ;)

Leave as-is:
  * nsq_to_file: move to go-options; pass Options struct around
  * nsq_to_file: cleanup startup

FileLogger refactor commits could be squashed:

  * nsq_to_file: remove redundant closing var
  * nsq_to_file: cleanup variable names/cfl responsibilities
  * nsq_to_file: drop ConsumerFileLogger; refactor FileLogger

FileLogger cleanup/rename commits could be squashed:

  * nsq_to_file: un-indent FileLogger.Close()
  * nsq_to_file: s/atomicRename/exclusiveRename 
  * nsq_to_file: drop superfluous makeOutputPath 
  * nsq_to_file: cleanup FileLogger var names 
  * nsq_to_file: dead return; whitespace cleanup 
  * nsq_to_file: s/needsFileRotate/needsRotation
  * nsq_to_file: s/makeOutputDir/makeDirFromPath 

Leave as-is:
  * nsq_to_file: don't open multiple GZIP streams

Gzip/Sync commits could be squashed:

  * nsq_to_file: fsync _after_ GZIP close
  * nsq_to_file: check and exit on errors for file-related operations 

Leave as-is:

  * nsq_to_file: refactor confusing Write interface

Finally, logging commits at the end can be squashed

apps/nsq_to_file/file_logger.go
```diff
@@ -151,123 +151,138 @@ func (f *FileLogger) router(r *nsq.Consumer) {
...
 func (f *FileLogger) Sync() error {
-	var err error
 	if f.gzipWriter != nil {
-		f.gzipWriter.Close()
-		err = f.out.Sync()
-		f.gzipWriter, _ = gzip.NewWriterLevel(f, f.compressionLevel)
-		f.writer = f.gzipWriter
-	} else {
-		err = f.out.Sync()
+		err := f.gzipWriter.Flush()
```
jehiah: I think i'm 👎on this switch, but I acknowledge that it is possible to encounter unexpected issues with a stream of indipendent gzip chunks (i.e. equiv to `cat file1.gz file2.gz`).

Without separate gzip streams, recovering from a corrupt file (kill -9 w/ partial data written but not sync'd) becomes problematic. 

I think it comes down to the fact that Flush does not write a gzip footer which has the checksum. Close does.

mreiferson: Unless I’m misunderstanding something, the recoverable section would be equivalent in either case (bound by the configured sync interval)?

You’re suggesting that by writing out _closing_ footers somehow more data would be recoverable? Given gzip’s streamable quality, I assume that a complete block (regardless of a closed stream) would be decompressable?

jehiah: > I assume that a complete block (regardless of a closed stream) would be decompressable?

decompressable yes, error checkable no. The footer contains the final crc for a stream which you don't get from a sync()

https://www.forensicswiki.org/wiki/Gzip#File_footer

### recoverability

With separate streams for each logical sync, I've found it easy to step through the streams of a file (using [`compress/gzip.Reader.Multistream`](https://golang.org/pkg/compress/gzip/#Reader.Multistream) and use the valid streams to recover a corrupt file. When you don't have a stream boundary recovery is less precise because you can't disambiguate between written but not sync'd data and written and sync'd data. That increases the risk of including records that were never ack'd back to nsqd which isn't desirable.

mreiferson: Nice, that feels like a strong argument to revert, thanks.

PR #1120, opened 2019-01-06
nsq_to_file: revert gzip behavior change

ploxiln:
In case we're all swayed by @jehiah's arguments in https://github.com/nsqio/nsq/pull/1117#discussion_r245462312 I opened this easy PR.

I'll squash before merge - I left the mechanical revert separate from additional tweaks needed due to the rest of the refactor, for possibly easier review.

cc @mreiferson


jehiah:
@mreiferson thoughts?

mreiferson:
As a followup, I feel like we should comment the reason why we sync the way we do in that function? WDYT @ploxiln?

ploxiln:
sure, will add in #1123

PR #1123, opened 2019-01-08
nsq_to_file: check for rotate size specifically in rev-incr loop

ploxiln:
instead of calling `needsRotation()` because it is theoretically possible that the time-format could
roll-over in the middle of the `updateFile()` rev-incr loop, so `needsRotation()` would always return `true`, and the loop would never terminate.

cc @mreiferson @jehiah 

I thought of this while playing with @mccutchen's `nsq_to_file_jepsen.go` and tweaking it to work with non-gzip mode as well. That doesn't work because all the concurrent `nsq_to_file` instances using the same filename format at the same time can open the same file as each other in append mode (if not using gzip), so only one successfully moves the file and the others fail due to source file missing, and exit with failure. (This seems like "working as designed" to me.)

While figuring that out, I figured that calling `needsRotation()` inside `updateFile()` is an iffy proposition - `f.filename` and `f.openTime` could cause `needsRotation()` to return true, but that won't update `f.filename` or `f.openTime`. It seemed like duplicating the filesize check was the simplest thing to do here. (And note that `updateFile()` is always called immediately after `needsRotation()`.)

### DIFF
--- a/apps/nsq_to_file/file_logger.go
+++ b/apps/nsq_to_file/file_logger.go
@@ -317,7 +317,7 @@ func (f *FileLogger) updateFile() {
 		}
 
 		openFlag := os.O_WRONLY | os.O_CREATE
-		if f.opts.GZIP {
+		if f.opts.GZIP || f.opts.RotateInterval > 0 {
 			openFlag |= os.O_EXCL
 		} else {
 			openFlag |= os.O_APPEND
--- a/apps/nsq_to_file/file_logger.go
+++ b/apps/nsq_to_file/file_logger.go
@@ -231,6 +231,8 @@ func (f *FileLogger) Write(p []byte) (int, error) {
 func (f *FileLogger) Sync() error {
 	var err error
 	if f.gzipWriter != nil {
+		// finish current gzip stream and start a new one (concatenated)
+		// gzip stream trailer has checksum, and can indicate which messages were ACKed
 		err = f.gzipWriter.Close()
 		if err != nil {
 			return err
--- a/apps/nsq_to_file/file_logger.go
+++ b/apps/nsq_to_file/file_logger.go
@@ -324,7 +324,7 @@ func (f *FileLogger) updateFile() {
 		if err != nil {
 			if os.IsExist(err) {
 				f.logf(lg.WARN, "working file already exists: %s", absFilename)
-				continue
+				continue // next rev
 			}
 			f.logf(lg.FATAL, "unable to open %s: %s", absFilename, err)
 			os.Exit(1)
@@ -337,15 +337,14 @@ func (f *FileLogger) updateFile() {
 			f.logf(lg.FATAL, "unable to stat file %s: %s", f.out.Name(), err)
 		}
 		f.filesize = fi.Size()
-		if f.filesize == 0 {
-			break // ok, new file
-		}
 
-		if f.needsRotation() {
+		if f.opts.RotateSize > 0 && f.filesize > f.opts.RotateSize {
+			f.logf(lg.INFO, "%s currently %d bytes (> %d), rotating...",
+				f.out.Name(), f.filesize, f.opts.RotateSize)
 			continue // next rev
 		}
 
-		break // ok, don't need rotate
+		break // good file
 	}
 
 	if f.opts.GZIP {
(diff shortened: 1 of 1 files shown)


mreiferson:
do we need the rotate interval check copied too?

ploxiln:
The rotate interval should not be hit, because `f.openTime` was updated at the beginning of `updateFile()`, and `updateFile()` should not take a significant amount of time. Put another way, I guess you're asking if we should try to figure out an older openTime for an existing file we just re-opened in append mode?

This is part of the loop is really about corner cases. What if a later REV of the same datestamp file already exists, and we are not gzipping so we can open an existing file with append mode, but then the existing file already meets the rotate size? I guess this is possible if `nsq_to_file` is restarted, and it needs to skip over already-full REVs for the current datestamp.

So back to the rotate-time check: I don't think we can figure out the create time of an existing file on disk in a good way. Posix offers last-metadata-change-time and last-contents-change-time. Maybe if rotating based on time, the best we can do is just always pick a new file instead of an existing file?

mreiferson:
Would some of this be easier if we just always used exclusive mode. What is the argument for preserving the behavior of multiple processes appending to the same file? Is that even safe to do?

ploxiln:
Multiple processes at the same time is not the intended case ... it's for when a service restarts, it can pick up where it left off, no need for REVs (if no gzip and no size or time limit, just "date-file-format")

mreiferson:
I think I'm confused from your comment above:

> That doesn't work because all the concurrent nsq_to_file instances using the same filename format at the same time **can open the same file as each other in append mode** (if not using gzip), so only one successfully moves the file and the others fail due to source file missing, and exit with failure. (This seems like "working as designed" to me.)

It doesn't seem like we want it to be possible for two processes to open the same file. The only way to do that would be to always use `O_EXCL`, which means in the single-process case, you'd generate more files when restarting. I _feel_ like systems built to use the output files of `nsq_to_file` already expect to have to handle multiple revisions, so this doesn't seem like a bad tradeoff. But, this would mean that `<REV>` is required, which is a backwards-incompatible change we cannot make lightly.

Regarding the addition of rotate time checking in this loop, your explanation makes sense. I'm still generally unhappy with the complexity of the code in `FileLogger`, so I'm still looking for ways to simplify and reduce edge cases.

ploxiln:
successfully tested with modified pseudo-jepsen:

```diff
--- nsq_to_file_jepsen.go.orig	2019-01-08 19:33:04.000000000 -0500
+++ nsq_to_file_jepsen.go	2019-01-08 19:28:43.000000000 -0500
@@ -45,6 +45,7 @@
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -97,7 +98,7 @@
 			"-work-dir", filepath.Join(outputRoot, "work"),
 			"-filename-format", "<TOPIC>/<TOPIC>.<DATETIME>+0000.<HOST><REV>.log",
 			"-datetime-format", "%Y-%m-%d_%H-%M",
-			"-gzip",
+			"-rotate-interval", "30s",
 		)
 		cmd.Stdout = logSink
 		cmd.Stderr = logSink
@@ -284,13 +285,18 @@
 		return fmt.Errorf("error opening output file %s: %s", path, err)
 	}
 	defer f.Close()
-	gzf, err := gzip.NewReader(f)
-	if err != nil {
-		return fmt.Errorf("error opening gzip reader: %s: %s", path, err)
+	var rf io.Reader = f
+
+	if strings.HasSuffix(path, ".gz") {
+		gzf, err := gzip.NewReader(f)
+		if err != nil {
+			return fmt.Errorf("error opening gzip reader: %s: %s", path, err)
+		}
+		defer gzf.Close()
+		rf = gzf
 	}
-	defer gzf.Close()
 
-	scanner := bufio.NewScanner(gzf)
+	scanner := bufio.NewScanner(rf)
 	for scanner.Scan() {
 		line := scanner.Bytes()
 		msg := &Msg{}
```

mreiferson:
Interesting, those separate writes stood out to me when I was refactoring, but I suspect that the `append(...)` you've added forces an allocation (which is basically the reason I wrote them separately to begin with).

Ideally we'd use something like vectorio (`writev`) to write multiple slices with one syscall. For a long time there was no support for this in the Go standard library, but it looks like at some point it was officially added as [`net.Buffers`](https://golang.org/pkg/net/#Buffers)?

Want to give it a try?

mreiferson:
heh, of course

PR #1125, opened 2019-01-15
nsqd: add support for customise output-buffer-timeout

andyxning:
This PR adds support for customising server output-buffer-timeout default value. 

We have encountered with a single nsqd instance with more than 40k connections. With a default 250ms output buffer timeout we have a high cpu usage. After customising it with 500ms, the CPU
usage drops about 10%. 

image


apps/nsqd/nsqd.go
```diff
@@ -117,6 +117,7 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Int64("max-rdy-count", opts.MaxRdyCount, "maximum RDY count for a client")
 	flagSet.Int64("max-output-buffer-size", opts.MaxOutputBufferSize, "maximum client configurable size (in bytes) for a client output buffer")
 	flagSet.Duration("max-output-buffer-timeout", opts.MaxOutputBufferTimeout, "maximum client configurable duration of time between flushing to a client")
+	flagSet.Duration("output-buffer-timeout", opts.OutputBufferTimeout, "server configurable duration of time between flushing to a client")
```
mreiferson: let's rename the help text here to:

```default duration of time between flushing data to clients```

mreiferson:
I'm OK with this, thanks for the PR @andyxning!

apps/nsqd/nsqd.go
andyxning: @mreiferson Done. PTAL.

Issue #727, opened 2016-03-16
nsqadmin: JSON warning messages


gjohnson:
How would one make these go away? 

![screen shot 2016-03-16 at 9 46 21 am](https://cloud.githubusercontent.com/assets/67717/13814423/93100c8a-eb5c-11e5-9f83-5859e96d0a1e.png)

mreiferson:
hey @gjohnson I'm ... very confused.

What page are you looking at when you see those messages?

mreiferson:
anything more interesting in the `nsqadmin` logs?

mreiferson:
@gjohnson is this still happening, were you able to get any logs?

mreiferson:
I've seen this from time to time, hopefully next time I'll remember to grab some logs.

ploxiln:
I think this was fixed

PR #1132, opened 2019-01-27
internal/http_api: fixed error messages quoting

FilWisher:
Error messages that get bubbled up are returned from the API wrapped
using a call to fmt.Sprintf(`{"message":"%s"}`, data). Double quotes
appearing inside `data` are unescaped causing incorrectly formatted JSON
(which breaks the UI).

### DIFF
--- a/internal/http_api/api_response.go
+++ b/internal/http_api/api_response.go
@@ -91,7 +91,9 @@ func RespondV1(w http.ResponseWriter, code int, data interface{}) {
 
 	if code != 200 {
 		isJSON = true
-		response = []byte(fmt.Sprintf(`{"message":"%s"}`, data))
+		response, _ = json.Marshal(struct {
+			Message string `json:"message"`
+		}{fmt.Sprintf("%s", data)})
 	}
 
 	if isJSON {


FilWisher:
The API request to `/api/topics` failed but the JSON response is also incorrectly formatted causing a UI error. The offending error message is:

```
{"message":"UPSTREAM_ERROR: Failed to query any nsqlookupd: Get http://127.0.0.1:4160/topics: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x00\x0eE_BAD_PROTOCOL""}
```
You can see the double quotes around `\x00\x00..._BAD_PROTOCOL` aren't escaped correctly.

internal/http_api/api_response.go
```diff
@@ -91,7 +92,7 @@ func RespondV1(w http.ResponseWriter, code int, data interface{}) {
 
 	if code != 200 {
 		isJSON = true
-		response = []byte(fmt.Sprintf(`{"message":"%s"}`, data))
+		response = []byte(fmt.Sprintf(`{"message":%s}`, strconv.Quote(fmt.Sprintf("%s", data))))
```
jehiah: Since we are building a JSON response, we should just update to use `encoding/json` here so there aren't any escaping issues in the future.

```go
response, _ := json.Marshal(struct {
    Message string `json:"message"`
}{fmt.Sprintf("%s", data)})
```

mreiferson:
I wonder if this explains #727 ???

internal/http_api/api_response.go
FilWisher: I agree, this is a cleaner and more complete solution. In the case that `json.Marshal` fails, where should the error be propagated? Or is it safe to assume it won't fail in the case?

FilWisher: I have updated this. I'm not marking it as resolved until I hear back about the question above.

ploxiln:
Thanks for the update. There's just one tiny bit which go is not happy with:

```
internal/http_api/api_response.go:8:2: imported and not used: "strconv"
```

internal/http_api/api_response.go
ploxiln: If we were to check for error from `json.Marshal()`, we would log it, and then for the reply probably:

```go
response = []byte(`{"message":"INTERNAL_ERROR"}`)
```

...but it does seem impossible for this `json.Marshal()` to fail.

mreiferson: agreed, should be impossible, no need to check for errors

FilWisher: OK sounds good :+1:

PR #1133, opened 2019-02-02
nsqd: add flag --min-output-buffer-timeout with default 25ms

ploxiln:
Lower timeouts are more expensive for nsqd, so it seems odd to put
a maximum on the timeout a client can request, but not a minimum.

Also change `--max-output-buffer-timeout` default to 30 seconds,
up from 1 second. Again, it's cheaper/easier for nsqd.
But we may still want to help people avoid unwittingly setting
it close to (or above) the default `--msg-timeout` (60 seconds).

----

I wasn't familiar with `output_buffer_timeout` until #1125 came up. I don't have a real need for these options changes, they just seem to make a bit more sense. But what do you think @mreiferson ?

possibly useful links to refresh memory:
https://github.com/nsqio/nsq/pull/236/commits/043b79acda5fe57056b3cc21b2ef536d5615a2c2
https://nsq.io/clients/tcp_protocol_spec.html#identify

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -117,6 +117,7 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Int64("max-rdy-count", opts.MaxRdyCount, "maximum RDY count for a client")
 	flagSet.Int64("max-output-buffer-size", opts.MaxOutputBufferSize, "maximum client configurable size (in bytes) for a client output buffer")
 	flagSet.Duration("max-output-buffer-timeout", opts.MaxOutputBufferTimeout, "maximum client configurable duration of time between flushing to a client")
+	flagSet.Duration("min-output-buffer-timeout", opts.MinOutputBufferTimeout, "minimum client configurable duration of time between flushing to a client")
 	flagSet.Duration("output-buffer-timeout", opts.OutputBufferTimeout, "default duration of time between flushing data to clients")
 
 	// statsd integration options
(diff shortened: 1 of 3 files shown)


ploxiln:
I'm also trying to figure out how disabled output-buffer-timeout works.

```go
func (c *clientV2) SetOutputBufferTimeout(desiredTimeout int) error {
	...
	switch {
	case desiredTimeout == -1:
		c.OutputBufferTimeout = 0
```

```go
func (p *protocolV2) messagePump(client *clientV2, startedChan chan bool) {
	...
	subEventChan := client.SubEventChan
	identifyEventChan := client.IdentifyEventChan
	outputBufferTicker := time.NewTicker(client.OutputBufferTimeout)
```

https://golang.org/pkg/time/#NewTicker
> The duration d must be greater than zero; if not, NewTicker will panic.

ploxiln:
Ah - `messagePump()` starts before `IDENTIFY` so it always starts with the default ...

ploxiln:
(The reason I was a bit confused about disabling the timeout is because I thought I read somewhere that -1 would disable output buffering, but that's just for OutputBufferSize. So it's odd that there's a --max-output-buffer-timeout but if the client requests -1 for output-buffer-timeout it gets a sort of infinitely long timeout.)

ploxiln:
This makes me think that if either `output_buffer_size` or `output_buffer_timeout` is disabled, the other should be too. I don't think that currently is the case.

ploxiln:
I'm going to let it sit for just a bit, then re-review and manually test a bit, before merge.
(I opened this partly for discussion, and was expecting some direction :)

mreiferson:
Ha, it seemed like the conversation (with yourself) sort of concluded 😜? The changes don't seem controversial, what do you feel like we need to discuss?

ploxiln:
Did some local testing, looked good. 

Side note: pynsq does not currently allow `-1` for `output_buffer_size` or `output_buffer_timeout`

PR #1136, opened 2019-02-20
nsqd: add max_channel_client_connection_count

andyxning:
This PR adds `max_channel_client_connection_count` to nsqd to limit the client connection count per channel.

This is mainly used to limit the channel consumer connections per channel.

### DIFF
--- a/apps/nsqd/nsqd.go
+++ b/apps/nsqd/nsqd.go
@@ -119,6 +119,7 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Duration("max-output-buffer-timeout", opts.MaxOutputBufferTimeout, "maximum client configurable duration of time between flushing to a client")
 	flagSet.Duration("min-output-buffer-timeout", opts.MinOutputBufferTimeout, "minimum client configurable duration of time between flushing to a client")
 	flagSet.Duration("output-buffer-timeout", opts.OutputBufferTimeout, "default duration of time between flushing data to clients")
+	flagSet.Int("max-channel-consumers", opts.MaxChannelConsumers, "maximum channel consumer connection count per nsqd instance (default 0, i.e., unlimited)")
 
 	// statsd integration options
 	flagSet.String("statsd-address", opts.StatsdAddress, "UDP <addr>:<port> of a statsd daemon for pushing stats")
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -389,15 +389,22 @@ func (c *Channel) RequeueMessage(clientID int64, id MessageID, timeout time.Dura
 }
 
 // AddClient adds a client to the Channel's client list
-func (c *Channel) AddClient(clientID int64, client Consumer) {
+func (c *Channel) AddClient(clientID int64, client Consumer) error {
 	c.Lock()
 	defer c.Unlock()
 
 	_, ok := c.clients[clientID]
 	if ok {
-		return
+		return nil
 	}
+
+	maxChannelConsumers := c.ctx.nsqd.getOpts().MaxChannelConsumers
+	if maxChannelConsumers != 0 && len(c.clients) >= maxChannelConsumers {
+		return errors.New("E_TOO_MANY_CHANNEL_CONSUMERS")
+	}
+
 	c.clients[clientID] = client
+	return nil
 }
 
 // RemoveClient removes a client from the Channel's client list
--- a/nsqd/channel_test.go
+++ b/nsqd/channel_test.go
@@ -152,7 +152,8 @@ func TestChannelEmptyConsumer(t *testing.T) {
 	channel := topic.GetChannel("channel")
 	client := newClientV2(0, conn, &context{nsqd})
 	client.SetReadyCount(25)
-	channel.AddClient(client.ID, client)
+	err := channel.AddClient(client.ID, client)
+	test.Equal(t, err, nil)
 
 	for i := 0; i < 25; i++ {
 		msg := NewMessage(topic.GenerateID(), []byte("test"))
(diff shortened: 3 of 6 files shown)


apps/nsqd/nsqd.go
```diff
@@ -119,6 +119,7 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Duration("max-output-buffer-timeout", opts.MaxOutputBufferTimeout, "maximum client configurable duration of time between flushing to a client")
 	flagSet.Duration("min-output-buffer-timeout", opts.MinOutputBufferTimeout, "minimum client configurable duration of time between flushing to a client")
 	flagSet.Duration("output-buffer-timeout", opts.OutputBufferTimeout, "default duration of time between flushing data to clients")
+	flagSet.Int("max-channel-client-connection-count", opts.MaxChannelClientConnectionCount, "maximum channel client connection count")
```
mreiferson: It needs to be possible to disable this feature, e.g. by setting the value to `0`?

mreiferson: Let's name this `max-channel-subscribers`?

mreiferson:
Thanks for the PR @andyxning!

apps/nsqd/nsqd.go
ploxiln: yeah - and I think, for backwards compatibility, it should default to unlimited?

mreiferson: +1

mreiferson: Actually that name isn't right either, because it's only on a per-node basis.

Thinking out loud... how about `max-local-channel-subscriptions`?

nsqd/protocol_v2.go
```diff
@@ -622,6 +622,10 @@ func (p *protocolV2) SUB(client *clientV2, params [][]byte) ([]byte, error) {
 			time.Sleep(1 * time.Millisecond)
 			continue
 		}
+		if channel.IsClientConnectionCountExceeded() {
```
ploxiln: I think this check could be inside `channel.AddClient()` - that way the channel lock is only taken once, and you avoid some races:
  * multiple clients connect concurrently, both are added, then both check if limit exceeded, both get True even though only one should
  * not sure but maybe: client added, can get some messages? then disconnects because limit exceeded

so `Channel.AddClient()` could be adjusted to do this check and return an error.

apps/nsqd/nsqd.go
ploxiln: agreed, could use a better name here

andyxning: Sounds reasonable. Will do.

andyxning: I think `max-channel-subscriptions` should be fine as other command line arguments are also in this pattern but they all apply to single nsqd instance.

andyxning: @mreiferson @ploxiln What do you think. Naming is the top one difficult problem. :)

nsqd/protocol_v2.go
andyxning: Good catch. Will do.

andyxning: > not sure but maybe: client added, can get some messages? then disconnects because limit exceeded

This should not happen. As this logic is all under `SUB` command callback. As an error raised in `SUB` command callback function, all io operations will not be further invoked.

apps/nsqd/nsqd.go
ploxiln: My take: `max-channel-consumers` and explain the "local" aspect in the description. But I'll defer to @mreiferson :)

andyxning: In nsq world, consumer is more official than subscription. I will prefer `max-channel-consumers`

andyxning:
All comments are addressed. PTAL. @mreiferson @ploxiln

nsqd/channel.go
```diff
@@ -389,15 +389,22 @@ func (c *Channel) RequeueMessage(clientID int64, id MessageID, timeout time.Dura
...
+	}
+
+	maxChannelConsumers := c.ctx.nsqd.getOpts().MaxChannelConsumers
+	if maxChannelConsumers == 0 || (maxChannelConsumers != 0 && len(c.clients) < maxChannelConsumers) {
+		c.clients[clientID] = client
+		return nil
 	}
-	c.clients[clientID] = client
+
+	return errors.New("E_TOO_MANY_CHANNEL_SUBSCRIPTIONS")
```
mreiferson: style nit-pick, I'd prefer the non-error case to be the default path and the error case to be the branch, please :)

nsqd/channel_test.go
```diff
@@ -152,7 +152,8 @@ func TestChannelEmptyConsumer(t *testing.T) {
 	channel := topic.GetChannel("channel")
 	client := newClientV2(0, conn, &context{nsqd})
 	client.SetReadyCount(25)
-	channel.AddClient(client.ID, client)
+	err := channel.AddClient(client.ID, client)
+	test.Equal(t, err, nil)
```
mreiferson: this reminded me, it should be pretty easy to add a new test that validates behavior when the max is exceeded

apps/nsqd/nsqd.go
mreiferson: SGTM

nsqd/channel.go
andyxning: Will do.

nsqd/channel_test.go
andyxning: Will do.

nsqd/protocol_v2.go
```diff
@@ -615,7 +615,10 @@ func (p *protocolV2) SUB(client *clientV2, params [][]byte) ([]byte, error) {
 	for {
 		topic := p.ctx.nsqd.GetTopic(topicName)
 		channel = topic.GetChannel(channelName)
-		channel.AddClient(client.ID, client)
+		if err := channel.AddClient(client.ID, client); err != nil {
+			return nil, protocol.NewFatalClientErr(nil, "E_TOO_MANY_CHANNEL_CONSUMERS",
+				fmt.Sprintf("TOO many channel consumers for %s:%s", topicName, channelName))
```
mreiferson: Last nit, let's change this message to be:

```go
fmt.Sprintf("channel consumers for %s:%s exceeds limit of %d", topicName, channelName, ...)
```

🙏

andyxning: Will do.

andyxning: @mreiferson Done. PTAL again. :)

mreiferson:
LGTM, I'll leave it for @ploxiln to take a final pass.

Thanks @andyxning! 💯

andyxning:
BTW, do you have a slack id for other IM id that is available for me to give you a DM about PR #1137 .

mreiferson:
Yes, I'm `mreiferson` on [Gophers Slack](https://invite.slack.golangbridge.org/).

andyxning:
@ploxiln PTAL.

ploxiln:
👍

andyxning:
@ploxiln Does it ok to paste your IM chat name here in order to give you a DM about #1137 ?

ploxiln:
I'm not on gophers slack ... but I am _occasionally_ on Freenode IRC #nsq channel, using the same name, when I remember to start up my irc client :)

It's very late here at the moment so I won't be on within the next 10 hours probably.

PR #1138, opened 2019-02-25
nsqd/nsqlookupd: exit when tcp server accept error

andyxning:
This is a follow up PR for #1135 .

### DIFF
--- a/internal/protocol/tcp_server.go
+++ b/internal/protocol/tcp_server.go
@@ -12,9 +12,10 @@ type TCPHandler interface {
 	Handle(net.Conn)
 }
 
-func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
+func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) error {
 	logf(lg.INFO, "TCP: listening on %s", listener.Addr())
 
+	var fatalErr error
 	for {
 		clientConn, err := listener.Accept()
 		if err != nil {
@@ -25,7 +26,8 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
 			}
 			// theres no direct way to detect this error because it is not exposed
 			if !strings.Contains(err.Error(), "use of closed network connection") {
-				logf(lg.ERROR, "listener.Accept() - %s", err)
+				logf(lg.FATAL, "listener.Accept() - %s", err)
+				fatalErr = err
 			}
 			break
 		}
@@ -33,4 +35,5 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
 	}
 
 	logf(lg.INFO, "TCP: closing %s", listener.Addr())
+	return fatalErr
 }
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -74,6 +74,8 @@ type NSQD struct {
 	exitChan             chan int
 	waitGroup            util.WaitGroupWrapper
 
+	shuttingDown bool
+
 	ci *clusterinfo.ClusterInfo
 }
 
@@ -263,7 +265,9 @@ func (n *NSQD) Main() {
 
 	tcpServer := &tcpServer{ctx: ctx}
 	n.waitGroup.Wrap(func() {
-		protocol.TCPServer(n.tcpListener, tcpServer, n.logf)
+		if err := protocol.TCPServer(n.tcpListener, tcpServer, n.logf); err != nil {
+			go n.Exit()
+		}
 	})
 	httpServer := newHTTPServer(ctx, false, n.getOpts().TLSRequired == TLSRequired)
 	n.waitGroup.Wrap(func() {
@@ -436,6 +440,13 @@ func (n *NSQD) Exit() {
 	}
 
 	n.Lock()
+
+	if n.shuttingDown {
+		n.Unlock()
+		return
+	}
+	n.shuttingDown = true
+
 	err := n.PersistMetadata()
 	if err != nil {
 		n.logf(LOG_ERROR, "failed to persist metadata - %s", err)
--- a/nsqlookupd/nsqlookupd.go
+++ b/nsqlookupd/nsqlookupd.go
@@ -21,6 +21,8 @@ type NSQLookupd struct {
 	httpListener net.Listener
 	waitGroup    util.WaitGroupWrapper
 	DB           *RegistrationDB
+
+	shuttingDown bool
 }
 
 func New(opts *Options) *NSQLookupd {
@@ -62,7 +64,9 @@ func (l *NSQLookupd) Main() error {
 
 	tcpServer := &tcpServer{ctx: ctx}
 	l.waitGroup.Wrap(func() {
-		protocol.TCPServer(tcpListener, tcpServer, l.logf)
+		if err := protocol.TCPServer(tcpListener, tcpServer, l.logf); err != nil {
+			go l.Exit()
+		}
 	})
 	httpServer := newHTTPServer(ctx)
 	l.waitGroup.Wrap(func() {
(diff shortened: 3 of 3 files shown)


internal/protocol/tcp_server.go
```diff
@@ -33,4 +34,5 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
 	}
 
 	logf(lg.INFO, "TCP: closing %s", listener.Addr())
+	return errors.New("accept new tcp connection error")
```
ploxiln: There's not much point in always returning the same error. Consider that this also happens for a normal exit (when it closes the Listener) and then because this returned an error, the wrapper func will call NSQD.Exit() again concurrently.

I suggest an NSQD "exiting" bit. In the wrapper (below in nsqd.go), when this function TCPServer finishes, if the exiting flag is already set, that's normal. If the exiting flag is not set, that is unexpected, and it should call NSQD.Exit().

nsqd/nsqd.go
```diff
@@ -263,7 +263,9 @@ func (n *NSQD) Main() {
 
 	tcpServer := &tcpServer{ctx: ctx}
 	n.waitGroup.Wrap(func() {
-		protocol.TCPServer(n.tcpListener, tcpServer, n.logf)
+		if err := protocol.TCPServer(n.tcpListener, tcpServer, n.logf); err != nil {
+			n.Exit()
```
ploxiln: ... and if this wrapper function calls NSQD.Exit() it has to launch a separate goroutine for it, because it will deadlock with the waitGroup waiting for this wrapper function to finish.

andyxning: Good catch. Will investigate

andyxning:
@ploxiln PTAL.

internal/protocol/tcp_server.go
```diff
@@ -26,11 +28,17 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
...
 		}
 		go handler.Handle(clientConn)
 	}
 
 	logf(lg.INFO, "TCP: closing %s", listener.Addr())
+	if fatalErr != nil {
+		return errors.New("accept new tcp connection error")
+	}
+
+	return nil
```
ploxiln: You could just `return fatalErr` in all cases, it will be `nil` if never set.

But actually, I suggest reverting this function to not return an error at all.

nsqd/nsqd.go
```diff
@@ -263,7 +265,9 @@ func (n *NSQD) Main() {
 
 	tcpServer := &tcpServer{ctx: ctx}
 	n.waitGroup.Wrap(func() {
-		protocol.TCPServer(n.tcpListener, tcpServer, n.logf)
+		if err := protocol.TCPServer(n.tcpListener, tcpServer, n.logf); err != nil {
+			go n.Exit()
```
ploxiln: I suggest checking the exiting flag here, and if it is not already set, then go Exit().

The flag variable could be an atomic int, to avoid the need to take a lock.

ploxiln: (this suggestion would require Exit() to set the flag as the first thing it does)

andyxning: I suggest we keep the check logic be within the `Exit` function. With this we should not take care about the internal exit status flag check before `Exit` is called in other places. And we should keep `Exit` function to be called only once.

With setting and checking separated with aotmic values, race condition may happens. For example, a normal exit and accept error exit happens simultaneously, before the normal exit call set the Exit flag, accept error exit checks it and then call Exit again. Although nsqd will exit correctly but it is called more than once.

With lock like this. Exit function logic except closing listeners will only be called once.

andyxning: BTW, i am not quite sure about your intention, lock consumption should be ok and will not invoke any new performance problems, because all this happens under exit phase.

mreiferson:
An alternative way to approach this, that would avoid having to protect Exit() from multiple calls would be to create a `fatalErrCh` channel, that gets passed to [`svc.Run`](https://github.com/nsqio/nsq/blob/master/apps/nsqd/nsqd.go#L190). We could `close()` it [here](https://github.com/nsqio/nsq/pull/1138/files#diff-13e146c62dc03f79a6c5963c9c13d4cdR269).

internal/protocol/tcp_server.go
```diff
@@ -26,11 +27,13 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
 			// theres no direct way to detect this error because it is not exposed
 			if !strings.Contains(err.Error(), "use of closed network connection") {
 				logf(lg.ERROR, "listener.Accept() - %s", err)
```
mreiferson: If we take the approach I proposed, this log line should communicate that the error is `FATAL`.

nsqd/nsqd.go
ploxiln: The reason I suggested the exiting flag early on is because deciding to call Exit here could be as simple as:

```diff
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -71,6 +71,7 @@ type NSQD struct {
 
        notifyChan           chan interface{}
        optsNotificationChan chan struct{}
+       exitFlag             int32
        exitChan             chan int
        waitGroup            util.WaitGroupWrapper
 
@@ -264,6 +265,10 @@ func (n *NSQD) Main() {
        tcpServer := &tcpServer{ctx: ctx}
        n.waitGroup.Wrap(func() {
                protocol.TCPServer(n.tcpListener, tcpServer, n.logf)
+               if atomic.LoadInt32(&t.exitFlag) == 0 {
+                       // abnormal listen loop exit
+                       go n.Exit()
+               }
        })
        httpServer := newHTTPServer(ctx, false, n.getOpts().TLSRequired == TLSRequired)
        n.waitGroup.Wrap(func() {
@@ -423,6 +428,10 @@ func (n *NSQD) PersistMetadata() error {
 }
 
 func (n *NSQD) Exit() {
+       if !atomic.CompareAndSwapInt32(&t.exitFlag, 0, 1) {
+               return
+       }
+
        if n.tcpListener != nil {
                n.tcpListener.Close()
        }
```

(but there still is the issue of exiting with an error code)

nsqd/nsqd.go
```diff
@@ -436,6 +440,12 @@ func (n *NSQD) Exit() {
 	}
 
 	n.Lock()
+
+	if n.shuttingDown {
+		return
```
ploxiln: should unlock if returning here

nsqd/nsqd.go
mreiferson: Agreed that the atomic would be slightly cleaner, but I prefer directing flow through the existing machinery that's setup to manage the lifecycle of the service rather than introducing a new exceptional path — to ensure that, whatever `svc.Run` is doing, it gets done.

ploxiln: yeah ... calling NSQD.Exit() cleans up but does not actually exit ... go-svc needs to be involved anyway

nsqd/nsqd.go
andyxning: Good catch.

internal/protocol/tcp_server.go
andyxning: Done.

nsqd/nsqd.go
ploxiln: one more cheesy idea to avoid needing to deal with go-svc:

```diff
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -264,6 +264,12 @@ func (n *NSQD) Main() {
        tcpServer := &tcpServer{ctx: ctx}
        n.waitGroup.Wrap(func() {
                protocol.TCPServer(n.tcpListener, tcpServer, n.logf)
+               if atomic.LoadInt32(&t.exitFlag) == 0 {
+                       go func() {
+                               n.Exit()
+                               os.Exit(1)
+                       }()
+               }
        })
```

mreiferson: Meh, let's just do it the "right" way.

mreiferson:
Let's finish this up in #1140, thanks @andyxning

PR #1140, opened 2019-02-26
nsqd/nsqlookupd: properly handle fatal accept errors

mdh67899:
#1138 
> An alternative way to approach this, that would avoid having to protect Exit() from multiple calls would be to create a `fatalErrCh` channel, that gets passed to [`svc.Run`](https://github.com/nsqio/nsq/blob/master/apps/nsqd/nsqd.go#L190). We could `close()` it [here](https://github.com/nsqio/nsq/pull/1138/files#diff-13e146c62dc03f79a6c5963c9c13d4cdR269).


apps/nsqd/nsqd.go
```diff
@@ -246,3 +247,10 @@ func (p *program) Stop() error {
 	}
 	return nil
 }
+
+func (p *program) GetExitChan() (chan interface{}, error) {
```
mreiferson: I don't think we need to change the API of `go-svc` to accomplish this. We can just pass `fatalErrCh` to `svc.Run` (it can accept multiple channels in its function signature).

mdh67899: I am confused. I can't imagine how to pass `fatalErrCh` to `svc.Run` without change the API of `go-svc`,  Can you write a code example?
https://github.com/judwhite/go-svc/blob/d83e900e4a688aad49f41263bebf8793f0bf6add/svc/svc_other.go#L14-L33

mreiferson: Ahhh, sorry, you're right, I misread the API.

😭

judwhite: Yeah, it takes signals, not channels.

What's wrong with `sync.Once`?

mdh67899: > What's wrong with `sync.Once`?

if the nsqd `tcpServer` and `httpServer` has some error at the same time,  they will trying to close `fatalErrChan` twice, the panic will happens, so we should add sync.once make sure we will close `fatalErrChan` once

mdh67899: @mreiferson @judwhite It seems that we need to change the API of `go-svc`, I have submitted my change to `https://github.com/judwhite/go-svc`, I did not expect to merge it, thanks.

judwhite: @mdh67899 Can you walk me through the problem, and proposed solution? I'm coming in with very little context, would be helpful and save me some time going through previous PR's if you can help explain the intent.

mdh67899: @judwhite
In the `svc.Run()` method, after we execute the `service.Init()` and `service.Start()` method, we immediately listen to the system signal

Imagine this situation: There are some problems inside the program, we want to use `service.Stop ()` method to make sure we will  exit program more graceful

`svc.Run ()` method does not provide the notification interface of the program, we can not actively notify the service smooth exit program, we can only exit with `os.Exit ()` way

my suggestion is: we shoud  add `GetExitChan()` method inside `service`, after the program finishes `service.Init ()` and `service.Start ()`, if you receive the system signals or the exit signal given by the program, we will exit with `service.Stop ()` method

sorry for my poor english.

judwhite: @mdh67899 You could create your own channel and spin up a goroutine:

```go
go func() {
    <-quit
    service.Stop() 
}
```

Close `quit` wherever you want. Probably protect `Stop()` from executing more than once, in case that'd be a bad thing.. say if the error condition is hit at the same time a `SIGTERM` is received.

Crash-only can also be a valid strategy (`os.Exit(1)` or `log.Fatal`), though it can be a lot of effort to deliver the same experience if you don't have crash-restart-crash-restart pretty solid.

mreiferson: Ahh thats a good suggestion, thanks @judwhite

ploxiln:
it looks to me like, after all this, we ended up with something equivalent to https://github.com/nsqio/nsq/pull/1138#discussion_r260128676 (check if already doing exit, call `NSQD.Exit()`, call `os.Exit(1)`)

mreiferson:
Indeed, thank you for pointing that out. I still feel like the application (not `nsqd` "library") should have the responsibility of handling "exiting" and that the library is just missing a way to signal that.

I'm going to try to modify this PR on my own to see if I can get it to a place where I'm reasonably happy with it and then we can discuss.

ploxiln:
I do have one more idea: send `SIGINT` to `getpid()`. Not sure if it works on windows though.

mreiferson:
pushed up my changes, if this seems reasonable I'll apply the same approach to `nsqlookupd`

mreiferson:
(it does not yet even compile, mostly due to logging related bullshit)

ploxiln:
`NSQD.New()` returning error makes good sense.

`NSQD.Main()` becoming synchronous ... is going to change things for tests

mreiferson:
> `NSQD.Main()` becoming synchronous ... is going to change things for tests

Yea, for sure, I'll clean all that up. I think the semantics of it being synchronous make more sense to me. I've come around to the idea that libraries shouldn't try to own concurrency in this way, and we've made this mistake in a few places (mostly thinking go-nsq). Also, we've never promised an API guarantee for `nsqd` as a library, so we don't need to consider backwards compat.

apps/nsqd/nsqd.go
```diff
@@ -218,31 +221,41 @@ func (p *program) Start() error {
...
+		lg.LogFatal("[nsqd] ", "failed to persist metadata - %s", err)
 	}
-	nsqd.Main()
 
-	p.nsqd = nsqd
+	go func() {
+		err := p.nsqd.Main()
+		if err != nil {
+			p.Stop()
+			lg.LogFatal("[nsqd] ", "%s", err)
```
ploxiln: might want to log before `Stop()` so error is logged closer to when it happened
(could log in the `once.Do(func() {exitCh <- err})` and then again here at fatal exit)

ploxiln:
I do like the "return err instead of just os.Exit(1)" pattern

apps/nsqd/nsqd.go
mreiferson: Yea, was just being lazy because `lg.LogFatal` handles the `os.Exit(1)` as implemented... For the record, I'm unhappy with how these logging changes ended up — they're an improvement from what was there, but also ugly in some ways, so I'm open to ideas if you have any.

andyxning:
> I do have one more idea: send `SIGINT` to `getpid()`. Not sure if it works on windows though.

This is what i think, also. This way we do not need to any api.

nsqd/nsqd.go
```diff
@@ -239,40 +231,46 @@ func (n *NSQD) RemoveClient(clientID int64) {
...
-			os.Exit(1)
+			return fmt.Errorf("listen (%s) failed - %s", n.getOpts().HTTPSAddress, err)
 		}
 	}
 
+	exitCh := make(chan error)
+	var once sync.Once
+	exitFunc := func(err error) {
+		once.Do(func() {
+			n.logf(LOG_FATAL, "%s", err)
```
ploxiln: I think this will still happen once during normal shutdown, usually with a `nil` error

ploxiln:
This way looks OK to me.

mreiferson:
Deeper down the rabbit hole...

Commit d6b87a5 improves how `LogLevel` is parsed (moves responsibility to application side flag parsing, and uses strict types on the `nsqd` side `Options` struct). This required [changes](https://github.com/mreiferson/go-options/pull/17) to `go-options`.

Commit 76f3702 should start to fix tests, which required moving some initializing to `New()`, which I'm interested in feedback on.

apps/nsqd/options.go
```diff
@@ -73,14 +63,45 @@ func (t *tlsMinVersionOption) String() string {
...
 
 	// basic options
 	flagSet.Bool("version", false, "print version string")
 	flagSet.String("config", "", "path to config file")
 
-	flagSet.String("log-level", "info", "set log verbosity: debug, info, warn, error, or fatal")
+	logLevel := lg.LogLevelOption(opts.LogLevel)
+	flagSet.Var(&logLevel, "log-level", "set log verbosity: debug, info, warn, error, or fatal")
 	flagSet.String("log-prefix", "[nsqd] ", "log message prefix")
 	flagSet.Bool("verbose", false, "deprecated in favor of log-level")
```
ploxiln: might want to clarify "verbose" is now "ignored - use log-level"
(I dunno if the `Verbose` options struct member can be removed?)

(moved this comment from the commit to the PR)

internal/lg/lg.go
```diff
@@ -41,35 +43,51 @@ func (l LogLevel) String() string {
...
+		return err
 	}
-	return lvl, nil
+	*l = LogLevelOption(lvl)
+	return nil
 }
 
+func (l *LogLevelOption) Get() interface{} { return LogLevel(*l) }
+
+func (l *LogLevelOption) String() string { return LogLevel(*l).String() }
```
ploxiln: just curious, why make another type, instead of just giving LogLevel these methods? is it because this `String()` needs to be for a pointer type?

mreiferson: that's a damn good question

apps/nsqd/options.go
mreiferson: yep

mreiferson:
Finally have a green test run, PTAL and I'll clean up these commits.

ploxiln:
```
$ ./build/nsqadmin --verbose --nsqd-http-address=http://127.0.0.1:4151
panic: interface conversion: *app.StringArray is not flag.Getter: missing method Get

goroutine 1 [running]:
github.com/nsqio/nsq/vendor/github.com/mreiferson/go-options.Resolve(0x13a69a0, 0xc0000cc580, 0xc00002e360, 0x0)
	/Users/pierce/go/src/github.com/nsqio/nsq/vendor/github.com/mreiferson/go-options/options.go:86 +0xa3e
main.(*program).Start(0xc00000c440, 0x14be420, 0x17e0bf0)
	/Users/pierce/go/src/github.com/nsqio/nsq/apps/nsqadmin/main.go:105 +0x38a
github.com/nsqio/nsq/vendor/github.com/judwhite/go-svc/svc.Run(0x14c0e40, 0xc00000c440, 0xc00000c460, 0x2, 0x2, 0xc000066f20, 0x0)
	/Users/pierce/go/src/github.com/nsqio/nsq/vendor/github.com/judwhite/go-svc/svc/svc_other.go:20 +0x74
main.main()
	/Users/pierce/go/src/github.com/nsqio/nsq/apps/nsqadmin/main.go:72 +0xbf
```

mreiferson:
Sonofa! Thanks, will fix.

mreiferson:
OK, cleaned up the commits 🙏

ploxiln:
As far as I can tell, this is pretty good. It's not easy to trigger the failure paths ... this is as far as I got with very low ulimit and parallel http requests:

```
$ ./build/nsqd --tcp-address 0.0.0.0:530 --log-level=debug
[nsqd] 2019/03/02 11:27:03.212215 INFO: nsqd v1.1.1-alpha (built w/go1.11.5)
[nsqd] 2019/03/02 11:27:03.212320 INFO: ID: 136
[nsqd] 2019/03/02 11:27:03.212645 INFO: NSQ: persisting topic/channel metadata to nsqd.dat
[nsqd] 2019/03/02 11:27:03.214231 INFO: TCP: listening on [::]:530
[nsqd] 2019/03/02 11:27:03.214297 INFO: HTTP: listening on [::]:4151
[nsqd] 2019/03/02 11:27:21.991592 WARNING: http: Accept error: accept tcp [::]:4151: accept: too many open files; retrying in 5ms
[nsqd] 2019/03/02 11:27:29.912527 INFO: 200 GET /stats ([::1]:56231) 214.582µs
[nsqd] 2019/03/02 11:27:41.040258 WARNING: http: Accept error: accept tcp [::]:4151: accept: too many open files; retrying in 5ms
[nsqd] 2019/03/02 11:27:45.535751 WARNING: http: Accept error: accept tcp [::]:4151: accept: too many open files; retrying in 10ms
[nsqd] 2019/03/02 11:27:45.814051 WARNING: http: Accept error: accept tcp [::]:4151: accept: too many open files; retrying in 20ms
[nsqd] 2019/03/02 11:27:46.626209 WARNING: http: Accept error: accept tcp [::]:4151: accept: too many open files; retrying in 40ms
[nsqd] 2019/03/02 11:27:52.394460 INFO: 200 GET /stats ([::1]:56230) 204.084µs
```

Gopkg.lock
```diff
@@ -80,14 +80,14 @@
...
-  digest = "1:e6d1805ead5b8f2439808f76187f54042ed35ee26eb9ca63127259a0e612b119"
+  digest = "1:74ce8329385730bb2d01e4e91b66183beceeb1512084e4253653ac2439f1ef8c"
   name = "golang.org/x/sys"
   packages = [
     "windows",
     "windows/svc",
   ]
   pruneopts = ""
-  revision = "b4a75ba826a64a70990f11a225237acd6ef35c9f"
+  revision = "a34e9553db1e492c9a76e60db2296ae7e5fbb772"
```
ploxiln: the mod files still have the "b4a7..." revision of golang.org/x/sys

ploxiln:
other than that 👍

Gopkg.lock
mreiferson: fixed

apps/nsqd/main.go
```diff
@@ -0,0 +1,100 @@
...
+	if err := svc.Run(prg, syscall.SIGINT, syscall.SIGTERM); err != nil {
+		logFatal("%s", err)
+	}
+}
+
+func (p *program) Init(env svc.Environment) error {
+	if env.IsWindowsService() {
+		dir := filepath.Dir(os.Args[0])
+		return os.Chdir(dir)
+	}
```
judwhite: @mreiferson It doesn't hurt anything to have this, but the package does this bit for you now. I generally consider it impolite to `os.Chdir` but in this case I thought it was warranted. https://github.com/judwhite/go-svc/blame/d83e900e4a688aad49f41263bebf8793f0bf6add/svc/svc_windows.go#L59-L66

PR #1143, opened 2019-02-28
nsqd/nsqlookupd: remove useless LimitReader

andyxning:
As `net/http` have used `LimitReader` to read from http request with `Content-Length` size in 

https://github.com/golang/go/blob/61170f85e62f1326d42c4dbd8aa17ab4a1305a87/src/net/http/transfer.go#L540.

In nsqd, we could first only check the `Content-Length` against max message/body size and then just `io.RealAll`.


andyxning:
/close

Sorry for not testing this with chunked encoding http request. This change will make nsqd can not handle chunked request with large body size. This will make nsqd vulnerable.

PR #1145, opened 2019-03-01
nsqd/nsqlookupd: exit when tcp server accept error

andyxning:
This is a following up PR for https://github.com/nsqio/nsq/pull/1138

### DIFF
--- a/internal/protocol/tcp_server.go
+++ b/internal/protocol/tcp_server.go
@@ -12,9 +12,10 @@ type TCPHandler interface {
 	Handle(net.Conn)
 }
 
-func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) {
+func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) error {
 	logf(lg.INFO, "TCP: listening on %s", listener.Addr())
 
+	var fatalErr error
 	for {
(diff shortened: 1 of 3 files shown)


andyxning:
In this way, we also do not need to do any sync or protect about `Exit` invocation, because only after we send the signal to the process, `go-svc` can move on to do a `Stop`.

ploxiln:
I agree, this is concise and should work ... but @mreiferson has other motivations: he wants the stuff in `nsqd/` to not do stuff like call `os.Exit()` directly, but rather just return error so the stuff in `apps/nsqd/` has full control over the "lifecycle" of the process. And he also refactored the logging option handling.

I'm sorry that we've pushed aside your recent pull-requests in order to do it our own way. You have brought up good points and been responsive to concerns and requests for change, but these recent topics (disk-queue-format and fatal-error-clean-exit) are rather core to nsqd, and I think it's understandable that the maintainers want to do it their way. #1140 turned into more changes than I would have liked in a single pull request, but I'm inclined to give @mreiferson more trust and defer more to his judgement.

andyxning:
I respect the maintainers options. But my suggestion is that keep each PR simple and dedicated. :)

mreiferson:
@ploxiln said it well, and I appreciate your contributions @andyxning! #1140 will address this issue, and I'm pretty close to wrapping it up :)

PR #1147, opened 2019-03-13
nsq_to_file: include topic/channel for all logging output

mreiferson:
We've been running `nsq_to_file` with "topic discovery" mode on and none of the non go-nsq log messages include enough context to be useful.


ploxiln:
👍

PR #1149, opened 2019-03-20
nsqd: add compression/decompression for messages

### LINKED ISSUE #1148: nsqd: enable compression/decompression pipeline from producer to consumer
NSQ currently supports snappy compression and decompression for TCP-protocol based producer and consumer. But it has some shortcomings.
* the memory and disk capacity nsqd used can not be decreased since all messages are stored just like what nsqd receives, i.e., the original messages.
* the in network bandwidth can not be decreased since the message are not compressed before it is send.
* they are just available for TCP protocol based producer and consumer.

I have made a proposal for this: https://docs.google.com/document/d/107yDH6pN8-b_i22AfvPkzdDU2yneOC-1lcUllpUmN4w/edit#heading=h.tla1oogr3wac

/cc @mreiferson @ploxiln

andyxning:
Fix #1148 

This PR adds compression/decompression pipeline for nsqd server.

Update:
* add proposal: https://docs.google.com/document/d/107yDH6pN8-b_i22AfvPkzdDU2yneOC-1lcUllpUmN4w/edit#heading=h.j5jbn91ci2bi

/cc @mreiferson @ploxiln

### DIFF
--- a/internal/compress/compress.go
+++ b/internal/compress/compress.go
@@ -1,0 +1,33 @@
+package compress
+
+import (
+	"compress/flate"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"bytes"
+
+	"github.com/golang/snappy"
+)
+
+func SnappyDecompress(compressedMsg []byte) ([]byte, error) {
+	body, err := snappy.Decode(nil, compressedMsg)
+	if err != nil {
+		return nil, fmt.Errorf("error in decode snappy compressed message: %v", err)
+	}
+
+	return body, nil
+}
+
+func DeflateDecompress(compressedMsg []byte) ([]byte, error) {
+	br := bytes.NewReader(compressedMsg)
+	fr := flate.NewReader(br)
+	defer fr.Close()
+	body, err := ioutil.ReadAll(fr)
+	if err != nil && !(err == io.ErrUnexpectedEOF && br.Len() == 0) {
+		return nil, fmt.Errorf("error in decode deflate compressed message: %v", err)
+	}
+
+	return body, nil
+}
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -13,6 +13,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/nsqio/nsq/internal/compress"
 	"github.com/nsqio/nsq/internal/protocol"
 	"github.com/nsqio/nsq/internal/version"
 )
@@ -126,7 +127,28 @@ func (p *protocolV2) SendMessage(client *clientV2, msg *Message) error {
 	p.ctx.nsqd.logf(LOG_DEBUG, "PROTOCOL(V2): writing msg(%s) to client(%s) - %s", msg.ID, client, msg.Body)
 	var buf = &bytes.Buffer{}
 
-	_, err := msg.WriteTo(buf)
+	adjustedBody := msg.Body
+	var err error
+	if msg.snappyCompressed {
+		// message is snappy compressed
+		adjustedBody, err = compress.SnappyDecompress(msg.Body)
+		if err != nil {
+			p.ctx.nsqd.logf(LOG_ERROR, "PROTOCOL(V2): decompress msg(%s) with snappy error", msg.ID)
+			return err
+		}
+		msg.snappyCompressed = false
+	} else if msg.deflateCompressed {
+		// message is deflate compressed
+		adjustedBody, err = compress.DeflateDecompress(msg.Body)
+		if err != nil {
+			p.ctx.nsqd.logf(LOG_ERROR, "PROTOCOL(V2): decompress msg(%s) with deflate error", msg.ID)
+			return err
+		}
+		msg.deflateCompressed = false
+	}
+
+	msg.Body = adjustedBody
+	_, err = msg.WriteTo(buf)
 	if err != nil {
 		return err
 	}
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -307,17 +307,12 @@ func (t *Topic) messagePump() {
 			goto exit
 		}
 
-		for i, channel := range chans {
-			chanMsg := msg
-			// copy the message because each channel
-			// needs a unique instance but...
-			// fastpath to avoid copy if its the first channel
-			// (the topic already created the first copy)
-			if i > 0 {
-				chanMsg = NewMessage(msg.ID, msg.Body)
-				chanMsg.Timestamp = msg.Timestamp
-				chanMsg.deferred = msg.deferred
-			}
+		for _, channel := range chans {
+			chanMsg := NewMessage(msg.ID, msg.Body)
+			chanMsg.Timestamp = msg.Timestamp
+			chanMsg.deferred = msg.deferred
+			chanMsg.deflateCompressed = msg.deflateCompressed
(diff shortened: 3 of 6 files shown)


andyxning:
Wait for @ploxiln 's re-design about nsqd storage backend.

nsqd/topic.go
```diff
@@ -307,17 +307,12 @@ func (t *Topic) messagePump() {
...
-			// copy the message because each channel
-			// needs a unique instance but...
-			// fastpath to avoid copy if its the first channel
-			// (the topic already created the first copy)
-			if i > 0 {
-				chanMsg = NewMessage(msg.ID, msg.Body)
-				chanMsg.Timestamp = msg.Timestamp
-				chanMsg.deferred = msg.deferred
-			}
+		for _, channel := range chans {
```
andyxning: This is mainly used to avoid the race condition between topic's `PutMessage` method which is detected by the `go test -race`.

nsqd/protocol_v2.go
```diff
@@ -126,7 +127,28 @@ func (p *protocolV2) SendMessage(client *clientV2, msg *Message) error {
...
+		adjustedBody, err = compress.DeflateDecompress(msg.Body)
+		if err != nil {
+			p.ctx.nsqd.logf(LOG_ERROR, "PROTOCOL(V2): decompress msg(%s) with deflate error", msg.ID)
+			return err
+		}
+		msg.deflateCompressed = false
+	}
+
+	msg.Body = adjustedBody
+	_, err = msg.WriteTo(buf)
```
mreiferson: I'm still confused why the changes here are necessary, the existing compression functionality already delivers compressed messages to consumers?

andyxning: Because the snappy or deflate compression is implemented by wraping the socket, i.e., nsqd will write original message over the wrapped socket and the compressed message will be actually be written to the wire. Reading is the same.

internal/compress/compress.go
```diff
@@ -0,0 +1,31 @@
...
+		return nil, fmt.Errorf("error in decode snappy compressed message: %v", err)
+	}
+
+	return body, nil
+}
+
+func DeflateDecompress(compressedMsg []byte) ([]byte, error) {
+	fr := flate.NewReader(bytes.NewReader(compressedMsg))
+	defer fr.Close()
+	body, err := ioutil.ReadAll(fr)
```
andyxning: mark

internal/compress/compress.go
```diff
@@ -0,0 +1,33 @@
...
+
+	return body, nil
+}
+
+func DeflateDecompress(compressedMsg []byte) ([]byte, error) {
+	br := bytes.NewReader(compressedMsg)
+	fr := flate.NewReader(br)
+	defer fr.Close()
+	body, err := ioutil.ReadAll(fr)
+	if err != nil && !(err == io.ErrUnexpectedEOF && br.Len() == 0) {
```
andyxning: new way

nsqd/protocol_v2.go
mreiferson: Isn't that what we want to happen?

As we had discussed on https://github.com/nsqio/nsq/issues/1148#issuecomment-475809636, I feel like the way to accomplish what you want is for your producers and consumers to compress data.

It probably makes sense to get on the same page about the feature and whether or not it belongs in NSQ core before we discuss code and implementation.

andyxning: Yes. @mreiferson  You're right. I have also consider this pattern about compressing message in producer and decompressing it in consumer. However, On this way we need to do:
1、pass the compression algorithm information to the consumer for it to determine how to decompress it.
2、not all compression algorithm libs are supported by Golang/Python standard lib, for example snappy. In case to decompress the message correctly, consumer should know which compression algorithm the message use or the message can not be consumed correctly and will be requeued or even worse dropped directly once the `max_tries` is reached(I know that we can just print an error log and exit the consumer to indicate that something is wrong, but how about users set the `max_tries` to 1 and the message will be dropped next time it is consumed).

mreiferson: How you compress the messages is entirely up to you. In most cases you control both the producer and consumer — is there a need for more than one compression type to be used?

If multiple compression formats do need to be supported, you could also build some identifier into the message structure to indicate to consumers how to decompress.

ploxiln: If you have to transition from uncompressed messages to compressed messages, one way to do it is to start a new topic for the compressed messages ... so if you have "events" you might create another topic "events_gzip" and during the transition consume from both. Messages from "events_gzip" are gzipped, in this hypothetical case. But pick whatever is supported by both producer and consumer.

This scheme is nice because nsqd does not have to compress for storage and then decompress for delivery.

andyxning: > If multiple compression formats do need to be supported, you could also build some identifier into the message structure to indicate to consumers how to decompress.

I know this. But not every case we both control producer and consumer nor every producer need add compression byte info into a message. 

> If you have to transition from uncompressed messages to compressed messages, one way to do it is to start a new topic for the compressed messages ... so if you have "events" you might create another topic "events_gzip" and during the transition consume from both. Messages from "events_gzip" are gzipped, in this hypothetical case. But pick whatever is supported by both producer and consumer.

Yes. I agree with this. But what i want to express is to do this transparently to end users(producers and consumers).

internal/compress/compress.go
waruqi: Great! :+1:

andyxning:
Let's wait for https://github.com/nsqio/nsq/pull/1170 to be merged, then it will be easy to add this feature. :)

mreiferson:
@andyxning sorry I've been away for so long!

Context: the intended use case of the existing compression features of `nsqd` can best be described as "proxy compression", similar to using something like `nginx` and enabling gzip compression.

I noticed that you had a design doc linked in the original issue description with the following motivation:

> NSQ currently supports snappy compression and decompression for TCP-protocol based producer and consumer. But it has some shortcomings.
> 1. the memory and disk capacity nsqd used can not be decreased since all messages are stored just like what nsqd receives, i.e., the original messages.
> 2. the in-network bandwidth can not be decreased since the message are not compressed before it is send.
> 3. they are just available for TCP protocol based producer and consumer.

I think we should discuss these individually:

1. This seems like a reasonable request, at least for storage on disk, let's consider this for #1170 

2/3. When you publish messages via TCP protocol and enable compression on the connection, messages are compressed over the wire to `nsqd`. A reasonable addition could be to add Snappy and Deflate support _specifically to HTTP publish methods_. (There is no HTTP consumer, so it doesn't matter on the read side).

For this PR specifically, I would consider landing just the changes to HTTP publish methods to support `Content-Encoding`.

What do you think @andyxning and @ploxiln?

jehiah:
👍 for adding `Content-Encoding` support on HTTP publish methods

ploxiln:
👍 from me as well for HTTP Content-Encoding gzip/deflate (since those are already standard).

That probably won't satisfy @andyxning ... but I still think the way to go, when you really need good compression, is for the producer/consumer to do it. That way avoids nsqd doing extra compress and uncompress work for each message, and allows you to squeeze the most efficiency out of the system, if you need it.

mreiferson:
Closing this, opened #1313 to track compression over HTTP pub.

PR #1153, opened 2019-03-26
nsqd: close connections that don't send magic

JoseFeng:
Prevent a lot of CLOSE_WAIT。

The probe service detects if nsqd is alive.
Since there is no call to close here, a lot of CLOSE_WAIT will be generated.


mreiferson:
This looks OK to me, will leave for others to take a peek.

JoseFeng:
> This LGTM, however i think there is one more spot to fix the same issue.
> 
> Can you also add a `clientConn.Close()` on line 46?


It's not necessary here.
Because close is done in `prot.IOLoop(clientConn)`

PR #1159, opened 2019-06-05
nsqd: set memoryMsgChan as nil when --mem-queue-size=0

bitpeng:
Note: generally, we intend to save all the message in the disk-queue when set `-mem-queue-size=0` option, but `make(chan *Message, 0)` just create a unbuffered chan which can also send/receive message as well.

the Go channel spec dictates that nil channel operations (read or write) in a select are skipped, we set memoryMsgChan as not nil noly when mem-queue-size > 0 as shown in [diskqueue.go](https://github.com/nsqio/go-diskqueue/blob/master/diskqueue.go#L636)

### DIFF
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -77,10 +77,14 @@ func NewChannel(topicName string, channelName string, ctx *context,
 	c := &Channel{
 		topicName:      topicName,
 		name:           channelName,
-		memoryMsgChan:  make(chan *Message, ctx.nsqd.getOpts().MemQueueSize),
+		memoryMsgChan:  nil,
 		clients:        make(map[int64]Consumer),
 		deleteCallback: deleteCallback,
 		ctx:            ctx,
+	}
+	// create mem-queue only if size > 0 (do not use unbuffered chan)
+	if ctx.nsqd.getOpts().MemQueueSize > 0 {
+		c.memoryMsgChan = make(chan *Message, ctx.nsqd.getOpts().MemQueueSize)
 	}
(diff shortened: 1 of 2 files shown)


bitpeng:
Thanks for your warmly, patient answer, but in my opinion it should give the user an option to have a chance to save all the messages on the disk-queue. What about you? @mreiferson

ploxiln:
This isn't really enough to "guarantee" that, though.

The disk segments are not necessarily synced after every message, so the message might still be buffered in the page cache and not written to the physical disk, though you could set `--sync-every` to `1` to address that, but it would hurt performance, quite significantly.

But more importantly, messages are taken out of the disk queue while they are in flight, and exist in the in-memory in-flight pqueue. If the server crashes hard at this point, they will probably be lost. They may still exist in some disk segment file, but the metadata will say that they are in an already-consumed part of that disk segment file, and they won't be re-played. (And they might have been in a segment file that was finished and cleaned-up).

To get what you're going for, you need something like #625

LiPL2017:
Without considering the network factors,It's very usefull when i need to keep message order.

ploxiln:
OK ... I'm convinced this change is a good idea. There are still serious limitations to the "disk-queue-only" mode which make me skeptical of its value, but anyone who configured `--mem-queue-size=0` very likely wanted and expected every message to go into the diskqueue.

ploxiln:
I took the liberty of changing the code and comment style. Local testing with `--mem-queue-size=0` shows that nsqd does not lock up or spin or anything crazy. I'll merge later today if there are no objections.

PR #1163, opened 2019-06-18
nsqd: add random load balancing for Authd query

shenhui0509:
In our production environment
1. we have proxies backended with nsqds
2. auth is on
3. proxy uses a connection pool to request on nsqd
when there is high pub concurrency, proxy will create new connection, thus will query authd. The first authd will be overload, and the latency will increase, result in the risk of avalanche. Add a load balance for authd can ease that.
on consumer sides, we also have a large number of consumer instances, Load balance for authd is also benefit.
ready for review

### DIFF
--- a/internal/auth/authorizations.go
+++ b/internal/auth/authorizations.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"math/rand"
 	"net/url"
 	"regexp"
 	"time"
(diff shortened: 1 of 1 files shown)


shenhui0509:
ready for review

shenhui0509:
> 👍 this looks great.
> 
> One last thing, can you rebase and squash your commits down to a single commit w/ an appropriate commit message before we merge?

resolved

andyxning:
@jehiah Thanks for reviewing! Commits are squashed. PTAL.

PR #1172, opened 2019-08-01
fix tcpServer handle

wzjgo1:


nsqd/tcp.go
```diff
@@ -31,7 +31,7 @@ func (p *tcpServer) Handle(clientConn net.Conn) {
 
 	var prot protocol.Protocol
 	switch protocolMagic {
-	case "  V2":
+	case "V2":
```
ploxiln: This line has been here since 2013. See also: https://github.com/nsqio/go-nsq/blob/master/protocol.go#L14

Issue #1185, opened 2019-09-09
Question Involving one topic being spread across NSQds


jenkinsdonovan:
I have a question involving an NSQ topology that I can't find an answer for anywhere.

My use case is a distributed network with about 10 servers. Each server will have an instance of NSQ, and one "master" will have an instance of NSQLookup and NSQAdmin. Data will be entered into topics by web crawlers in Docker containers spun up by Kubernetes.

My question is, due to how I intend for Kubernetes to work, it is undefined which server will have which crawler - and one server can be running two different crawlers at the same time (different crawlers will input information into different topics).

How will this work? When I want to request data from one topic, how will NSQLookup decide which NSQ instance to pull from, given the topic is spread across multiple NSQ instances?

I apologize if this is a trivial question, or if it has been already asked.

jenkinsdonovan:
update - i've referenced #1078 and #980, but still stuck.

ploxiln:
You can ask general questions on the "google group" mailing list: https://groups.google.com/forum/#!forum/nsq-users

A topic can exist on multiple nsqd, and this is common. When a consumer process uses nsqlookupd to find a topic, it gets all nsqd which have that topic, and connects to them all. The consumer process continues to poll nsqlookupd (by default every 30 seconds I think) and if the topic appears on a new nsqd the consumer process will soon find out.

ploxiln:
see https://nsq.io/overview/design.html#eliminating-spofs

PR #1186, opened 2019-09-11
nsqadmin: fix nodes list links with ipv6 addresses

ploxiln:
alternative to #1179 

@andyxning please try this. I've clicked around and all links seem to work. (be sure to do a full refresh of nsqadmin, if you have the web page generated from a previous version already open in a tab.)

test setup:
  * `./build/nsqlookupd`
  * `./build/nsqd --broadcast-address=::1 --lookupd-tcp-address=[::1]:4160`
  * `curl localhost:4151/pub?topic=whatever -d '{"foo": 20}' ; echo`
  * `./build/nsq_tail --lookupd-http-address=[::1]:4161 --topic=whatever --channel=t3`
  * `./build/nsqadmin --lookupd-http-address=[::1]:4161`

cc @jehiah


ploxiln:
note that the topic and channel pages do this, and it results in the correct links:

```
        {{#each nodes}}
        <tr>
            <td>
                <button class="btn-link red tombstone-link" data-node="{{node}}" data-topic="{{../name}}" style="padding: 0 6px; border: 0;">✘</button>
                {{#if show_broadcast_address}}
                {{hostname_port}} (<a class="link" href="{{basePath "/nodes"}}/{{node}}">{{node}}</a>)
                {{else}}
                <a class="link" href="{{basePath "/nodes"}}/{{node}}">{{hostname_port}}</a>
                {{/if}}
```

andyxning:
@ploxiln I have tested this locally. But it can not be operated successfully.





Have you test this PR against non localhost ipv6 address?

andyxning:
@ploxiln I have investigated this further. It turns out that something has changed after v0.3.8, which we use internally. I have tested your changes against master branch, it is ok. And i think we can merge this. :)

As for our cases, we need to just merge #1185 internally since it is based on v0.3.8. :)

ploxiln:
> It turns out that something has changed after v0.3.8, which we use internally.

Thanks explains the confusion :)
Thank you for identifying the problem and testing this!

andyxning:
I think we can merge this for now. But i have no merge right. 

BTW, what about the request about promoting myself to be a NSQ member?

ploxiln:
yeah, was just giving some more comment time before merge :)

It turns out I actually don't have sufficient privileges to add other members to the organization, I will ping some people about that.

andyxning:
> It turns out I actually don't have sufficient privileges to add other members to the organization, I will ping some people about that.

Thanks for your help @ploxiln . Any more details we can chat on email. :)

Issue #1157, opened 2019-05-07
WARNING: DATA RACE when running test.sh


MilenkoM:
I compiled from source nsq on Ubuntu 18.04
```
git clone https://github.com/nsqio/nsq $GOPATH/src/github.com/nsqio/nsq
cd $GOPATH/src/github.com/nsqio/nsq
dep ensure
dep: WARNING: revision "a3a6125de932" should not be in abbreviated form
dep: WARNING: revision "b1c65ca7ae62" should not be in abbreviated form
dep: WARNING: revision "d9a9656a3a4b" should not be in abbreviated form
dep: WARNING: revision "2e65f85255db" should not be in abbreviated form
dep: WARNING: revision "0c63f026bcd6" should not be in abbreviated form
dep: WARNING: revision "74cfbc9de839" should not be in abbreviated form
```

When I run tests.sh
```
WARNING: DATA RACE
Read at 0x00c00043c343 by goroutine 47:
  testing.(*common).logDepth()
      /usr/local/go/src/testing/testing.go:621 +0x94
  testing.(*common).Log()
      /usr/local/go/src/testing/testing.go:614 +0x78
  testing.(*T).Log()
      <autogenerated>:1 +0x61
  github.com/nsqio/nsq/internal/test.(*testLogger).Output()
      /home/miki/go/src/github.com/nsqio/nsq/internal/test/logger.go:16 +0xd6
  github.com/nsqio/nsq/internal/lg.Logf()
      /home/miki/go/src/github.com/nsqio/nsq/internal/lg/lg.go:80 +0x122
  github.com/nsqio/nsq/nsqlookupd.(*LookupProtocolV1).IOLoop()
      /home/miki/go/src/github.com/nsqio/nsq/nsqlookupd/logger.go:18 +0x991
  github.com/nsqio/nsq/nsqlookupd.(*tcpServer).Handle()
      /home/miki/go/src/github.com/nsqio/nsq/nsqlookupd/tcp.go:44 +0xa83

Previous write at 0x00c00043c343 by goroutine 89:
  testing.tRunner.func1()
      /usr/local/go/src/testing/testing.go:856 +0x354
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:869 +0x17f

Goroutine 47 (running) created at:
  github.com/nsqio/nsq/internal/protocol.TCPServer()
      /home/miki/go/src/github.com/nsqio/nsq/internal/protocol/tcp_server.go:33 +0x521
[...]

PR #1190, opened 2019-09-18
nsqlookupd: synchronize goroutines to avoid t.Log data races

benjsto:
Fixes the data race seen here: https://github.com/nsqio/nsq/issues/1157 by providing a mechanism for all handler goroutines in `nsqlookupd` to exit before tests finish.

I'm obviously new here so let me know if there are better/simpler ways to accomplish this.

### DIFF
--- a/internal/protocol/tcp_server.go
+++ b/internal/protocol/tcp_server.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/nsqio/nsq/internal/lg"
 )
@@ -15,6 +16,8 @@ type TCPHandler interface {
 
 func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) error {
 	logf(lg.INFO, "TCP: listening on %s", listener.Addr())
+
+	var wg sync.WaitGroup
 
 	for {
 		clientConn, err := listener.Accept()
@@ -30,8 +33,16 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) er
 			}
 			break
 		}
-		go handler.Handle(clientConn)
+
+		wg.Add(1)
+		go func() {
+			handler.Handle(clientConn)
+			wg.Done()
+		}()
 	}
+
+	// wait to return until all handler goroutines complete
+	wg.Wait()
 
 	logf(lg.INFO, "TCP: closing %s", listener.Addr())
 
--- a/nsqlookupd/tcp.go
+++ b/nsqlookupd/tcp.go
@@ -3,12 +3,14 @@ package nsqlookupd
 import (
 	"io"
 	"net"
+	"sync"
 
 	"github.com/nsqio/nsq/internal/protocol"
 )
 
 type tcpServer struct {
-	ctx *Context
+	ctx   *Context
+	conns sync.Map
 }
 
 func (p *tcpServer) Handle(clientConn net.Conn) {
@@ -41,9 +43,19 @@ func (p *tcpServer) Handle(clientConn net.Conn) {
 		return
 	}
 
+	p.conns.Store(clientConn.RemoteAddr(), clientConn)
+
 	err = prot.IOLoop(clientConn)
 	if err != nil {
 		p.ctx.nsqlookupd.logf(LOG_ERROR, "client(%s) - %s", clientConn.RemoteAddr(), err)
-		return
 	}
+
+	p.conns.Delete(clientConn.RemoteAddr())
+}
+
+func (p *tcpServer) CloseAll() {
+	p.conns.Range(func(k, v interface{}) bool {
+		v.(net.Conn).Close()
+		return true
+	})
 }
--- a/nsqlookupd/lookup_protocol_v1_test.go
+++ b/nsqlookupd/lookup_protocol_v1_test.go
@@ -40,6 +40,8 @@ func testIOLoopReturnsClientErr(t *testing.T, fakeConn test.FakeNetConn) {
 	test.Nil(t, err)
 	prot := &LookupProtocolV1{ctx: &Context{nsqlookupd: nsqlookupd}}
 
+	nsqlookupd.tcpServer = &tcpServer{ctx: prot.ctx}
+
 	errChan := make(chan error)
 	testIOLoop := func() {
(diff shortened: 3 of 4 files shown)


nsqlookupd/lookup_protocol_v1.go
```diff
@@ -27,9 +28,30 @@ func (p *LookupProtocolV1) IOLoop(conn net.Conn) error {
...
 	for {
-		line, err = reader.ReadString('\n')
-		if err != nil {
+		exitLoop := false
+
+		// do this read in a goroutine so we can exit this loop if needed
+		go func() {
+			line, e := reader.ReadString('\n')
+			readChan <- line
+			errChan <- e
```
ploxiln: I think it would be better to instead find a way to call `conn.Close()` when `exitChan` indicates. Here's one idea:

```go
doneChan := make(chan interface{})
defer close(doneChan)
go func() {
        select {
        case <-p.exitChan:
                conn.Close()
        case <-doneChan:
                break
        }
}
for {
        line, err = reader.ReadString('\n')
        ...
```

Here's a different way that does not involve an extra goroutine and channel for each connection:

```diff
--- a/nsqlookupd/nsqlookupd.go
+++ b/nsqlookupd/nsqlookupd.go
@@ -66,6 +66,7 @@ func (l *NSQLookupd) Main() error {
        tcpServer := &tcpServer{ctx: ctx}
        l.waitGroup.Wrap(func() {
                exitFunc(protocol.TCPServer(l.tcpListener, tcpServer, l.logf))
+               tcpServer.CloseAll()
        })
        httpServer := newHTTPServer(ctx)
        l.waitGroup.Wrap(func() {

--- a/nsqlookupd/tcp.go
+++ b/nsqlookupd/tcp.go
@@ -3,12 +3,14 @@ package nsqlookupd
 import (
        "io"
        "net"
+       "sync"
 
        "github.com/nsqio/nsq/internal/protocol"
 )
 
 type tcpServer struct {
        ctx *Context
+       conns sync.Map
 }
 
 func (p *tcpServer) Handle(clientConn net.Conn) {
@@ -41,9 +43,20 @@ func (p *tcpServer) Handle(clientConn net.Conn) {
                return
        }
 
+       p.conns.Store(clientConn.RemoteAddr(), clientConn)
+
        err = prot.IOLoop(clientConn)
        if err != nil {
                p.ctx.nsqlookupd.logf(LOG_ERROR, "client(%s) - %s", clientConn.RemoteAddr(), err)
-               return
        }
+
+       p.conns.Delete(clientConn.RemoteAddr())
+}
+
+func (p *tcpServer) CloseAll() {
+       p.conns.Range(func (k, v interface{}) bool {
+               v.(net.Conn).Close()
+               p.conns.Delete(k)
+               return true
+       })
 }
```

but I'm not sure if the overhead of the `sync.Map()` is something to worry about. probably not :)

ploxiln: hmm, I think it would need a different way to call `tcpServer.CloseAll()` for my second proposal, ordering after `protocol.TcpServer()` returns seems like it shouldn't work

internal/protocol/tcp_server.go
```diff
@@ -30,9 +33,18 @@ func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) er
 			}
 			break
 		}
-		go handler.Handle(clientConn)
+
+		wg.Add(1)
+
+		go func() {
+			defer wg.Done()
```
ploxiln: minor nit: you do not need `defer` because there are not multiple return paths from this function. So it's a tiny bit simpler and a tiny bit more efficient to just put this statement second, without `defer`.

nsqlookupd/lookup_protocol_v1.go
ploxiln: a better way to call `CloseAll()` for that idea:

```diff
--- a/nsqlookupd/nsqlookupd.go
+++ b/nsqlookupd/nsqlookupd.go
@@ -18,6 +18,7 @@ type NSQLookupd struct {
        opts         *Options
        tcpListener  net.Listener
        httpListener net.Listener
+       tcpServer    *tcpServer
        waitGroup    util.WaitGroupWrapper
        DB           *RegistrationDB
 }
@@ -63,9 +64,9 @@ func (l *NSQLookupd) Main() error {
                })
        }
 
-       tcpServer := &tcpServer{ctx: ctx}
+       l.tcpServer = &tcpServer{ctx: ctx}
        l.waitGroup.Wrap(func() {
-               exitFunc(protocol.TCPServer(l.tcpListener, tcpServer, l.logf))
+               exitFunc(protocol.TCPServer(l.tcpListener, l.tcpServer, l.logf))
        })
        httpServer := newHTTPServer(ctx)
@@ -87,6 +88,7 @@ func (l *NSQLookupd) RealHTTPAddr() *net.TCPAddr {
 func (l *NSQLookupd) Exit() {
        if l.tcpListener != nil {
                l.tcpListener.Close()
+               l.tcpServer.CloseAll()
        }
```

benjsto: Trying that now, I like it. Thank you for taking a look at this!

ploxiln: thanks for the updates! I'm going to take a day or two to test it a bit myself

nsqlookupd/tcp.go
```diff
@@ -41,9 +43,20 @@ func (p *tcpServer) Handle(clientConn net.Conn) {
...
-		return
 	}
+
+	p.conns.Delete(clientConn.RemoteAddr())
+}
+
+func (p *tcpServer) CloseAll() {
+	p.conns.Range(func(k, v interface{}) bool {
+		v.(net.Conn).Close()
+		p.conns.Delete(k)
```
ploxiln: I think this `Delete()` is not really needed here, it'll be done at the end of `Handle()` anyway.

ploxiln:
worth noting: nsqd also uses `TCPServer`, and already closes all connections via `Topic.Close()` -> `Channel.Close()` -> `Consumer.Close()`

I think this is good. Please squash commits together before merge.

benjsto:
I assumed something like that was going on for nsqd's use of TCPServer but hadn't dug into it. Thanks for clarifying. Thanks for your help on this - I've removed that unnecessary Delete and squashed and rebased.

PR #1192, opened 2019-09-23
Support configurable Auth failure behavior.

blakesmith:
Hi NSQ friends!

We recently had a production issue that led to some data loss. This was caused by a service failure in our load balanced auth service, that's configured on all our nsqd instances in our topology.

In some of our systems, we'd like to favor availability of nsqd access in cases where the auth service is unresponsive. This is usually in scenarios where nsqd being unavailable "globally" would lead to dropping messages on the floor (IE - if we can't get the messages reliably to at least *one* running nsqd instead, since they all depend on one auth server, and the messages are gone forever). Since auth is decoupled from the running nsqd instance, we'd like to give a little more wiggle room to the failure behavior, if / when our auth server has problems again. We'd prefer this failure behavior, as opposed to disabling auth completely on nsqd instances that need availability.

This change adds a top an nsqd `auth-fail-closed` option, which adds the ability to fail closed in cases where the auth server is down, while still defaulting to the current behavior of failing open (if the auth server is down, don't allow any clients access). I added several supporting option flags to configure what values will be returned in a fail closed scenario.

Is this behavior something you all are interested in accepting into NSQ? Would you prefer to see this work a different way? If not, we might explore:

1. Putting some other journal / write ahead log in front of NSQ to prevent any central points of failure around nsq auth.
2. Keeping this patch-set maintained locally.

Thanks, and thanks for NSQ! We continue to be very happy with it.

Tests: Added integration tests, verified locally.

### DIFF
--- a/internal/auth/authorizations.go
+++ b/internal/auth/authorizations.go
@@ -26,6 +26,16 @@ type State struct {
 	Expires        time.Time
 }
 
+type AuthConfig struct {
+	ConnectTimeout time.Duration
+	RequestTimeout time.Duration
+
+	FailClosed             bool
+	FailDefaultTTL         int
+	FailDefaultIdentity    string
+	FailDefaultIdentityURL string
+}
+
 func (a *Authorization) HasPermission(permission string) bool {
 	for _, p := range a.Permissions {
 		if permission == p {
@@ -78,19 +88,38 @@ func (a *State) IsExpired() bool {
 }
 
 func QueryAnyAuthd(authd []string, remoteIP string, tlsEnabled bool, commonName string, authSecret string,
-	connectTimeout time.Duration, requestTimeout time.Duration) (*State, error) {
+	authConfig *AuthConfig) (*State, error) {
 	start := rand.Int()
 	n := len(authd)
 	for i := 0; i < n; i++ {
 		a := authd[(i+start)%n]
-		authState, err := QueryAuthd(a, remoteIP, tlsEnabled, commonName, authSecret, connectTimeout, requestTimeout)
+		authState, err := QueryAuthd(a, remoteIP, tlsEnabled, commonName, authSecret, authConfig.ConnectTimeout, authConfig.RequestTimeout)
 		if err != nil {
 			log.Printf("Error: failed auth against %s %s", a, err)
 			continue
 		}
 		return authState, nil
 	}
-	return nil, errors.New("Unable to access auth server")
+
+	if authConfig.FailClosed {
+		log.Printf("Unable to access an auth server, failing closed and allowing all authorizations")
+		closedAuthorizations := []Authorization{
+			{
+				Topic:       ".*",
+				Channels:    []string{".*"},
+				Permissions: []string{"publish", "subscribe"},
+			},
+		}
+		return &State{
+			TTL:            authConfig.FailDefaultTTL,
+			Authorizations: closedAuthorizations,
+			Identity:       authConfig.FailDefaultIdentity,
+			IdentityURL:    authConfig.FailDefaultIdentityURL,
+			Expires:        time.Now().Add(time.Duration(authConfig.FailDefaultTTL) * time.Second),
+		}, nil
+	} else {
+		return nil, errors.New("Unable to access auth server")
+	}
 }
 
 func QueryAuthd(authd string, remoteIP string, tlsEnabled bool, commonName string, authSecret string,
--- a/apps/nsqd/options.go
+++ b/apps/nsqd/options.go
@@ -113,6 +113,10 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.String("tcp-address", opts.TCPAddress, "<addr>:<port> to listen on for TCP clients")
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> to query auth server (may be given multiple times)")
+	flagSet.Bool("auth-fail-closed", opts.AuthFailClosed, "When unable to contact the auth server, fail closed and allow authorizations. (defaults to false)")
+	flagSet.Int("auth-fail-default-ttl", opts.AuthFailDefaultTTL, "When auth fails closed, define the default TTL to return to clients when unable to query the auth server. (defaults to 360)")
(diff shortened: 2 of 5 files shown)


internal/auth/authorizations.go
```diff
@@ -78,19 +88,38 @@ func (a *State) IsExpired() bool {
...
 		if err != nil {
 			log.Printf("Error: failed auth against %s %s", a, err)
 			continue
 		}
 		return authState, nil
 	}
-	return nil, errors.New("Unable to access auth server")
+
+	if authConfig.FailClosed {
+		log.Printf("Unable to access an auth server, failing closed and allowing all authorizations")
```
stantonk: i'm not sure how the logging semantics work for nsq, but showing this as a warning or error might be good?

ploxiln:
I have the opposite idea of the meaning of "fail-open" and "fail-closed". I think that "fail-closed" means that if something goes wrong, access is blocked, like a closed door or a closed port (not like a closed circuit).

internal/auth/authorizations.go
ploxiln: nsqd has a special logging function with levels. For example, in `nsqd/nsqd.go`, it calls:

```go
			n.logf(LOG_WARN, "skipping creation of invalid topic %s", t.Name)
```

Internal modules do not have access to the NSQD struct, so for some of them, it passes the log function when it initializes the internal module, for example:

```go
		exitFunc(protocol.TCPServer(n.tcpListener, tcpServer, n.logf))
```

```go
func TCPServer(listener net.Listener, handler TCPHandler, logf lg.AppLogFunc) error {
	logf(lg.INFO, "TCP: listening on %s", listener.Addr())
        ...
```

This internal/authorizations module has thus far not had the log function passed because it only returned errors which would be logged by another module.

jehiah:
@blakesmith thank you for your context in the PR about this use case and the challenges you have run into.

I'm curious - have you considered trying to implement your desire for "fail open" as a local process that you can run along side nsqd that wraps your auth service. That process can return your desired "fail open" response if it encounters an error connecting to your auth service, encounters an upstream timeout, etc. This might be away to implement your logic outside of nsqd itself (which is always desirable).

nsqd is also setup to have multiple configured auth endpoints to support availability - i'm also curious if you are using that and that's been helpful?

blakesmith:
@ploxiln Ah, I was going more for the electrical circuit closed metaphor. Even reading the [wikipedia](https://en.wikipedia.org/wiki/Fail-safe) entry isn't always clear. :-)

@jehiah Running another "fail safe" nsq auth was another option we discussed, with the tradeoffs:

- PRO: Logic is outside of NSQ itself.
- CON: Another moving piece that can fail.

Also, if I'm reading the `QueryAnyAuthd` code right, it seems like nsqd currently load balances among the configured auth backends (See: https://github.com/nsqio/nsq/blob/master/internal/auth/authorizations.go#L82), rather than having a guaranteed fallback order. I think we'd need to have our "central" authd always be the first attempted service, and our "failsafe" authd always be the second attempt for this strategy to work. @jehiah Does this sound right, or is there a different auth feature you had in mind?

I like the idea of running a tiny side-car "failsafe" auth backend, if we could get guaranteed auth order stability. What do you all think about having an nsqd command line flag that drives the ordering of authd querying? Something like:

```
--auth-query-ordered: When multiple auth servers are specified, whether they should be queried in order, or randomized for load balancing behavior. Default: false
```

Thanks for the feedback!

blakesmith:
Quoting @jehiah:

> I'm curious - have you considered trying to implement your desire for "fail open" as a local process that you can run along side nsqd that wraps your auth service. That process can return your desired "fail open" response if it encounters an error connecting to your auth service, encounters an upstream timeout, etc. This might be away to implement your logic outside of nsqd itself (which is always desirable).

I hadn't really thought about that. 🤔 That would allow us to implement the desired behavior without any NSQd changes. Which makes more sense to you, providing auth ordering stability, or creating some kind of wrapper like this?

jehiah:
@blakesmith If I was in your shoes i'd probably go the proxy route, *but* I think a nsqd flag that switches the authd queries from "load balanced" mode to "priority order" would also be a reasonable change to land. I'd much prefer that as a nsqd addition than the more intricate logic around what auth state to use when the authd query is unavailable.

blakesmith:
> I think a nsqd flag that switches the authd queries from "load balanced" mode to "priority order" would also be a reasonable change to land.

@jehiah Cool. Let me put together a patch for switching load balanced / priority order, and we can see how it looks? Any inputs on a flag name? `--auth-load-balanced`, defaulting to true?, `--auth-priority-order`, defaulting to false?

>  I'd much prefer that as a nsqd addition than the more intricate logic around what auth state to use when the authd query is unavailable.

For sure! We kind of suspected that it might not feel 100% clean to put this logic in here, but wanted to take a crack and see how gross the change would make us feel. If nothing else, having code to look at helps start the conversation. :-)

jehiah:
Let's model naming after the `nsq_to_nsq` `-mode` option.

```
-auth-query-mode
    the auth server request mode options: 'round-robin' (default), 'priority-order'
```

cc: @ploxiln @mreiferson in case either of you have particular thoughts on this approach.

ploxiln:
That proposal looks good to me @jehiah. 

I initially commented on a couple of  technical details, because that was easier :) while I was thinking about how to express that we would probably not accept the initially proposed feature. And then Jehiah jumped in, and said it better than I would have.

Another thought I had was the option name `--auth-ordered`, default false to keep current behavior, but also leaving room to improve the current  behavior to be a smarter balanced host-pool kind of thing, if anyone ends up needing that. But `--auth-query-mode` looks good, maybe even more clear with named options.

blakesmith:
To bring this full circle, we've had some more discussions internally, and are either planning on either:

1. Implementing the local nsqauth wrapper service that proxies to our central auth service, and whatever other custom logic we need.
2. Putting a different write-ahead-log style journal in front of NSQ to account for cases when unable to publish to nsqd due to auth problems.

The pain we experienced is self-inflicted by the way we've deploy nsq-auth, so I think it makes more sense to fix this outside of nsq as much as possible.

Thanks everyone for the great discussion on this issue. NSQ continues to be awesome, and we keep pushing it harder and harder! (in a good way!)

Cheers,

PR #1204, opened 2019-10-25
*: update build process and ssh in bench.py

### LINKED ISSUE #1202: nsqd: ec2 benchmark is broken
bench/bench.py seems to follow the old build pipeline

martingrambow:
There were some small errors in bench.py (fixes #1202):

- ssh_connect_with_retries() did not ask for the ssh key
- I also added some logging in ssh_cmd()
- bootstrap(addr) followed the old build process
- run() called the async ssh command, however, sometimes the pkill killed the deamon immediately; so I switched to the usual ssh_cmd and to screen

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -79,16 +79,16 @@ def _bootstrap(addr):
             'sudo -S tar -C /usr/local -xzf go%s.linux-amd64.tar.gz' % golang_version,
             'sudo -S apt-get update',
             'sudo -S apt-get -y install git mercurial',
-			'sudo -S apt-get -y install golang',
-			'curl https://raw.githubusercontent.com/golang/dep/master/install.sh > installdep.sh',
-			'chmod +x installdep.sh',
-			'sudo GOPATH=/usr/local/go ./installdep.sh',
+            'sudo -S apt-get -y install golang',
+            'curl https://raw.githubusercontent.com/golang/dep/master/install.sh > installdep.sh',
+            'chmod +x installdep.sh',
+            'sudo GOPATH=/usr/local/go ./installdep.sh',
             'mkdir -p go/src/github.com/nsqio',
             'cd go/src/github.com/nsqio && git clone https://github.com/nsqio/nsq',
             'cd go/src/github.com/nsqio/nsq && git checkout %s' % commit,
-			'cd go/src/github.com/nsqio/nsq && GOPATH=/home/ubuntu/go /usr/local/go/bin/dep ensure',
-			'cd go/src/github.com/nsqio/nsq && make',
-			'cd go/src/github.com/nsqio/nsq/bench/bench_writer && GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
+            'cd go/src/github.com/nsqio/nsq && GOPATH=/home/ubuntu/go /usr/local/go/bin/dep ensure',
+            'cd go/src/github.com/nsqio/nsq && make',
+            'cd go/src/github.com/nsqio/nsq/bench/bench_writer && GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
             'cd go/src/github.com/nsqio/nsq/bench/bench_reader && GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
             'sudo -S mkdir -p /mnt/nsq',
             'sudo -S chmod 777 /mnt/nsq']:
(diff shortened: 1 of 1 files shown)


ploxiln:
Clearly we have not maintained this script in the last couple years. Thanks.

bench/bench.py
```diff
@@ -76,20 +79,17 @@ def _bootstrap(addr):
...
-            'cd go/src/github.com/nsqio/nsq/apps/nsqd && \
-                GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
-            'cd go/src/github.com/nsqio/nsq/bench/bench_writer && \
-                GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
-            'cd go/src/github.com/nsqio/nsq/bench/bench_reader && \
-                GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
+			'cd go/src/github.com/nsqio/nsq && GOPATH=/home/ubuntu/go /usr/local/go/bin/dep ensure',
+			'cd go/src/github.com/nsqio/nsq && make',
+			'cd go/src/github.com/nsqio/nsq/bench/bench_writer && GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
+            'cd go/src/github.com/nsqio/nsq/bench/bench_reader && GOPATH=/home/ubuntu/go /usr/local/go/bin/go build',
```
ploxiln: The inconsistent indenting shown here is because you're using tabs for some lines. Because this is python, 4-space indent everywhere is the way to go.

bench/bench.py
```diff
@@ -147,11 +147,9 @@ def run():
             for cmd in [
                     'sudo -S pkill -f nsqd',
                     'sudo -S rm -f /mnt/nsq/*.dat',
-                    'GOMAXPROCS=32 ./go/src/github.com/nsqio/nsq/apps/nsqd/nsqd \
-                        --data-path=/mnt/nsq --mem-queue-size=10000000 --max-rdy-count=%s' % (
-                        tornado.options.options.rdy
-                        )]:
-                nsqd_chans.append((ssh_client, ssh_cmd_async(ssh_client, cmd)))
+					'echo starting nsq &>> nsq.log',
+                    'GOMAXPROCS=32 screen -mdS "nsq" ./go/src/github.com/nsqio/nsq/build/nsqd --data-path=/mnt/nsq --mem-queue-size=10000000 --max-rdy-count=%s &>> nsq.log' % (tornado.options.options.rdy)]:
```
ploxiln: While we have no explicit formatting rules for python helper scripts in this repo, this line is a bit too long. Please wrap somewhere around 100 chars.

Here's a trick: in python, if there are two string literals next to each other, they are concatenated by the python parsing step. Something like this should work:

```python
            for cmd in [
                'sudo -S pkill -f nsqd',
                'sudo -S rm -f /mnt/nsq/*.dat',
                'echo starting nsq &>> nsq.log',
                'GOMAXPROCS=32 screen -mdS "nsq" ./go/src/github.com/nsqio/nsq/build/nsqd'
                ' --data-path=/mnt/nsq --mem-queue-size=10000000'
                ' --max-rdy-count=%s &>> nsq.log' % tornado.options.options.rdy,
            ]:
```

(admittedly a bit tricky, since a missing or present comma makes a big difference, silently)

ploxiln:
while you're here, the defaults for the `instance_type` and `ami` options should probably be updated

bench/bench.py
```diff
@@ -287,11 +289,11 @@ def decomm():
                            help='AWS account access key')
     tornado.options.define('secret_key', type=str, default='',
                            help='AWS account secret key')
-    tornado.options.define('ami', type=str, default='ami-48fd2120',
+    tornado.options.define('ami', type=str, default='ami-02df9ea15c1778c9c',
```
ploxiln: This ami does not seem to exist in the default region, us-east-1 (and I also checked a couple of other common regions and could not find it). Can you find an ami from the default region?

```
$ aws --region us-east-1 ec2 describe-images --image-ids ami-48fd2120
{
    "Images": [
        {
            "Architecture": "x86_64",
            "CreationDate": "2014-08-16T21:40:48.000Z",
            "ImageId": "ami-48fd2120",
            "ImageLocation": "ubuntu-us-east-1/images/hvm-instance/ubuntu-trusty-14.04-amd64-server-20140816.manifest.xml",
            "ImageType": "machine",
            "Public": true,
            "OwnerId": "099720109477",
            "State": "available",
            "BlockDeviceMappings": [],
            "Hypervisor": "xen",
            "Name": "ubuntu/images/hvm-instance/ubuntu-trusty-14.04-amd64-server-20140816",
            "RootDeviceType": "instance-store",
            "SriovNetSupport": "simple",
            "VirtualizationType": "hvm"
        }
    ]
}
$ aws --region us-east-1 ec2 describe-images --image-ids ami-02df9ea15c1778c9c

An error occurred (InvalidAMIID.NotFound) when calling the DescribeImages operation: The image id '[ami-02df9ea15c1778c9c]' does not exist
```

(I guess we should update to ubuntu-16.04 or ubuntu-18.04, but you probably did that, I just can't tell because I can't find the ami ...)

bench/bench.py
```diff
@@ -287,11 +289,11 @@ def decomm():
...
                            help='AMI ID')
     tornado.options.define('ssh_key_name', type=str, default='default',
                            help='SSH key name')
-    tornado.options.define('instance_type', type=str, default='c3.2xlarge',
+    tornado.options.define('instance_type', type=str, default='t2.medium',
```
ploxiln: I think c4.xlarge or c5.xlarge would be more in the spirit of the previous default. A bit expensive, but they'll run for less than an hour, and t2 or t3 instances are probably not good for benchmarking because they are more oversubscribed and throttled.

bench/bench.py
```diff
@@ -301,19 +303,21 @@ def decomm():
...
-    }
-    cmd = cmd_map.get(cmd_name, bootstrap)
-
+    logging.info('CMD: %s', cmd_name)
+    cmd = bootstrap
+    if (cmd_name == 'bootstrap'):
+        cmd = bootstrap
+    if (cmd_name == 'run'):
+        cmd = run
+    if (cmd_name == 'decomm'):
```
ploxiln: This part did not have to change. The restructure is OK with me, except that using parentheses like this is odd and unusual for python (they're not needed).

mreiferson:
see #1264

PR #1206, opened 2019-10-27
Replace sort.Sort with sort.Ints

lelenanam:
Fix staticcheck issue


ploxiln:
I see the same pattern in `nsqd/in_flight_pqueue_test.go`, want to take care of that one too?

lelenanam:
Sure, I fixed `sort.Sort` in the test.
Not sure why it fails now, looks unrelated. Maybe let's just retry?

ploxiln:
Yeah, was a spurious/flaky failure, unfortunately, in 1 of the 6 test configs. I restarted it and it indeed passed.

PR #1207, opened 2019-10-30
Add JSON destination topic partitioning to nsq_to_nsq

seattlecarl:
I wanted to distribute messages over several systems as a way to horizontally scale work without changing producers or consumers and it looks like I can leverage the existing JSON work to partition messages into subtopics. (This allows me to "shard" my consumers.)

Is this a vailid approach? It seems to work in my local testing, but I'm new to go.

### DIFF
--- a/apps/nsq_to_nsq/nsq_to_nsq.go
+++ b/apps/nsq_to_nsq/nsq_to_nsq.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"hash/fnv"
 	"log"
 	"os"
 	"os/signal"
@@ -30,6 +31,7 @@ const (
 )
 
 var (
+	// single value params
 	showVersion = flag.Bool("version", false, "print version string")
 	channel     = flag.String("channel", "nsq_to_nsq", "nsq channel")
 	destTopic   = flag.String("destination-topic", "", "use this destination topic for all consumed topics (default is consumed topic name)")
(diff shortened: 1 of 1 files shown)


ploxiln:
We usually do not "shard" into multiple topics, because that is what channels are for:

f1434dc8-6029-11e3-8a66-18ca4ea10aca

It is very reasonable to copy one of the sample consumer apps and modify it into exactly what you need. But I do not think we want to add this particular functionality to `nsq_to_nsq`.

ploxiln:
eh let me clarify ... look at the multiple consumers for a single channel, the topic messages are spread among them

seattlecarl:
The reason for the "sharding" is to get "stickiness" with a correlation key landing on the same topic (which itself has multiple consumers and possibly multiple channels).

As I understand, it, channels are greedy and will evenly distribute a copy individual messages to the channel's consumers, but I needed a feature to match "start" and "stop" events on the same group of consumers using a correlation ID.

Yes, I could have modified the producer to be shard-aware, but this is a generic change where I can use it for all producers and it greatly simplifies my deployment since I can just pause the source queue if I need to rebalance the sharding.

Anyway, that correlation key stickiness was the reason for the change. If there's a better way to do it, let me know.

This seemed to solve my problem, and I can keep it in a private repo if it's not a good fit or work on a more generic tool. I wanted to do a plugin system for message filtering, but OS compat scared me off.

Thanks.

ploxiln:
I see, that use case does make some sense. Note that delivery order is not guaranteed, and a message in the middle of a related group may be timed-out and then queued behind a bunch of unrelated messages ... there are some complications, but maybe that is all OK for your system.

Your json key sharding nsq relay could be in your own *public* repo, and others may find it useful. I consider "a plugin system for message filtering" to not really be easier than writing a few lines of golang (or python) to filter as needed in a custom service. We write lots of custom nsq consumers. As you suspected, figuring out plugins would be a bit complicated, but then there's also all the options and modes and variants that various people will ask for, and wiring them all up, and documenting them ...

seattlecarl:
Yeah, I'll close this, then, and in my repo I'll give the tool a different name. The existing JSON filtering made it easy for a go newbie to make it work. For historical reference, here is an example of 10 start/stop events matched over 5 shards. The ordering considerations are not different for me than what I've got now since I need both events to proceed and I have multiple consumers sharing a channel. My consumers scale fine but my stack is going to hit a scaling cap on a shared datastore at some point. Sharding makes it look like I can solve my problem and get N-way scaling by splitting the load to full copies of my ecosystem (each with an independent datastore) with very little code change - my end-user query then can work like a map-reduce and hit each copy in parallel and combine results.

```
nsqd v1.1.0 (built w/go1.10.3)
start_time 2019-07-02T18:26:29Z
uptime 2882h1m8.46557115s
...
   [sharded_test_acc_re.0] depth: 2     be-depth: 0     msgs: 2        e2e%:

   [sharded_test_acc_re.1] depth: 2     be-depth: 0     msgs: 2        e2e%:

   [sharded_test_acc_re.2] depth: 2     be-depth: 0     msgs: 2        e2e%:

   [sharded_test_acc_re.3] depth: 2     be-depth: 0     msgs: 2        e2e%:

   [sharded_test_acc_re.4] depth: 2     be-depth: 0     msgs: 2        e2e%:

   [test_acc       ] depth: 0     be-depth: 0     msgs: 5207057  e2e%:
      [nsq_to_nsq               ] depth: 0     be-depth: 0     inflt: 0    def: 0    re-q: 0     timeout: 0     msgs: 10       e2e%:
```

Issue #1217, opened 2019-11-22
failed to instantiate nsqd - --data-path xxx possibly by another instance of nsqd


coanor:
- docker version: Docker version 18.06.3-ce, build d7080c1
- docker-compose version:

    docker-compose version 1.17.1, build unknown
    docker-py version: 2.5.1
    CPython version: 2.7.15+
    OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
- host info:  Ubuntu 18.04.3 LTS (ubuntu server)

----

I'm using docker-compose to setup NSQ according to [here](https://nsq.io/deployment/docker.html#using-docker-compose).

Additionally, I have add `--data-path=/data/nsqd` to `nsqd` node, and I got

    FATAL: failed to instantiate nsqd - --data-path=/data/nsqd in use (possibly by another instance of nsqd)

According to [source code](https://github.com/nsqio/nsq/blob/master/nsqd/nsqd.go#L110), this may be caused by a `dirlock` opertion.

I've used `auditctl -w /data/nsqd -p war -k whatsgoingon` to monitor `/data/nsqd`, when someone touched (`war`: `write/access/read`) it, some log will print to `/var/log/audit/audit.log`.

I reastart the `nsqd`, but still got the same error (via `docker-compose logs`)

    FATAL: failed to instantiate nsqd - --data-path=/data/nsqd in use (possibly by another instance of nsqd)

but no message showing in `audit.log` (via `tailf /var/log/audit/audit.log | grep "whatsgoingon"`). this means that during starting the `nsqd`, no one touched `/data/nsqd`.

I scraped the [ dir lock source code](https://github.com/nsqio/nsq/blob/master/internal/dirlock/dirlock.go#L22) and build a test program, when `os.Open("/data/nsqd")`, the monitor shows the touch action as expected.

I wonder why nsq think someone touching the `--data-path` but acturally no one there, and this maybe a bug.

ploxiln:
I don't use docker compose, can't help much with that.

But I wonder if the filesystem type underlying the nsqd storage volume could be the problem?

coanor:
> I don't use docker compose, can't help much with that.
> 
> But I wonder if the filesystem type underlying the nsqd storage volume could be the problem?

coanor:
@ploxiln I've tried in raw docker install, but still failed with the same error. BTW, my fs type is ext4 (via `df -T`).

mdh67899:
[...]

PR #1220, opened 2019-12-02
fix the log when data-path is not exist

mdh67899:
https://github.com/nsqio/nsq/issues/1217

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -108,7 +108,7 @@ func New(opts *Options) (*NSQD, error) {
 
 	err = n.dl.Lock()
 	if err != nil {
-		return nil, fmt.Errorf("--data-path=%s in use (possibly by another instance of nsqd)", dataPath)
+		return nil, fmt.Errorf("failed to lock data-path: %v", err)
 	}
 
 	if opts.MaxDeflateLevel < 1 || opts.MaxDeflateLevel > 9 {
--- a/internal/dirlock/dirlock.go
+++ b/internal/dirlock/dirlock.go
@@ -27,7 +27,7 @@ func (l *DirLock) Lock() error {
 	l.f = f
 	err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
 	if err != nil {
-		return fmt.Errorf("cannot flock directory %s - %s", l.dir, err)
+		return fmt.Errorf("cannot flock directory %s - %s (possibly in use by another instance of nsqd)", l.dir, err)
 	}
 	return nil
 }


nsqd/nsqd.go
```diff
@@ -108,7 +108,7 @@ func New(opts *Options) (*NSQD, error) {
 
 	err = n.dl.Lock()
 	if err != nil {
-		return nil, fmt.Errorf("--data-path=%s in use (possibly by another instance of nsqd)", dataPath)
+		return nil, fmt.Errorf("trying lock --data-path=%s failed: %v (possibly by another instance of nsqd)", dataPath, err)
```
jehiah: @mdh67899 what is the specific error you are getting? If you can share that it might help to better understand how to structure this error to handle the normal case (Another nsqd is running) and the case it sounds like you encountered (filesystem support)

mdh67899: @jehiah If the directory specified by ```--data-path``` does not exist, then nsqd startup will fail when trying to execute ```n.dl.Lock ()```.

In addition, if there are other nsqd processes locking the ```--data-path``` , an error will also be reported
https://github.com/nsqio/nsq/blob/5b67f580f3cbb63c02c4678e76896f5807147145/internal/dirlock/dirlock.go#L22-L33

If the ```n.dl.Lock ()``` operation fails, the error should be printed in the log for troubleshooting.
or do you think it is a good idea to create data path when ```--data-path``` is not exist?

mdh67899: @jehiah in issue #1217, it's not a filesystem support problem, the reason is the ```--data-path=/data/nsq``` is no exist in docker container

ploxiln: I think it is a good idea to add the underlying error text here. I think it could be worded in a simpler way, maybe just `"failed to lock data-path: %v"`, and we may not need to include the path explicitly, nor the "(possibly by another instance ..." part . The main error possibilities are (according to my testing just now with this branch and go-1.13 on macOS):
  * "cannot flock directory ./tmp - resource temporarily unavailable"
  * "open ./tmp2: no such file or directory"
  * "open ./tmp2: permission denied"

We could add the "possibly by another instance" explanation just for the flock error ...

jehiah: thanks for the context @ploxiln. I think it would be good to keep explicit guidance in the flock error case that nsqd doesn't support multiple instances running w/ the same data directory.

mdh67899: @jehiah @ploxiln Do you think it is appropriate to adjust the code this way?

nsq/internal/dirlock/dirlock.go
```
 func (l *DirLock) Lock() error { 
 	f, err := os.Open(l.dir) 
 	if err != nil { 
 		return fmt.Errorf("cannot open directory %s - %s", l.dir, err) 
 	} 
 	l.f = f 
 	err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB) 
 	if err != nil { 
 		return fmt.Errorf("cannot flock directory %s - %s (possibly by another instance of nsqd)", l.dir, err)
  	} 
 	return nil 
 } 
```

nsqd/nsqd.go

```
	err = n.dl.Lock()
	if err != nil {
		return nil, fmt.Errorf("failed to lock data-path: %v", err)
	}
```

ploxiln: Yeah, adding the "possibly by another instance" to the Flock error return works for me. With the other changes, I would rephrase it "(possibly in use by another instance of nsqd)" here.

I don't think the Open error needs any additions, the text seems to already include "open {dir}".

I tested briefly in a linux VM and saw the same error text from Open and Flock.

mdh67899:
@ploxiln PTAL

PR #1221, opened 2019-12-02
tests: update travis-ci go versions - up to 1.13

ploxiln:


ploxiln:
cc @jehiah @mreiferson

.travis.yml
```diff
@@ -3,6 +3,8 @@ go:
   - 1.9.x
   - 1.10.x
   - 1.11.x
+  - 1.12.x
```
jehiah: Do we want to switch to just support the last 2 releases here and in go-nsq? (i.e. 1.12.x and 1.13.x)

ploxiln: I like supporting old versions when it is easy. I was thinking about dropping the versions in the middle though (assuming that if 1.9 works and 1.12 works then 1.10 and 1.11 very likely are fine).

jehiah: 👍 Can you capture that as a comment in this file along the lines of "Tests are run against the two most recent Go versions, and a minimum supported version (when different)"

ploxiln: sounds good, done

PR #1226, opened 2019-12-23
nsqd: reduce ReadyCount update frequency

andyxning:
This PR will reduce the update frequency for `ReadyCount` when the new rdy count equals with the old one.

### DIFF
--- a/nsqd/client_v2.go
+++ b/nsqd/client_v2.go
@@ -327,8 +327,11 @@ func (c *clientV2) IsReadyForMessages() bool {
 }
 
 func (c *clientV2) SetReadyCount(count int64) {
-	atomic.StoreInt64(&c.ReadyCount, count)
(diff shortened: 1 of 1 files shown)


ploxiln:
This results in two atomic operations for the "change" case. It could be just one like:

```go
old_count = atomic.SwapInt64(&c.ReadyCount, count)
if old_count != count {
    c.tryUpdateReadyState()
}
```

(I suppose the "no change" case happens quite a lot for some of your services? if it's not too frequent, it's not expensive ...)

andyxning:
@ploxiln Thanks for your good advice. 

> I suppose the "no change" case happens quite a lot for some of your services? if it's not too frequent, it's not expensive ...

Yes. it happens a lot. But even if it is not, avoiding useless logic is good, right :)

ploxiln:
cool. thanks

PR #1237, opened 2020-02-20
Dockerfile: add nsswitch.conf so go resolver uses /etc/hosts first

ploxiln:
go upstream might fix this soon, but for now it is rather surprising
see https://github.com/golang/go/issues/35305


mreiferson:
Holy hell that issue is an interesting read, LGTM

PR #1239, opened 2020-02-25
ignore build folder

00arthur00:
Signed-off-by: yapo.yang <yang_yapo@126.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .godeps
 vendor
-./build
+build/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 .godeps
 vendor
-./build
+build/
```
ploxiln: This intended to be `/build`, which works. But this way works fine too.

PR #1242, opened 2020-03-20
internal/util: use defer function in waitgroup wrapper

tianhongw:
Use defer function  due to the possibility of panic occurs in `cb()` func.


ploxiln:
If `cb()` panics, then the whole process will crash/exit, because it is not recovered anywhere.

`panic()` is supposed to be used much more rarely than exceptions in other languages. nsq does not use panic+recover. A panic should never happen, but if it does, the whole process will exit and show backtraces for all goroutines.

tianhongw:
I got it, thanks for your reply！

Issue #1124, opened 2019-01-11
nsqd: stops send messages after moving time backward


vostrik:
**Environment/Pre-Conditions**
nsqd version: 1.1.0
Node.js application with nsqjs

**Steps to Reproduce:**

1. Run nsqd and Node.js application.
2. Generate some messages. Everything works correctly.
3. Move time backward (for ex.: 5 minutes or 1 hour).
4. Generate some messages. Subscriber does't receive messages.
5.1. After some time (may be delay from (3)?) message will be published.
5.2. If I move time forward then messages will be published immediately.

**Actual Result:**

Messages aren't sent to subscriber after moving time back.

**Expected Result:**

Messages are sent to subscriber after moving time back.

ploxiln:
I generally expect various things to not work right when system time moves backwards :)

Go-1.9 added "transparent monotonic time" in certain cases: https://golang.org/doc/go1.9#monotonic-time
There may be ways we can adjust nsq code to get the monotonic-time based comparisons and timeouts ... if nsqd is the problem here rather than the nodejs client library ...

vostrik:
I think It is certainly nsqd because nodejs restart didn’t help.

vostrik:
Can I help with this issue? Maybe you can prompt modules and methods names which should be refactored.

ploxiln:
thanks, we'll look into it

ploxiln:
since Go does not use separate fuctions/options/objects for monotonic time, but includes it hidden in the wall-clock time struct, it's probably a bit tricky to figure out where it's being lost, or how to keep it along where needed

mreiferson:
I'd be surprised if "normal" channel sends were the problem, so I'd start by looking at other timeout related things. The two that come to mind are:

1. network connection level timeouts
2. `time.Ticker` use cases, e.g. to flush buffered messages to a client

vostrik:
Sorry for bothering but is there any progress?
Our team has some resources to investigate this behaviour, i. e. you can simply point potential problem places.
We want to help.

ploxiln:
No progress.

I suspect the issue may be around https://github.com/nsqio/nsq/blob/master/nsqd/guid.go#L59

mreiferson:
> I suspect the issue may be around https://github.com/nsqio/nsq/blob/master/nsqd/guid.go#L59

Ahhh, yes, I completely forgot about the GUID code, good call!

[...]

PR #1249, opened 2020-05-06
nsqd: switch to monotonic clocks for ID generation

slayercat:
1. nsqd/guid.go: uses pseudo-random generate method. will not fail to wait-retry loop when time gone backwards
2. time diff: use time.Time for diff. with monotonic clock to measure elapsed time [1]
3. pqueue(s): use time.Time for heap / priority compare
4. fix tests.

See also:
[1] golang Monotonic Clocks: https://golang.org/pkg/time

### DIFF
--- a/internal/pqueue/pqueue.go
+++ b/internal/pqueue/pqueue.go
@@ -2,11 +2,12 @@ package pqueue
 
 import (
 	"container/heap"
+	"time"
 )
 
 type Item struct {
 	Value    interface{}
-	Priority int64
+	Priority time.Time
 	Index    int
 }
 
@@ -23,7 +24,7 @@ func (pq PriorityQueue) Len() int {
 }
 
 func (pq PriorityQueue) Less(i, j int) bool {
-	return pq[i].Priority < pq[j].Priority
+	return pq[i].Priority.Before(pq[j].Priority)
 }
 
 func (pq PriorityQueue) Swap(i, j int) {
@@ -60,14 +61,14 @@ func (pq *PriorityQueue) Pop() interface{} {
 	return item
 }
 
-func (pq *PriorityQueue) PeekAndShift(max int64) (*Item, int64) {
+func (pq *PriorityQueue) PeekAndShift(max time.Time) (*Item, time.Duration) {
 	if pq.Len() == 0 {
 		return nil, 0
 	}
 
 	item := (*pq)[0]
-	if item.Priority > max {
-		return nil, item.Priority - max
+	if item.Priority.After(max) {
+		return nil, item.Priority.Sub(max)
 	}
 	heap.Remove(pq, 0)
 
--- a/internal/pqueue/pqueue_test.go
+++ b/internal/pqueue/pqueue_test.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"sort"
 	"testing"
+	"time"
 )
 
 func equal(t *testing.T, act, exp interface{}) {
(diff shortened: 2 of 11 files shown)


slayercat:
# manually test:

## open a server
```
 ./nsqd -log-level 'debug'
```
## pub
```
sh -c 'while true; do uuidgen; done' |./to_nsq -topic 'asdf#ephemeral' --nsqd-tcp-address '127.0.0.1:4150'
```
## sub
``` 
./nsq_tail -topic 'asdf#ephemeral' --nsqd-tcp-address 127.0.0.1:4150
```
## change time
```
 date -s '2020-01-01 00:00:00'
```

# expected result
change server time will not  break the nsqd data stream.

ploxiln:
previously discussed in #1124 

Looks like good stuff. I'm not sure about the random ID algo ... but it might be time to break previous behavior and make the switch. I'll properly review eventually ...

slayercat:
# UUID generate.
A similar algorithm is UUIDv4 (rfc4122) generate alg. see  [this](https://github.com/satori/go.uuid/blob/b2ce2384e17bbe0c6d34077efa39dbab3e09123b/generator.go#L166) for an implement.

We may read [rand.Reader](https://golang.org/pkg/crypto/rand/) for a better seed either. But i think nanosecond mixs `nodeID` will be good enough to make unique ids (orderless of course).

# The nodeID
After a futher dig in this day, The `nodeID` is much smaller then initially thought...

The `nodeID` is defaultly assigned [here](https://github.com/nsqio/nsq/blob/6774510b92a4d39c00a9ba59f9a284c0a2b8ef7e/nsqd/options.go#L94). Could not understand why it's limited in [0,1024).  It will much easy to collision with other nodes in such a small range.

This may be better:
```
	h := md5.New()
	io.WriteString(h, hostname)
	md5hash := h.Sum(nil)
	var defaultID int64
	binary.Read(bytes.NewBuffer(md5hash), binary.LittleEndian, &defaultID)
```

# UUID generate improvement with such a small nodeID

- method 1, turn to rand.Reader for seed. discard the `nodeID`
- method 2, make `nodeID` a cycle shift to the higher bits. cycle shift for avoid lose precision. shift 40 bits seems good enough for avoid the nanosecond part.
- method 3, turn default `nodeID` to a better way. **But** I'm not sure if it'll break any cluster existing.

mreiferson:
Did a little digging on this. The historical requirements here are two-fold:

1. Per-`nsqd` / per-topic "globally" unique ID
2. Unique across `nsqd` restarts

Regarding (1), channels re-use the ID generated when the message is published to a topic. IDs are persisted via the diskqueue, which results in (2). The current algorithm satisfies (2) by assuming that time only ever moves forward, and generates IDs that are roughly time sorted.

IMO, the "right" answer here is an actual GUID (across all `nsqd`) for each message in a topic, but that obviously requires some persistent state and coordination. This is possible if we do things like #625 and #510, but practically those may never happen.

The question then becomes: is breaking backwards compatibility to be able to handle time going backwards actually worth it?

The argument in favor is that that you already can't _really_ depend on these message IDs outside of NSQ, given the scarcely documented and limited use of `--node-id`, so what value do they have as is?

The argument against seems to be that we're worried that newly generated message IDs may conflict with diskqueue persisted messages upon upgrade? Is that all?

slayercat:
If we do not need ID to be time sorted, use [rand.Reader](https://golang.org/pkg/crypto/rand/) to generate it could certainly make the ID globally unique (or at least make the probability of conflict to be negligible. )

image

daoxuans:
Isn't that a good solution?

Issue #1251, opened 2020-05-09
nsqd: ephemeral topic/channel churn can live-lock


ploxiln:
from https://github.com/nsqio/nsq/pull/1246

If consumers of ephemeral topics and channels come and go fast enough, for a short period of time, some can get stuck in indefinite retry loops.

Demo by @slayercat:

```go
package main

import nsq "github.com/nsqio/go-nsq"
import "time"
import "log"
import "fmt"

func nsqSubscribe(addr string, topic string, channel string, hdlr nsq.HandlerFunc) error {
	consumer, err := nsq.NewConsumer(topic, channel, nsq.NewConfig())
	if err != nil {
		print("new consumer error: ", err, "\n")
		time.Sleep(1 * time.Second) //wait 1s
		panic(err)
	}
	consumer.AddHandler(hdlr)
	for {
		err = consumer.ConnectToNSQD(addr)
		if err != nil {
			log.Printf("connect nsqd error: %v. retry\n", err)
			time.Sleep(1 * time.Second) //wait 1s
			continue
		} else {
			break
		}
	}
	_ = <-consumer.StopChan
	panic("nsq conn dead topic=" + topic + " channel=" + channel)
	return nil
}

func handlerFunc1(message *nsq.Message) error {
	message.Finish()
	return nil
}

func main() {
	for i := 0; i < 10; i++ {
		go nsqSubscribe("127.0.0.1:4150",
			fmt.Sprintf("testTopic%d#ephemeral", i),
			fmt.Sprintf("testChannel%d#ephemeral", i),
			handlerFunc1)
	}

	nsqSubscribe("127.0.0.1:4150", ("testTopic#ephemeral"), ("testChannel#ephemeral"), handlerFunc1)
}
```

Start a simple stand-alone `nsqd`, and then build and run the demo above in a bash loop:

```sh
go build -o demo demo.go

for I in $(seq 100); do
    ./demo &
    sleep 0.1
    kill $!
done
```

Then, nsqd will get stuck in the client topic/channel subscribe retry loop, seemingly indefinitely (long after the above loop finishes, and even after nsqd is signalled to exit gracefully):

```
[nsqd] 2020/05/09 13:13:17.694506 INFO: CHANNEL(testChannel6#ephemeral): deleting
[nsqd] 2020/05/09 13:13:17.694701 INFO: TOPIC(testTopic6#ephemeral): new channel(testChannel6#ephemeral)
[...]

PR #1253, opened 2020-05-12
Ephemeral: delete unneeded topic/channel periodically

slayercat:
fixs: #1251

### DIFF
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -56,7 +56,6 @@ type Channel struct {
 	paused         int32
 	ephemeral      bool
 	deleteCallback func(*Channel)
-	deleter        sync.Once
 
 	// Stats tracking
 	e2eProcessingLatencyStream *quantile.Quantile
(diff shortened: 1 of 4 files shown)


mreiferson:
Thanks for the PR, but I'm going to take a deeper look at #1251 and try to fix it "properly", so I'm going to close this.

Altruiste1:
if i want to delete unneeded topic/channel periodically,how to configure it

ploxiln:
This is not configurable. You could however write a script and run it periodically, that checks the `/stats` output of nsqd for topics and channels that meet your conditions for "unneeded", and delete them.

There are ephemeral channels and ephemeral topics which may or may not fit your use case ... but I think they are not appropriate for most use cases, because you don't want messages lost so easily while the channel/topic was used "recently" and will be used again "soon". Thus, a custom script with conditions appropriate for you (zero depth? idle time threshold?) is the way to go.

PR #1198, opened 2019-10-15
nsqd: fix stall in Exit() due to tcp producer conns

ploxiln:
Consumer connections are closed when topics are closed,
but tcp-protocol publish connections are not,
so add tcpServer.CloseAll().

problem introduced by https://github.com/nsqio/nsq/pull/1190

cc @jehiah @mreiferson

PR #1262, opened 2020-06-14
nsqd: close connections on exit (without another map)

mreiferson:
Going through the backlog of PRs that have landed, I propose a simpler alternative to #1198, which fixes a bug introduced in #1190.

We're already tracking _all_ client connections, including producers in `nsqd.clients`, so let's just use it to trigger connection close on exit.

Note: we should probably do this consistently across `nsqd` and `nsqlookupd` w/r/t what struct owns tracking and closing client conns, so maybe I should make a similar set of changes to `nsqlookupd`.

cc @ploxiln

### DIFF
--- a/nsqd/client_v2.go
+++ b/nsqd/client_v2.go
@@ -85,6 +85,11 @@ type ClientV2Stats struct {
 func (s ClientV2Stats) String() string {
 	connectTime := time.Unix(s.ConnectTime, 0)
 	duration := time.Since(connectTime).Truncate(time.Second)
+
+	_, port, _ := net.SplitHostPort(s.RemoteAddress)
+	id := fmt.Sprintf("%s:%s %s", s.Hostname, port, s.UserAgent)
+
+	// producer
 	if len(s.PubCounts) > 0 {
 		var total uint64
 		var topicOut []string
@@ -94,15 +99,17 @@ func (s ClientV2Stats) String() string {
 		}
 		return fmt.Sprintf("[%s %-21s] msgs: %-8d topics: %s connected: %s",
 			s.Version,
-			s.ClientID,
+			id,
 			total,
 			strings.Join(topicOut, ","),
 			duration,
 		)
 	}
+
+	// consumer
 	return fmt.Sprintf("[%s %-21s] state: %d inflt: %-4d rdy: %-4d fin: %-8d re-q: %-8d msgs: %-8d connected: %s",
 		s.Version,
-		s.ClientID,
+		id,
 		s.State,
 		s.InFlightCount,
 		s.ReadyCount,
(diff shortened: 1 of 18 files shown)


ploxiln:
Looks great :)

It does look like nsqlookupd could use similar treatment.

nsqd/client_v2.go
```diff
@@ -150,6 +150,15 @@ func newClientV2(id int64, conn net.Conn, ctx *context) *clientV2 {
 	return c
 }
 
+func (c *clientV2) Close() error {
+	err := c.Conn.Close()
+	close(client.ExitChan)
+	if client.Channel != nil {
+		client.Channel.RemoveClient(client.ID)
```
ploxiln, line 157: `s/client/c/`

nsqd/nsqd.go
```diff
@@ -427,9 +404,12 @@ func (n *NSQD) Exit() {
 	if n.tcpListener != nil {
 		n.tcpListener.Close()
 	}
-	if n.tcpServer != nil {
-		n.tcpServer.CloseAll()
+
+	n.clientLock.RLock()
+	for _, c := range n.clients {
```
ploxiln, line 409: looks like you removed `clientLock` and `clients` above?

mreiferson:
@ploxiln I don't remember what I was trying to do here with the latest commit 😬, so I'll ping when it's ready for review.

mreiferson:
alright I think this is ready

ploxiln (approved):
Looks good to me.

Seems like going through the `protocol.Client` interface causes a couple more type-assertions than would otherwise be needed, and might be avoidable ... but really not a big deal, it's fine :)

ploxiln:
(tested a bit locally too, seems to work)

nsqd/client_v2.go
```diff
@@ -47,6 +48,79 @@ type identifyEvent struct {
...
+	CipherSuite                   string `json:"tls_cipher_suite"`
+	TLSVersion                    string `json:"tls_version"`
+	TLSNegotiatedProtocol         string `json:"tls_negotiated_protocol"`
+	TLSNegotiatedProtocolIsMutual bool   `json:"tls_negotiated_protocol_is_mutual"`
+}
+
+func (s ClientV2Stats) String() string {
+	connectTime := time.Unix(s.ConnectTime, 0)
+	// truncate to the second
+	duration := time.Duration(int64(time.Since(connectTime).Seconds())) * time.Second
```
jehiah, line 88:
```suggestion
	duration := time.Since(connectTime).Truncate(time.Second)
```

jehiah (approved):

mreiferson:
Wow, that GitHub feature to "commit suggestion" is 😍

PTAL

jehiah (approved):
👍 squash?

mreiferson:
squashed

PR #1264, opened 2020-06-14
*: update and fix benchmark script

### LINKED ISSUE #1202: nsqd: ec2 benchmark is broken

mreiferson:
replaces #1204
fixes #1202

cc @martingrambow @ploxiln

### DIFF
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -1,4 +1,22 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
+
+#
+# This script bootstraps an NSQ cluster in EC2 and runs benchmarks.
+#
+# Requires python3 and the following packages:
+#   - boto3
+#   - paramiko
+#   - tornado
+#
+# AWS authentication is delegated entirely to the boto3 environment, see:
+#
+# https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html
+#
+# EC2 instances are launched into EC2 Classic, expecting a 'default' security group
+# that allows allows SSH (port 22) from 0.0.0.0/0 and an EC2 key pair created
+# (named 'default', but configurable via --ssh-key-name).
+#
+
 import sys
 import logging
 import time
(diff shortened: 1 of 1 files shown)


bench/bench.py
```diff
@@ -250,30 +257,22 @@ def run():
                      max_duration / total_ops * 1000 * 1000)
 
     for ssh_client, chan in nsqd_chans:
-        chan.close()
+        if hasattr(chan, 'close'):
```
ploxiln, line 260: odd that this is needed

bench/bench.py
```diff
@@ -283,15 +282,12 @@ def decomm():
...
-    tornado.options.define('secret_key', type=str, default='',
-                           help='AWS account secret key')
-    tornado.options.define('ami', type=str, default='ami-48fd2120',
+    # ubuntu 18.04 HVM instance store us-east-1
+    tornado.options.define('ami', type=str, default='ami-0938f2289b3fa3f5b',
                            help='AMI ID')
     tornado.options.define('ssh_key_name', type=str, default='default',
                            help='SSH key name')
-    tornado.options.define('instance_type', type=str, default='c3.2xlarge',
+    tornado.options.define('instance_type', type=str, default='c3.xlarge',
```
ploxiln, line 290: may I suggest `c5.large`

(which would also require changing the AMI to an HVM-EBS variant, I think ami-0ac80df6eff0e70b5 would do it, and maybe instance-create would then need a root-ebs-size parameter)

... but this could be updated in a separate PR, I could try that after this is merged

bench/bench.py
mreiferson: agreed, missed that change, updated

bench/bench.py
mreiferson: `c5.large` is VPC-only, so until we rewrite it to setup an entire VPC (not likely), this is Good Enough ™️ 

I didn't mean to change this from the original default, I'll update it.

PR #1266, opened 2020-07-02
add sig kill to gracefully stop nsqd docker container when running cm…

00arthur00:
add sig kill to gracefully stop nsqd docker container when running cmds like 'docker kill, docker-compose kill'

Signed-off-by: yapo.yang <arthur@yapoyangdeMacBook-Pro.local>

### DIFF
--- a/apps/nsqd/main.go
+++ b/apps/nsqd/main.go
@@ -25,7 +25,7 @@ type program struct {
 
 func main() {
 	prg := &program{}
-	if err := svc.Run(prg, syscall.SIGINT, syscall.SIGTERM); err != nil {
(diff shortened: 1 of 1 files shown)


ploxiln:
It is not possible to catch SIGKILL and gracefully shut down. If you use "docker stop" or "docker restart", then it sends SIGTERM first (which nsqd does catch and gracefully handle).

`man 7 signal` says:
> The signals SIGKILL and SIGSTOP cannot be caught, blocked, or ignored.

00arthur00:
@ploxiln Thanks for your reply, verified in my local env. Will close the issue soon.

Issue #1270, opened 2020-07-29
nsqlookupd: listen error lists incorrect address.


TangDH03:
File : nsq/nsqlookupd/nsqlookupd.go
Around line 43-46
````
l.httpListener, err = net.Listen("tcp", opts.HTTPAddress)
if err != nil {
	return nil, fmt.Errorf("listen (%s) failed - %s", opts.TCPAddress, err)
}
````
why it will print TCPAddress instead of HTTPAddress

jehiah:
@TangDH03 good catch. Would you be interested in contributing a fix for this?

TangDH03:
Sure,I will raise a PR

PR #1271, opened 2020-07-29
nsqlookupd: fix listen error lists incorrect address

TangDH03:
fix [#1270 ](https://github.com/nsqio/nsq/issues/1270)


ploxiln:
thanks

PR #1274, opened 2020-08-08
Sync topic channel state

### LINKED ISSUE #1199: nsqd: discover topic/channel paused state on new topic discovery
I'd like to provide an experience report from introducing new nsqd nodes to a cluster while operational activities which involve pausing channels are underway.

---

Currently nsqd discovers the topic/channel topology from `nqlookupd` when it receives a new message. This ensures that if you have say 10 channels on a topic, all 10 channels get a copy of the first message.

If a channel is in a paused state when the first message is published to a new `nsqd` node it's naturally expected that the new node would also end up with channels in an appropriately paused states. That doesn't happen unfortunately because `nsqlookupd` has no context about the paused state of a channel, and so it can't provide it when queried by nsqd on a new topic creation. (There also isn't technically anything that requires all hosts to be paused or unpaused together, even though that's obviously most common)

---

Feature Request: `nsqd`  propagate topic/channel pause states to `nsqlookupd`. On discovery, nsqlookupd can advise if the majority of clients for a topic/channel are paused and if so, nqsd will create a pre-paused topic or channel. If there are no clients for a topic/channel connected to nsqd at the time of topic discovery, nsqd should create a topic and channel un-paused.  When `nsqd` queries multiple `nqslookupd` if any response indicates paused, the result should be a paused state. i.e. `paused` is preferred.

Note: nsqlookupd response changes to support this should be backwards compatible.

thekingofworld:
Fix #1199


thekingofworld:
@jehiah @mreiferson @ploxiln This PR is ready for review. PTAL.

thekingofworld:
Here is the design:
1. Supports send topic/channel paused state through command[register], see [go-nsq pr](https://github.com/nsqio/go-nsq/pull/305).
2. Nsqd send states to nsqlookupd when topic/channel create,delete and paused,unpaused.
3. Nsqlookupd store these states, and nsqd can discover topic/channel paused state on new topic discovery through the HTTP API that nsqlookupd provides.

see more details in the code. : )

internal/clusterinfo/data.go
```diff
@@ -119,14 +119,18 @@ func (c *ClusterInfo) GetLookupdTopics(lookupdHTTPAddrs []string) ([]string, err
 
 // GetLookupdTopicChannels returns a []string containing a union of all the channels
 // from all the given lookupd for the given topic
-func (c *ClusterInfo) GetLookupdTopicChannels(topic string, lookupdHTTPAddrs []string) ([]string, error) {
+func (c *ClusterInfo) GetLookupdTopicChannels(topic string, lookupdHTTPAddrs []string) ([]string, map[string]bool, bool, error) {
```
mreiferson, line 122: This is an internal API, we can change it to return a proper data structure that includes various properties (like paused state) rather than a separate map.

nsqd/lookup.go
```diff
@@ -54,10 +55,11 @@ func connectCallback(n *NSQD, hostname string) func(*lookupPeer) {
 		for _, topic := range n.topicMap {
 			topic.RLock()
 			if len(topic.channelMap) == 0 {
-				commands = append(commands, nsq.Register(topic.name, ""))
+				commands = append(commands, nsq.Register(topic.name, "", int(atomic.LoadInt32(&topic.paused))))
```
mreiferson, line 58: As per my comment https://github.com/nsqio/go-nsq/pull/305#issuecomment-670968864, I think we should explore a new `nsqlookupd` protocol command like `SyncState` (or `UpdateRegistration`, naming TBD) that is sent whenever a topic/channel pause state changes.

nsqlookupd/http.go
```diff
@@ -95,9 +95,31 @@ func (s *httpServer) doChannels(w http.ResponseWriter, req *http.Request, ps htt
...
+	producers = producers.FilterByActive(s.ctx.nsqlookupd.opts.InactiveProducerTimeout,
+		s.ctx.nsqlookupd.opts.TombstoneLifetime)
+	if producers.IsTopicPaused() {
+		topicPaused = true
+	}
 	return map[string]interface{}{
-		"channels": channels,
+		"channels":             channels,
+		"channel_paused_stats": channelPausedStatsMap,
+		"topic_paused":         topicPaused,
```
mreiferson, line 122: This API is more challenging because it's public.

First, it doesn't seem appropriate to return `topic_paused` when the endpoint is `/channels` (regretting not making this a _little_ more RESTful).

Second, it's correct that we'll need to add a new response key for channel paused state, but perhaps what we should do is name it `channel_meta` which could be map of:

```go
type channelMeta struct {
    Paused bool `json:"paused"`
}
```

Which will allow us to add additional properties in the future more easily?

mreiferson:
Thanks for taking a first pass at this @thekingofworld! 🙏

internal/clusterinfo/data.go
thekingofworld: that's a good suggestion

thekingofworld:
@jehiah @mreiferson @ploxiln This PR is ready for review. PTAL.

nsqd/lookup.go
thekingofworld: 1. Add new command SyncState for topic/channel pause state change.
2.I think we can keep the ability of pause state delivery in command Register, through it we can reduce a call of command SyncState when topic/channel create.

internal/clusterinfo/data.go
thekingofworld: Add new data structure: TopicMeta, TopicChannelsMeta to solve it. see code. : )

thekingofworld:
@mreiferson  This PR is ready for review. PTAL.
/cc @ploxiln @jehiah

thekingofworld:
@mreiferson Ping.

internal/clusterinfo/data.go
```diff
@@ -119,24 +119,57 @@ func (c *ClusterInfo) GetLookupdTopics(lookupdHTTPAddrs []string) ([]string, err
 
 // GetLookupdTopicChannels returns a []string containing a union of all the channels
 // from all the given lookupd for the given topic
-func (c *ClusterInfo) GetLookupdTopicChannels(topic string, lookupdHTTPAddrs []string) ([]string, error) {
-	var channels []string
+func (c *ClusterInfo) GetLookupdTopicChannels(topic string, lookupdHTTPAddrs []string) (*TopicMeta, *TopicChannelsMeta, error) {
```
mreiferson, line 122: We can continue to simplify the return values here, since it's an internal API. How about:

```go
type ChannelState struct {
    Name string
    Paused bool
}

func (c *ClusterInfo) GetLookupdTopicChannels(topic string, lookupdHTTPAddrs []string) ([]ChannelState, error) {
```

I don't think it makes sense to force a return value that includes topic metadata.

nsqd/lookup.go
```diff
@@ -196,3 +228,38 @@ func (n *NSQD) lookupdHTTPAddrs() []string {
...
+
+func getChannelState(channel *Channel) *state {
+	if channel == nil {
+		return &state{Paused: false}
+	}
+	s := state{
+		Paused: channel.IsPaused(),
+	}
+	return &s
+}
```
mreiferson, line 265: For a few reasons, we should be able to drop all these "helpers" after we make the changes discussed here https://github.com/nsqio/go-nsq/pull/305#pullrequestreview-467949543

nsqd/nsqd.go
```diff
@@ -550,7 +572,16 @@ func (n *NSQD) DeleteExistingTopic(topicName string) error {
...
+func (n *NSQD) Notify(v interface{}, notifyType int) {
+	var nc chan interface{}
+	switch notifyType {
+	case NotifyTypeRegistration, NotifyTypeUnRegistration:
+		nc = n.notifyChan
+	case NotifyTypeStateUpdate:
+		nc = n.stateNotifyChan
+	default:
+		nc = n.notifyChan
+	}
```
mreiferson, line 584: Seems like it might come out simpler if we use the same channel, but send a more complex type across the channel to perform logic on the channel receive side.

nsqlookupd/http.go
```diff
@@ -39,6 +39,7 @@ func newHTTPServer(ctx *Context) *httpServer {
 	router.Handle("GET", "/topics", http_api.Decorate(s.doTopics, log, http_api.V1))
 	router.Handle("GET", "/channels", http_api.Decorate(s.doChannels, log, http_api.V1))
 	router.Handle("GET", "/nodes", http_api.Decorate(s.doNodes, log, http_api.V1))
+	router.Handle("GET", "/topic/meta", http_api.Decorate(s.doTopicMeta, log, http_api.V1))
```
mreiferson, line 42: I'd prefer to not add an endpoint named `/topic/meta`. The best alternative I can come up with is `/topics?topic=...`

nsqlookupd/registration_db.go
```diff
@@ -30,10 +30,16 @@ type PeerInfo struct {
...
+type State struct {
+	Paused bool `json:"paused"`
+}
+
 type Producer struct {
 	peerInfo     *PeerInfo
 	tombstoned   bool
 	tombstonedAt time.Time
+	topicState   *State
+	channelState *State
```
mreiferson, line 42: Let's just add a field `paused` here, it's internal.

Also these don't need to be pointer types.

thekingofworld:
/cc @mreiferson 
Thank you for suggestions above, it's great. I have finished all, the code seems to more clearly, please take another look.
Thanks again for your patience.

nsqlookupd/lookup_protocol_v1.go
```diff
@@ -191,6 +193,54 @@ func (p *LookupProtocolV1) UNREGISTER(client *ClientV1, reader *bufio.Reader, pa
...
+	}
+	var state bodyType
+	err = json.Unmarshal(body, &state)
+	if err != nil {
+		return nil, protocol.NewFatalClientErr(err, "E_INVALID", fmt.Sprintf("state body should be json"))
+	}
+
+	if channel != "" {
+		key := Registration{"channel", topic, channel}
+		if p.ctx.nsqlookupd.DB.UpdateProducer(key, client.peerInfo.id, state.Paused) {
```
mreiferson, line 229: I don't think we need the semantics here where `UpdateProducer` returns a `bool` indicating whether there was an actual state change. In other words, we can always set the value to the current state and log.

nsqlookupd/registration_db.go
```diff
@@ -230,6 +247,18 @@ func (pp Producers) PeerInfo() []*PeerInfo {
...
+func (pp Producers) IsPaused() bool {
+	pausedCount := 0
+	total := 0
+	for _, p := range pp {
+		if p.paused == true {
+			pausedCount++
+		}
+		total++
+	}
+	return pausedCount*2 > total
```
mreiferson, line 259: `>=`? I think in the two producer case, if one is paused we probably want to know that.

nsqd/context.go
```diff
@@ -3,3 +3,8 @@ package nsqd
 type context struct {
 	nsqd *NSQD
 }
+
+type notifyContext struct {
+	notifyType int
+	v          interface{}
+}
```
mreiferson, line 10: this type definition probably belongs in `nsqd/lookup.go` (where it's used)

nsqd/nsqd.go
```diff
@@ -32,6 +32,12 @@ const (
 	TLSRequired
 )
 
+const (
+	NotifyTypeRegistration = iota
+	NotifyTypeUnRegistration
+	NotifyTypeStateUpdate
+)
```
mreiferson, line 39: same for these const definitions, probably belong in `nsqd/lookup.go`

nsqd/nsqd.go
```diff
@@ -495,15 +501,28 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
 	// this makes sure that any message received is buffered to the right channels
 	lookupdHTTPAddrs := n.lookupdHTTPAddrs()
 	if len(lookupdHTTPAddrs) > 0 {
-		channelNames, err := n.ci.GetLookupdTopicChannels(t.name, lookupdHTTPAddrs)
+		topicMeta, err := n.ci.GetLookupdTopic(t.name, lookupdHTTPAddrs)
 		if err != nil {
-			n.logf(LOG_WARN, "failed to query nsqlookupd for channels to pre-create for topic %s - %s", t.name, err)
+			n.logf(LOG_WARN, "failed to query nsqlookupd for topic to pre-create for topic %s - %s", t.name, err)
```
mreiferson, line 506: > failed to query nsqlookupd for metadata to pre-create topic %s - %s

nsqd/nsqd.go
```diff
@@ -495,15 +501,28 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
...
 		}
-		for _, channelName := range channelNames {
-			if strings.HasSuffix(channelName, "#ephemeral") {
+		if topicMeta.Paused {
+			err = t.Pause()
+			if err != nil {
+				n.logf(LOG_WARN, "failed to pause topic to pre-pause for topic %s - %s", t.name, err)
```
mreiferson, line 511: > failed to pre-pause topic %s - %s

nsqd/nsqd.go
```diff
@@ -495,15 +501,28 @@ func (n *NSQD) GetTopic(topicName string) *Topic {
...
+		for _, channelState := range channelStates {
+			if strings.HasSuffix(channelState.Name, "#ephemeral") {
 				continue // do not create ephemeral channel with no consumer client
 			}
-			t.GetChannel(channelName)
+			c := t.GetChannel(channelState.Name)
+			if channelState.Paused {
+				err = c.Pause()
+				if err != nil {
+					n.logf(LOG_WARN, "failed to pause channel %s to pre-pause for topic %s - %s", c.name, t.name, err)
```
mreiferson, line 523: > failed to pre-pause channel %s for topic %s - %s

nsqd/nsqd.go
```diff
@@ -550,7 +569,7 @@ func (n *NSQD) DeleteExistingTopic(topicName string) error {
 	return nil
 }
 
-func (n *NSQD) Notify(v interface{}) {
+func (n *NSQD) Notify(v notifyContext) {
```
mreiferson, line 572: don't think this change is necessary, right?

nsqlookupd/http.go
```diff
@@ -77,13 +77,40 @@ func (s *httpServer) doInfo(w http.ResponseWriter, req *http.Request, ps httprou
 	}, nil
 }
 
+type topicMeta struct {
```
mreiferson, line 80: I'd prefer these response types be inlined in some way to more clearly indicate that they're not used elsewhere

nsqlookupd/http.go
```diff
@@ -77,13 +77,40 @@ func (s *httpServer) doInfo(w http.ResponseWriter, req *http.Request, ps httprou
...
+		}, nil
+	}
+
 	topics := s.ctx.nsqlookupd.DB.FindRegistrations("topic", "*", "").Keys()
 	return map[string]interface{}{
 		"topics": topics,
 	}, nil
 }
 
+type channelMeta struct {
```
mreiferson, line 110: same here, let's inline this please

nsqlookupd/http.go
```diff
@@ -77,13 +77,40 @@ func (s *httpServer) doInfo(w http.ResponseWriter, req *http.Request, ps httprou
...
+	if topicName != "" {
+		var tm topicMeta
+		producers := s.ctx.nsqlookupd.DB.FindProducers("topic", topicName, "*")
+		producers = producers.FilterByActive(s.ctx.nsqlookupd.opts.InactiveProducerTimeout,
+			s.ctx.nsqlookupd.opts.TombstoneLifetime)
+		if producers.IsPaused() {
+			tm.Paused = true
+		}
+		return map[string]interface{}{
+			"topic_meta": tm,
```
mreiferson, line 100: This new `topic` query param acts as a _filter_, so we should respond with the same fields and data either way.

nsqd/lookup.go
```diff
@@ -118,35 +133,68 @@ func (n *NSQD) lookupLoop() {
...
-			case *Topic:
-				// notify all nsqlookupds that a new topic exists, or that it's removed
-				branch = "topic"
-				topic := val.(*Topic)
-				if topic.Exiting() == true {
-					cmd = nsq.UnRegister(topic.name, "")
-				} else {
-					cmd = nsq.Register(topic.name, "")
+			notifyCtx, ok := val.(notifyContext)
+			if ok {
```
mreiferson, line 139: I realize it's theoretically possible for there to be a case where a type other than `notifyContext` is sent over the channel, but it's not _practically_ possible (we can verify this manually at all call sites).

Let's drop the indentation here and do this instead:

```go
if !ok {
    panic("non-notifyContext sent to notifyChan - should never happen")
}
```

nsqd/lookup.go
```diff
@@ -118,35 +133,68 @@ func (n *NSQD) lookupLoop() {
...
+						// notify all nsqlookupds that topic state changed
+						branch = "topic"
+						topic := notifyCtx.v.(*Topic)
+						cmd, err = nsq.SyncState(topic.name, "", map[string]interface{}{"paused": topic.IsPaused()})
+						if err != nil {
+							n.logf(LOG_ERROR, "NSQD: build cmd err: %s", err)
+						}
+					}
+				default:
+					n.logf(LOG_ERROR, "NSQD: unknown notifyType: %d", notifyCtx.notifyType)
```
mreiferson, line 187: Same here, this should be a `panic` because it's essentially a programming error and behavior is undefined.

thekingofworld:
> Continuing to improve! Thanks for your patience 🙏
> 
> My big high-level concern now is that these new paths in `nsqlookupd` are going to be significantly more expensive. In the `/channels` case, we're doing multiple passes over data in `RegistrationDB` to filter and lookup paused state in producers. Considering #1208, we should think about better data structures to use to store this data to make these newer, more complex queries performant.
> 
> cc @andyxning

okay, the performance issues need more time to solve, I will try it.

nsqd/context.go
thekingofworld: done.

nsqd/nsqd.go
thekingofworld: done.

nsqd/nsqd.go
thekingofworld: done.

nsqd/nsqd.go
thekingofworld: done.

nsqd/nsqd.go
thekingofworld: done.

nsqd/nsqd.go
thekingofworld: done.

nsqd/lookup.go
thekingofworld: done.

nsqd/lookup.go
thekingofworld: done.

nsqlookupd/http.go
thekingofworld: done.

nsqlookupd/http.go
thekingofworld: done.

nsqlookupd/http.go
thekingofworld: done.

nsqlookupd/registration_db.go
thekingofworld: done.

nsqlookupd/lookup_protocol_v1.go
thekingofworld: done.

thekingofworld:
I added some benchmarks for the `/channels` case, the execution results are as follows:
image
It seems still fast enough, I doubt if the situation of write lock starvation really exists?

thekingofworld:
@mreiferson Ping.

thekingofworld:
@mreiferson @jehiah @ploxiln Anyone can help to take a look at this? It has waited for more than 10 days...

thekingofworld:
Hi folks, let's keep going.[Doge] if there are any problems, please let me know and I will try my best to solve them.
Let us make nsq more powerful.

mreiferson:
hi @thekingofworld, sorry for the delay, I'll try to find some time this week to take a closer look. We do know that at larger scale there are performance issues that exist (https://github.com/nsqio/nsq/pull/1208), and I this PR will make some queries perform worse.

thekingofworld:
Well, I read the related discussion about the PR #1208.
I'm wondering if there is more detailed data that indicate how slow these queries are, such as how long does a query take, and how many topics, how many channels, how many producers...
After recurring this slow situation, we can verify how much efficiency the new optimization scheme can improve.

thekingofworld:
> @mreiferson
> I added some benchmarks for the `/channels` case, the execution results are as follows:
> 
> image
> 
> It seems still fast enough, I doubt if the situation of write lock starvation really exists?

By the way, benchmarks show that these queries are still fast.

ploxiln:
I think that the situation that #1208 is attempting to optimize, which the existing benchmark setups here do not emulate, is when there are many channels per topic, and they are constantly being added and removed (as ephemeral channels tend to do when consumers connect and disconnect). Each add and remove does a "write" lock/unlock of the RWMutex, and this blocks all the "read" locks of the lookups. The current benchmarks setup just two channels per topic, and do lookups without concurrent updates, and only time the lookups part.

thekingofworld:
@ploxiln yeah, the situation you described is right.
But I think the key is how `FindRegistrations ` will behave in `*` case when db is very large.
So I added some benchmarks for this, the results are shown below: 
image
When topic=64, channel=512, producer=512, each operation takes **11ms**.
When topic=512, channel=512, producer=512, Memory exceeds 10G, I think this is beyond normal.
It seems like we can add additional maps to improve performance for `*` case.
/cc @ploxiln @mreiferson

thekingofworld:
@mreiferson @ploxiln Let's continue to this.
Our previous discussion stopped at the performance of the case of `/channels`. Recently I added an additional map to improve the performance of the query, which is essentially optimizing the function `FindRegistrations`, which is called with `* `.
PTAL.

thekingofworld:
@mreiferson @ploxiln  Ping.

mreiferson:
Thanks for continuing to push this forward @thekingofworld.

The approach here will now be much more costly on the write (registration) side, because each write will now have to walk the whole structure and produce the cached "high performance" map.

Instead, we should concurrently update both the standard and new "high performance" maps during each mutation.

The discussion in https://github.com/nsqio/nsq/pull/1208#issuecomment-549509152 (from "if sync.Map everywhere is too much overhead" down) and https://github.com/nsqio/nsq/pull/1208#issuecomment-643821646 has the most relevant context for what needs to happen here.

thekingofworld:
> Instead, we should concurrently update both the standard and new "high performance" maps during each mutation.
agree, update highspeed map during each mutation instead walk the whole structure and produce the cached "high performance" map.
@mreiferson done, PTAL.

thekingofworld:
@mreiferson  ping.let's finish it...

PR #1283, opened 2020-09-16
log GenerateID error instead of silently spinning

Chaiyq:
we adjust system time, than topic publish error, than break point found error when create guid and keep retry

### DIFF
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -489,12 +489,13 @@ func (t *Topic) GenerateID() MessageID {
 		if err == nil {
 			return id.Hex()
 		}
-		//avoid log output too many times, so print the first error and sampling
+		//avoid log output too many times, so just print the first error and sampling
 		if i%10000 == 0 {
 			t.ctx.nsqd.logf(LOG_ERROR,
 				"TOPIC(%s): failed to create guid - %s", t.name, err)
 		}
 		time.Sleep(time.Millisecond)
 		i++
+		continue
 	}
 }
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -496,6 +496,5 @@ func (t *Topic) GenerateID() MessageID {
 		}
 		time.Sleep(time.Millisecond)
 		i++
-		continue
 	}
 }
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -487,6 +487,8 @@ func (t *Topic) GenerateID() MessageID {
 retry:
 	id, err := t.idFactory.NewGUID()
 	if err != nil {
+		t.ctx.nsqd.logf(LOG_ERROR,
+			"TOPIC(%s): failed to create guid - %s", t.name, err)
 		time.Sleep(time.Millisecond)
 		goto retry
 	}
--- a/nsqd/topic.go
+++ b/nsqd/topic.go
@@ -3,15 +3,14 @@ package nsqd
 import (
 	"bytes"
 	"errors"
-	"strings"
-	"sync"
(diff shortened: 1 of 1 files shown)


ploxiln:
I think this "spamming" the logging output with 1000 messages per second would not be good. How about logging just once, on just the first error?

Also, I think this `goto retry` loop could pretty easily be converted to a `for ;; {` loop, which IMHO would be a good little cleanup.

Chaiyq:
@ploxiln ,yes, you are right, thank you very much, we should avoid printing too many logs. I will optimize and than push.

nsqd/topic.go
```diff
@@ -3,15 +3,14 @@ package nsqd
...
-	"time"
-
 	"github.com/nsqio/go-diskqueue"
 	"github.com/nsqio/nsq/internal/lg"
 	"github.com/nsqio/nsq/internal/quantile"
 	"github.com/nsqio/nsq/internal/util"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
```
ploxiln, line 13: can you put these imports back at the top please?

(If you leave a blank line between them and the github.com/... imports, then gofmt won't sort them together, it will keep them in separate sections.)

nsqd/topic.go
```diff
@@ -484,11 +483,19 @@ func (t *Topic) IsPaused() bool {
...
+		}
+		//avoid log output too many times, so just print the first error and sampling
+		if i%10000 == 0 {
+			t.ctx.nsqd.logf(LOG_ERROR,
+				"TOPIC(%s): failed to create guid - %s", t.name, err)
+		}
 		time.Sleep(time.Millisecond)
-		goto retry
+		i++
+		continue
```
ploxiln, line 499: I don't think any `continue` is needed, that's what happens normally at the end of the loop body

nsqd/topic.go
Chaiyq: OK，thank you for your reminding

nsqd/topic.go
Chaiyq: yes, you are right

ploxiln:
When done revising this, please squash down into one git commit.

nsqd/topic.go
Chaiyq: i delete the redundant continue

Chaiyq:
@ploxiln ,thank you very much

Chaiyq:
> 
> 
> When done revising this, please squash down into one git commit.

ok, i'll squash down into one git commit. thankyou

Chaiyq:
@ploxiln , I'm very sorry, it seems like this request could not associate to my fork branch. Can i close this pull request and create a new? Very sorry for the inconvenience.

ploxiln:
It's OK :) you can open a new pull request if that's the easiest way for you. But it should be possible to force-push up over the original branch name you used when opening the pull request, in order to update this pull-request...

Chaiyq:
@ploxiln ,I create a new pull request: log GenerateID error instead of silently spinning #1285, but it remind me Coverage decreased (-0.07%) to 60.634%, is this ok? or i need optimize it. Thanks sincerely.

ploxiln:
The small coverage fluctuation is not a real problem, don't worry about that.

ploxiln:
merged in #1285

Chaiyq:
@ploxiln ，thank you very much. I learned a lot from the push process. Thank you for your guidance sincerely.

ploxiln:
thank you for contributing and responding to feedback :)

PR #1291, opened 2020-10-28
use iota opt code.

imxyb:


ploxiln:
I do not like this change, sorry.
  * I think 0, 1, 2 are very concise and clear
  * these values used in messages sent over the network, so as other constants may be added later, it's important that these don't change accidentally

mreiferson:
agreed, appreciate the PR but think this is more readable as is.

PR #1292, opened 2020-10-28
nsqd: add pool to minimize buffer allocation on message send

imxyb:


ploxiln:
Maybe the function names should be more specific, like `GetMessageBuffer()`?

Do you have a benchmark that shows a significant improvement in performance from this change?

imxyb:
> Maybe the function names should be more specific, like `GetMessageBuffer()`?
> 
> Do you have a benchmark that shows a significant improvement in performance from this change?

Thank you for your advice. I see the buffer in the paclage `nsqd`:`buffer_ pool.go` have similar codes, can I extract this part of the code for common use?

ploxiln:
It looks like it is already extracted, and you can just use it, in one more place.

imxyb:
> have a benchmark that shows a significant improvement in performance from this change?

before: 
```
# using --mem-queue-size=1000000 --data-path= --size=200 --batch-size=200
# compiling/running nsqd
# creating topic/channel
# compiling bench_reader/bench_writer
PUB: [bench_writer] 2020/11/01 12:00:24 duration: 10.021804433s - 32.735mb/s - 171625.780ops/s - 5.827us/op
SUB: [bench_reader] 2020/11/01 12:00:34 duration: 10.00603778s - 32.787mb/s - 171896.213ops/s - 5.817us/op
waiting for pprof...
```

after:
```
# using --mem-queue-size=1000000 --data-path= --size=200 --batch-size=200
# compiling/running nsqd
# creating topic/channel
# compiling bench_reader/bench_writer
PUB: [bench_writer] 2020/11/01 12:15:40 duration: 10.027253623s - 33.219mb/s - 174165.336ops/s - 5.742us/op
SUB: [bench_reader] 2020/11/01 12:15:50 duration: 10.00180929s - 33.304mb/s - 174608.408ops/s - 5.727us/op
waiting for pprof...
```

ploxiln:
Looks good to me, thanks. The commit title could be a bit better, I suggest something like

> nsqd: use buffer pool for serializing message for client

nsqd/protocol_v2.go
```diff
@@ -124,7 +124,9 @@ func (p *protocolV2) IOLoop(conn net.Conn) error {
 
 func (p *protocolV2) SendMessage(client *clientV2, msg *Message) error {
 	p.ctx.nsqd.logf(LOG_DEBUG, "PROTOCOL(V2): writing msg(%s) to client(%s) - %s", msg.ID, client, msg.Body)
-	var buf = &bytes.Buffer{}
+
+	buf := bufferPoolGet()
+	defer bufferPoolPut(buf)
```
mreiferson, line 129: looks like defers finally got basically as fast as inline calls, which means we can separately clean up a lot of places we were avoiding them...

nsqd/buffer_pool.go
```diff
@@ -18,5 +18,6 @@ func bufferPoolGet() *bytes.Buffer {
 }
 
 func bufferPoolPut(b *bytes.Buffer) {
+	b.Reset()
```
mreiferson, line 21: This is interesting, what does this imply was happening before without this? Were the pooled buffers _growing_ because they had not been reset?

mreiferson:
LGTM other than that small question

nsqd/buffer_pool.go
imxyb: i find that the `writeMessageToBackend` of `message.go` also called buf.Reset, I think it should be called in bufferPoolPut

nsqd/protocol_v2.go
imxyb: Yes, but so what do you want me to modify here?

ploxiln: no, I think nothing to modify here, it was just a side comment that we used to avoid using defer in performance-sensitive functions, but now it is fine

nsqd/buffer_pool.go
ploxiln: There were a couple different schemes for caching the `bytes.Buffer` for `writeMessageToBackend()`, so the reset was in the latter function ... it all worked, and still does. I think the remaining task is to make it so there aren't redundant `Reset()` calls, which are a bit confusing. I think there are two options:
   * take it out of `writeMessageToBackend()`, and then put it in the callers that don't use the buffer pool (they re-use one buffer for a loop of message ops, which is probably more efficient than going through the pool)
   * change `writeMessageToBackend()` to call `bufferPoolGet()` itself, and remove the `*bytes.Buffer` param to it

imxyb: done, pls check again

ploxiln: looks good to me ... can you repeat the benchmark once more please?

imxyb:
@ploxiln the latest benchmark:
# using --mem-queue-size=1000000 --data-path= --size=200 --batch-size=200
# compiling/running nsqd
# creating topic/channel
# compiling bench_reader/bench_writer
PUB: [bench_writer] 2020/11/12 10:21:33 duration: 10.029296172s - 33.213mb/s - 174129.866ops/s - 5.743us/op
SUB: [bench_reader] 2020/11/12 10:21:43 duration: 10.005897058s - 33.290mb/s - 174537.074ops/s - 5.729us/op

ploxiln:
thanks!

PR #1294, opened 2020-10-31
apps: optimized float array string method

donutloop:
Hi there,

I clean up a bit the float array type in apps.

**Old performance**

goos: linux
goarch: amd64
pkg: github.com/nsqio/nsq/internal/app
BenchmarkFloatArray
BenchmarkFloatArray-16    	  745626	      1541 ns/op

**New performance**

goos: linux
goarch: amd64
pkg: github.com/nsqio/nsq/internal/app
BenchmarkFloatArray
BenchmarkFloatArray-16    	  821539	      1399 ns/op

### DIFF
--- a/internal/app/float_array.go
+++ b/internal/app/float_array.go
@@ -17,7 +17,6 @@ func (a *FloatArray) Set(param string) error {
 		v, err := strconv.ParseFloat(s, 64)
 		if err != nil {
 			log.Fatalf("Could not parse: %s", s)
-			return nil
 		}
 		*a = append(*a, v)
 	}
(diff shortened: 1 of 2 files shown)


ploxiln:
Well, I guess it doesn't hurt ... but this is only used for parsing one option flag at startup. So I personally don't think it is worth any additional lines of code, sorry.

donutloop:
@ploxiln isn’t just changing implementation of the float array additionally it  contains new test cases for it to improve the test code coverage... in case of you don‘t change your mind then please close the pull request.

mreiferson:
agree with @ploxiln but thanks for proposing this change

Issue #1073, opened 2018-08-21
nsqd: docker networking, broadcast TCP and HTTP ports


heat:
Hi folks, sory about my bad english.
I'm trying get up nsq services at containerum, at containerum the public port to services and container are random. With nsqlookupd and nsqadmin it's easy not a big trouble.
But when we try to run nsqd it's registrer ther right broadcast url but the follow ports  that nsqd register is that expose in container and not the service port random assigned with the service.
```
  |Name   | Port       | Protocol  | Link                    |   
  | http  | 24455:4150 | TCP       | x2.containerum.io:24455 |  
  | https | 14384:4151 | TCP       | x2.containerum.io:14384 |

```
In admin I got this link to node x2.containerum.io:4151 but it should be *x2.containerum.io:14384*.
I appreciate your time

ploxiln:
This is not currently possible.

nsqd tells nsqlookupd how this nsqd can be reached like so:

```
		ci["version"] = version.Binary
		ci["tcp_port"] = n.RealTCPAddr().Port
		ci["http_port"] = n.RealHTTPAddr().Port
		ci["hostname"] = hostname
		ci["broadcast_address"] = n.getOpts().BroadcastAddress

		cmd, err := nsq.Identify(ci)
```

So you can use the `-broadcast-address` option to give nsqd a different hostname or ip address to give to nsqlookupd and thus to other clients (and nsqadmin). But you can't specify different ports than the actual ports that nsqd listens on.

I use `--net=host` when running nsqd in a docker container. Alternatively, you could pick random ports, but pass them through as the same ports, and configure nsqd to listen on those particular random ports.

heat:
Thank you @ploxiln .
I my service docker containerum isn't possible to especify `--net`. 

>  Alternatively, you could pick random ports, but pass them through as the same ports

This will be my next shot, but could I open a PR that adds two arguments `-broadcast-tcp-port` and `-broadcast-http-port` as a optional params ?  Or we could  add some doc in docker session about how to 
 configure port foward?

ploxiln:
I'm not really opposed to either idea.

cloudzhou:
@ploxiln face the same question
so any pr?

sidvenu:
I'm not sure why this feature is not given priority yet, considering this can help simplify distributed deployments on systems. @ploxiln is anyone working on this PR right now? If not, can someone from my organisation work on this feature?

ploxiln:
[...]

PR #1297, opened 2020-11-13
nsqd: seperate broadcast ports for tcp and http

shyam-king:
# Description

When using nsqd within containerised instances, we would require not only a different broadcast address but also different ports to which consumers must contact in order to reach the nsqd, thus, solving #1073 .

# Changes
- adds flags to accept `broadcast-tcp-address` and `broadcast-http-address`
- registers nsqd with lookupd using `broadcast-tcp-address` and `broadcast-http-address` instead of the bound ports

### DIFF
--- a/apps/nsqd/options.go
+++ b/apps/nsqd/options.go
@@ -114,6 +114,8 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> to query auth server (may be given multiple times)")
 	flagSet.String("broadcast-address", opts.BroadcastAddress, "address that will be registered with lookupd (defaults to the OS hostname)")
+	flagSet.Int("broadcast-tcp-port", opts.BroadcastTCPPort, "TCP port that will be registered with lookupd (defaults to the TCP port that this nsqd is listening on)")
+	flagSet.Int("broadcast-http-port", opts.BroadcastHTTPPort, "HTTP port that will be registered with lookupd (defaults to the HTTP port that this nsqd is listening on)")
 	lookupdTCPAddrs := app.StringArray{}
 	flagSet.Var(&lookupdTCPAddrs, "lookupd-tcp-address", "lookupd TCP address (may be given multiple times)")
 	flagSet.Duration("http-client-connect-timeout", opts.HTTPClientConnectTimeout, "timeout for HTTP connect")
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -119,20 +119,6 @@ func New(opts *Options) (*NSQD, error) {
 		return nil, errors.New("--node-id must be [0,1024)")
 	}
 
-	if opts.StatsdPrefix != "" {
-		var port string
-		_, port, err = net.SplitHostPort(opts.HTTPAddress)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse HTTP address (%s) - %s", opts.HTTPAddress, err)
-		}
-		statsdHostKey := statsd.HostKey(net.JoinHostPort(opts.BroadcastAddress, port))
-		prefixWithHost := strings.Replace(opts.StatsdPrefix, "%s", statsdHostKey, -1)
-		if prefixWithHost[len(prefixWithHost)-1] != '.' {
-			prefixWithHost += "."
-		}
-		opts.StatsdPrefix = prefixWithHost
-	}
-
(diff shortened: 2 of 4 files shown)


nsqd/options.go
```diff
@@ -149,3 +156,27 @@ func NewOptions() *Options {
 		TLSMinVersion: tls.VersionTLS10,
 	}
 }
+
+func applyDefaultOptions(opts *Options) error {
```
ploxiln, line 160: We don't really need a new function for this, we already have a pattern for option defaults that depend on other options:

```go
func New(opts *Options) (*NSQD, error) {
	...
	if opts.Logger == nil {
		opts.Logger = log.New(os.Stderr, opts.LogPrefix, log.Ldate|log.Ltime|log.Lmicroseconds)
	}
	...
	if opts.StatsdPrefix != "" {
		var port string
		_, port, err = net.SplitHostPort(opts.HTTPAddress)
		if err != nil {
			return nil, fmt.Errorf("failed to parse HTTP address (%s) - %s", opts.HTTPAddress, err)
		}
		statsdHostKey := statsd.HostKey(net.JoinHostPort(opts.BroadcastAddress, port))
		prefixWithHost := strings.Replace(opts.StatsdPrefix, "%s", statsdHostKey, -1)
		if prefixWithHost[len(prefixWithHost)-1] != '.' {
			prefixWithHost += "."
		}
		opts.StatsdPrefix = prefixWithHost
	}

	if opts.TLSClientAuthPolicy != "" && opts.TLSRequired == TLSNotRequired {
		opts.TLSRequired = TLSRequired
	}
	...
```

I defer to @mreiferson about whether we want to move all of that kind of logic out of `nsqd.New()` over to here, but I lean towards "nah".

shyam-king: I moved it to `New()`. 

I guess shifting to new a function would be more appropriate in a different PR.

ploxiln: Maybe we should also update the logic quoted above, which uses the http port and broadcast address to form a StatsdPrefix, to use the broadcast http port?

apps/nsqd/options.go
```diff
@@ -114,6 +114,8 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> to query auth server (may be given multiple times)")
 	flagSet.String("broadcast-address", opts.BroadcastAddress, "address that will be registered with lookupd (defaults to the OS hostname)")
+	flagSet.Int("broadcast-tcp-port", opts.BroadcastTCPPort, "TCP port that will be registered with lookupd (defaults to the TCP port that this nsqd is listening to)")
+	flagSet.Int("broadcast-http-port", opts.BroadcastHTTPPort, "HTTP port that will be registered with lookupd (defaults to the HTTP port that this nsqd is listening to)")
```
ploxiln, line 118: minor nit: we usually write "listening _on_ a port" instead of "listening _to_ a port"

nsqd/nsqd.go
```diff
@@ -80,6 +81,26 @@ type NSQD struct {
...
 	var err error
 
+	if opts.BroadcastHTTPPort == 0 {
+		_, portStr, err := net.SplitHostPort(opts.HTTPAddress)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse HTTP address (%s) - %s", opts.HTTPAddress, err)
+		}
+
+		portInt, _ := strconv.ParseInt(portStr, 10, 32)
+		opts.BroadcastHTTPPort = int(portInt)
```
mreiferson, line 91: The reason why all the tests are failing is because they use port `0` to listen on a random free port.

E.g. https://github.com/nsqio/nsq/pull/1297/files#diff-6b512c5c86dadb7cbbe6f77d08d7a080d3b0d8780eb8801ee41dc59c803d9c2eL19 lazily uses `RealTCPAddr`() to send the actual port that was bound.

mreiferson (changes requested):

nsqd/nsqd.go
ploxiln: ... so if this moves down below where the Listeners are created, and uses `RealTcpAddr()` like before, that could work ...

ploxiln:
Can you squash down to one commit please? The prefix of the commit title should be "nsqd:"

(I suggest a commit title like: "nsqd: add flags --broadcast-tcp-port, --broadcast-http-port")

mreiferson (approved):

ploxiln:
one last minor nit, commit title has "broadcast-http-address" instead of "broadcast-http-port"

Issue #1300, opened 2020-11-20
nsq: support region/zone aware msg consumption [RFC]


jehiah:
NSQ does not facilitate easy ways to manage a cluster that spans a geographic boundary where it's ideal to prefer same-datacenter consumption to avoid extra network bandwidth and latency. This proposal is to provide a way to influence message consumption while preserving realtime behavior and promoting distributed fault tolerant configuration

# Problem Overview 

Consider a nsqd instance on host `h1` and `h2` both in different datacenters (`d1` and `d2`) with a message on topic `t` and two consumers on hosts `c1` and `c2` also in datacenters `d1` and `d2`. It's desirable to have `c1` primarily consume messages from `h1` and to have `c2` primarily consume messages from `h2` while still providing some reliability so that if `c1` or `c2` are interrupted the other will consume all messages w/o delay.

# Current Implementation

nsqd uses a Go channel `memChannel` for each channel and it's consumed by a "message pump" goroutine for each client.  When a new message arrives it's put onto `memChannel` and a message pump (at random) which is available will consume that message and write to the client. If the `memChannel` is not available (full) the message will be written to disk and separately read from disk to put back onto `memChannel`.

# Proposal

nsqd will have three Go channels for messages internal to each "channel" - an unbuffered zoneLocal channel, and a secondary unbuffered regionLocal channel and a third global channel (buffer controlled by CLI arguments).  A new message will first attempt to be written to the zoneLocal channel, then the regionLocal and finally the global channel; if all nonblocking channel writes are skipped the message will be written to disk matching existing behavior.

The IDENTIFY message will provide new parameters "topology-region" and "topology-zone" that when matched with a new nsqd parameters "--topology-region" and "--topology-zone" will allow the message pump to consume from those channels in addition to the global channel. When either client or server values are unset it will cause a client to consume only from the global priority.

# Expected Behavior

At low number of clients (and generally higher message arrival rate) it is expected that some new messages will arrive while the message pump is busy, and some messages that might otherwise be handled by the zoneLocal channel will get handled by the regionLocal or global channel. This behavior may be particularly frequent when messages are read from disk to recover from a backlog as messages will arrive immediately.

This may make picking good MaxInFlight values more difficult because clients will normally only receive messages from a zoneLocal portion of their max-in-flight.

# Performance

This may lower message throughput in cases where messages are not sent to a zoneLocal client.

cc: @mreiferson @ploxiln 
cc: #1254 tracking "cloud native" improvements

[...]

PR #1301, opened 2020-11-21
nsqd: add topology region/zone aware message consumption

jehiah:
This updates nsqd to have a `--topology-region` and `--topology-zone` config argument and prefer sending messages to same-zone and same-region consumers as proposed in #1300


jehiah:
I think this is ready for a first round of review @mreiferson @ploxiln (along w/ it's pair nsqio/go-nsq#312

Some items to decide on - 
* Should `-mem-queue-size=0` disable zone local and region local consumption? (i.e. should it continue to effectively write to disk first?)
* Is there an approach to changing the consumption of the disk read chan so it prefers zoneLocal and regionLocal consumers? (this would probably mean a goroutine that consumes it and does `put()` instead of messagePump consuming the disk backend directly. If there is an approach is it important to do as part of this?

Once this is squared away and we are happy with it i'll follow up with documentation PR's and exposing region/zone in nsqadmin.

nsqd/protocol_v2.go
```diff
@@ -313,6 +330,32 @@ func (p *protocolV2) messagePump(client *clientV2, startedChan chan bool) {
...
+			}
+			msg.Attempts++
+
+			subChannel.StartInFlightTimeout(msg, client.ID, msgTimeout)
+			client.SendingMessage()
+			err = p.SendMessage(client, msg)
+			if err != nil {
+				goto exit
+			}
+			flushed = false
```
ploxiln, line 352: I think this duplication could be reduced, if `msg` was declared outside the select, then after the select if not nil the common msg logic runs.

ploxiln (commented):
I had some ideas, for what I think are minor simplifications that leave the overall design the same. Overall looks good to me.

nsqd/protocol_v2.go
```diff
@@ -249,12 +252,24 @@ func (p *protocolV2) messagePump(client *clientV2, startedChan chan bool) {
 			// last iteration we flushed...
 			// do not select on the flusher ticker channel
 			memoryMsgChan = subChannel.memoryMsgChan
+			if subChannel.topologyZone == topologyZone && topologyZone != "" {
```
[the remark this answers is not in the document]
ploxiln: Here's another little simplification idea: only one `localMsgChan`. If region+zone match, use zone chan, elif region match use region chan, else nil. The zone chan is always written before the region chan so it's redundant to listen on both.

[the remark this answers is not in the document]
jehiah: I like the way [Kubernetes frames this](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone)

> Kubernetes makes a few assumptions about the structure of zones and regions:
> 
> 1. regions and zones are hierarchical: zones are strict subsets of regions and no zone can be in 2 regions
> 1. zone names are unique across regions; for example region "africa-east-1" might be comprised of zones "africa-east-1a" and "africa-east-1b"

I'm thinking of zone labels in the same way - they are expected to be unique (which i think is how all cloud platforms treat this) Practically this means a client might specify only zone, and a nsqd might specify region & zone and that's ok. the zone matches or it doesn't. 

(this would be something to cover in docs of course)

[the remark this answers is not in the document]
ploxiln: OK, that works fine. It still works with the suggestions to calculate "isLocalZone" and "isLocalRegion" boolean properties just once per identify, and have just one "localMsgChan" in `messagePump()` :)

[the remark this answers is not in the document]
jehiah: oh yes, definitely 👍 on some booleans to simplify this code

thekingofworld:
After reading the issue discussion and PR related code on this topic, I prefer the view of flow control:
> What this proposal feels like is another form of flow control. @mreiferson 
Under the current situation, nsq does not have a flow control strategy for message consumption, because multiple clients receive msg directly from the channel, and there is no additional code to intervene in this process.
The code in this PR adds zone and region-related channels, but they are unbuffered, which still does not play a role in flow control based on region in the case of high concurrency.
Can we consider adding a layer of control logic between the channel and the client? For example, choose a suitable client for a message,these clients should be sorted and grouped based on a certain strategy,like zone and region etc.

zoemccormick:
@jehiah @ploxiln @mreiferson This PR is now officially ready for review - in conjunction with [nsqio/go-nsq#](https://github.com/nsqio/go-nsq/pull/312) and [nsqio/nsqio.github.io#89](https://github.com/nsqio/nsqio.github.io/pull/89).

We have written up an experience report based on our observations running these changes for 2 months in our production environment - it can be found [here](https://docs.google.com/document/d/1YoMVmStoRjLGJ0SREVsQl3yNKxyjVyIQ2lXnNFK8jKo/edit?usp=sharing). It also lays out the changes to nsqadmin.

Please let me know if you have any other questions! Thanks in advance!

apps/nsqd/options.go
```diff
@@ -196,5 +198,12 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Int("max-deflate-level", opts.MaxDeflateLevel, "max deflate compression level a client can negotiate (> values == > nsqd CPU usage)")
 	flagSet.Bool("snappy", opts.SnappyEnabled, "enable snappy feature negotiation (client compression)")
 
+	experiments := app.StringArray{}
+	var validExperiments []string
+	for _, e := range nsqd.AllExperiments {
+		validExperiments = append(validExperiments, fmt.Sprintf("'%s'", string(e)))
```
jehiah, line 204:
```suggestion
		validExperiments = append(validExperiments, fmt.Sprintf("%q", string(e)))
```

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/stretchr/testify v1.7.0 // indirect
```
jehiah, line 21: Let's bump this dependency to v1.9.0

nsqadmin/static/js/views/channel.hbs
```diff
@@ -92,6 +93,24 @@
...
+                            <span class="sr-only">{{floatToPercent zone_local_percentage}}%</span>
+                        </div>
+                        <div class="progress-bar progress-bar-warning progress-bar-striped" style="background-color:rgb(254,254,194); width: {{floatToPercent region_local_percentage}}%">
+                            <span class="sr-only">{{floatToPercent region_local_percentage}}%</span>
+                        </div>
+                        <div class="progress-bar progress-bar-danger" style="background-color: white; width: {{floatToPercent global_percentage}}%">
+                            <span class="sr-only">{{floatToPercent global_percentage}}%</span>
+                        </div>
+                    </div>
+                    <span class="popuptext" id="popup-{{node}}"><span style="background-color:rgb(221,255,221); color:#4b4b4b">{{floatToDecimalPercent zone_local_percentage}}%</span> | <span style="background-color:rgb(254,254,194); color:#4b4b4b">{{floatToDecimalPercent region_local_percentage}}%</span> | <span style="color:#4b4b4b">{{floatToDecimalPercent global_percentage}}%</span></span>
```
jehiah, line 110: It's probably worth putting these in `base.css` as classes `bg-zone-local`, `bg-region-local` and `bg-global-zone` to this reads a little clearer. (can be re-used on the progress bar too)

nsqd/channel.go
```diff
@@ -303,17 +330,52 @@ func (c *Channel) PutMessage(m *Message) error {
...
-		if err != nil {
-			c.nsqd.logf(LOG_ERROR, "CHANNEL(%s): failed to write message to backend - %s",
-				c.name, err)
-			return err
+	if c.topologyAwareConsumption {
+		// Attempt zone local, region local and finally the memory channel
+		// we do this to ensure that we preferentially deliver messages based on toplogy
+		//
+		// Because messagePump is intermittently unavailable while writing a msg to a client
+		// we continue to have higher priority channels in the select loop, this means at each
```
jehiah, line 338: small re-wording for clarity (i hope)

```suggestion
		// Because messagePump is intermittently unavailable while writing a msg to a client
		// we continue to add lower priority channels in the select loop, this means at each
```

Ulminator (approved):

mreiferson:
@jehiah hello! what's the plan here for actually landing this? Are y'all waiting on me to review everything 😏?

jehiah:
@mreiferson sorry for the radio silence on this - I think we are ready to give more attention to this in Q4 so if you have any comments on this PR (or the related go-nsq changes) in the next few weeks please share them otherwise the topology changes are ready to land and we'll do that next month.

I think the general plan is to still land this behind the feature flag and get them in a 1.4 release and consider removing the FF in 1.5. We have been happy with this running at Bitly for a while now, but i'm keeping an eye on kubernetes/enhancements#4747 as it will make adopting in a K8s environment easier.

martin-sucha:
I tried this in our deployment and it enabling the experiment didn't have the intended effect. I didn't see any effect on cross-zone traffic after enabling it, messages were sent to all zones.

I changed the `zoneLocalMsgChan` to have a large capacity (30k), then I observed an effect on consumer clients.

However, I observed that the message routing probably interfered with how concurrency is managed, as our consumer pods scaled down to almost zero after a few minutes. Since the nsq client divides the concurrency (RDY) between the nsqd instances, it seems to me that 2/3 of the in-flight messages that would be cross-zone traffic otherwise are not used for sending messages.

So perhaps the go-nsq client should divide the RDY requests only by the number of nsqd instances in the current zone? Resp. divide MaxInFlight for each zone separately?

What value of `nsq.Config.MaxInFlight` do you usually use with the topology-aware consumption? We have low values (7-10) per pod.

Issue #1303, opened 2020-11-24
nsq_tail: add -gunzip option


3dbrows:
Optionally gunzip messages in nsq_tail. Useful if your messages are gzipped and you want to human-read them. If you set `-gunzip` and the messages are not gzipped, it continues to work.

3dbrows:
PR is ready for review: #1304 .

mreiferson:
see #1304

PR #1304, opened 2020-11-24
nsq_tail: added -gunzip option

3dbrows:
* if the -gunzip option is set, an attempt is made to gunzip the message before output
* if the -gunzip option is set but the message was not gzipped, do not alter the message

### DIFF
--- a/apps/nsq_tail/nsq_tail.go
+++ b/apps/nsq_tail/nsq_tail.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"bytes"
+	"compress/gzip"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


3dbrows:
Ready for review. Please see #1303

jehiah:
@3dbrows can you give some context of why or when this would be preferable vs sending an un-gzipped message body? NSQ provides gzip (or snappy) compression on the transport level by default which minimizes the need to manually compress message body in most cases.

mreiferson:
Granted that `nsq_tail` does make some assumptions about the encoding of messages (line oriented text), however I don't think that we want to start baking in other formats.

Thanks for the proposed contribution, gonna close.

3dbrows:
@jehiah I hadn't appreciated that NSQ compresses by default, thanks for pointing that out. In that case it is unnecessary for me to manually compress at the application level. 👍

PR #1307, opened 2020-11-26
*: remove *Context

jehiah:
nsqd use of `*Context` and `*context` predates the stdlib `context.Context` and is now confusing; It was used to qualify references to the application struct often used for accessing configuration. This refactors to just remove the wrapper structs which doesn't significantly reduce clarity of code (and improves it in light of stdlib context)

This refactor is needed to import stdlib `context` anywhere in `nsqd/...` without naming conflicts.

### DIFF
--- a/nsqadmin/context.go
+++ b/nsqadmin/context.go
@@ -1,5 +1,0 @@
-package nsqadmin
-
-type Context struct {
-	nsqadmin *NSQAdmin
-}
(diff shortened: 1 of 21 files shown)


mreiferson (approved):
Yep, this seems reasonable due to the naming conflict. I suppose this means that we'll have to hang anything we want to plumb through on `struct nsqd`, etc. Given we've never done that in nearly a decade, LGTM.

FYI there's a typo in the commit msg for `nsqlookupd` (two k's)

PR #1319, opened 2021-02-02
update go-svc

jehiah:
This extracts `go-svc` changes out of #1305 and to update the version `go-svc` we import.

### DIFF
--- a/apps/nsqadmin/main.go
+++ b/apps/nsqadmin/main.go
@@ -9,7 +9,7 @@ import (
 	"syscall"
 
 	"github.com/BurntSushi/toml"
-	"github.com/judwhite/go-svc/svc"
+	"github.com/judwhite/go-svc"
 	"github.com/mreiferson/go-options"
 	"github.com/nsqio/nsq/internal/app"
 	"github.com/nsqio/nsq/internal/lg"
--- a/apps/nsqd/main.go
+++ b/apps/nsqd/main.go
@@ -1,17 +1,19 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"math/rand"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
 
 	"github.com/BurntSushi/toml"
-	"github.com/judwhite/go-svc/svc"
+	"github.com/judwhite/go-svc"
 	"github.com/mreiferson/go-options"
 	"github.com/nsqio/nsq/internal/lg"
 	"github.com/nsqio/nsq/internal/version"
@@ -25,7 +27,8 @@ type program struct {
 
 func main() {
 	prg := &program{}
-	if err := svc.Run(prg, syscall.SIGINT, syscall.SIGTERM); err != nil {
+	// SIGTERM handling is in Start()
+	if err := svc.Run(prg, syscall.SIGINT); err != nil {
 		logFatal("%s", err)
 	}
 }
@@ -62,6 +65,7 @@ func (p *program) Start() error {
 	cfg.Validate()
 
 	options.Resolve(opts, flagSet, cfg)
+
 	nsqd, err := nsqd.New(opts)
 	if err != nil {
 		logFatal("failed to instantiate nsqd - %s", err)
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -8,12 +10,9 @@ require (
 	github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/snappy v0.0.1
-	github.com/judwhite/go-svc v1.1.2
+	github.com/judwhite/go-svc v1.2.0
```
jehiah, line 13: blocked on judwhite/go-svc#27

jehiah:
RFR @mreiferson @ploxiln

apps/nsqadmin/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"syscall"
 
 	"github.com/BurntSushi/toml"
-	"github.com/judwhite/go-svc/svc"
+	svc "github.com/judwhite/go-svc"
```
judwhite, line 12: @jehiah I don't think you need the import alias. I'm not using it in the example project

https://github.com/judwhite/go-svc/blob/a2deda217eb3db0e1264c3386d5e41f8c8b08207/example/main.go#L10

apps/nsqd/main.go
```diff
@@ -1,17 +1,19 @@
...
 	"os"
+	"os/signal"
 	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
 
 	"github.com/BurntSushi/toml"
-	"github.com/judwhite/go-svc/svc"
+	svc "github.com/judwhite/go-svc"
```
judwhite, line 16: Same as above

apps/nsqlookupd/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"syscall"
 
 	"github.com/BurntSushi/toml"
-	"github.com/judwhite/go-svc/svc"
+	svc "github.com/judwhite/go-svc"
```
judwhite, line 12: Same here

apps/nsqadmin/main.go
jehiah: hmmm by convention i expect aliases when the package names doesn't match the directory name (i.e. `svc` and `go-svc`) but clearly there is more to `goimports` logic around this that I don't understand because it doesn't automatically add this alias (but leaves it if it's there). 

I think either approach here is ok 🤷‍♂️

apps/nsqadmin/main.go
```diff
@@ -9,7 +9,7 @@ import (
...
 	"github.com/mreiferson/go-options"
```
judwhite, line 13: @jehiah You know, it was the early days. Every package started with `go-` 😆

jehiah: goodness knows I have a bunch of those myself 😢

judwhite: @jehiah Sometimes I forget to mention what I'm trying to point out 🤦‍♂️ `go-options` (the line I clicked 'Comment' on) is used like `options`. It's no big deal, just pointing it out for consistency.

jehiah: @judwhite That's a good point (and consistency is a compelling argument)

What i didn't know when i wrote this (it took me too long to track this down) is that `goimports` does special case the `"go-"` prefix when handling import paths; I didn't know that.

https://github.com/golang/tools/blob/b894a3290fff7ed8373c3156460600f8216a6c2d/internal/imports/fix.go#L1128-L1151

apps/nsqd/main.go
```diff
@@ -77,6 +81,17 @@ func (p *program) Start() error {
...
+	signalChan := make(chan os.Signal, 1)
+	go func() {
+		// range over all term signals
+		// we don't want to un-register our sigterm handler which would
+		// cause default go behavior to apply
+		for range signalChan {
+			p.nsqd.TermSignal()
+		}
+	}()
+	signal.Notify(signalChan, syscall.SIGTERM)
```
mreiferson, line 93: This still feels odd to me that we have to separate out a signal to get the behavior we want. Feels like `go-svc` should be able to handle multiple signals, and "programs" can implement a callback act on a received signal and return something to `go-svc` indicating whether it should shut down?

PR #1323, opened 2021-02-23
Added support for writing messages to files in nsq_tail

crazyweave:
When trying to decode gRPC messages embedded in the NSQ message, there can be embedded NULL and newline characters which makes capturing the messages from stdout impossible.

This PR adds two new command-line switches:
* --write-to-files (default: false)
* --filePrefix (default: "message")

The file prefix is the name to use for the files when writing them to disk.  The name is `<file-prefix>-<msg-number>.bin`.

### DIFF
--- a/apps/nsq_tail/nsq_tail.go
+++ b/apps/nsq_tail/nsq_tail.go
@@ -22,6 +22,8 @@ var (
 	maxInFlight   = flag.Int("max-in-flight", 200, "max number of messages to allow in flight")
 	totalMessages = flag.Int("n", 0, "total messages to show (will wait if starved)")
 	printTopic    = flag.Bool("print-topic", false, "print topic name where message was received")
+	writeToFiles  = flag.Bool("write-to-files", false, "prints messages to file - one file per message")
+	filePrefix    = flag.String("file-prefix", "message", "filename prefix when printing to file")
 
 	nsqdTCPAddrs     = app.StringArray{}
 	lookupdHTTPAddrs = app.StringArray{}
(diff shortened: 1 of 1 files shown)


crazyweave:
@jehiah or @mreiferson -> Would you take a look at this?

jehiah:
@crazyweave could you solve for your need with`nsq_to_file` ? 

I don't think we would add any built in capabilities to write to a file to `nsq_tail` as that's already the purpose of `nsq_to_file`, but I think we'd be willing to add a flag that disabled newlines in the output.

crazyweave:
@jehiah I was in a rush to deal with an issue and missed that utility. 🤦 

Looks like it writes all messages to a single file which doesn't help my use case.  I might be able to reasonably adapt it to my need to write to multiple files.  Let me noodle on that.  Thanks.

jehiah:
@crazyweave no worries; You mentioned using `gRPC` messages, and i'll acknowledge that some of the tooling binaries (`nsq_tail`, `nsq_to_file`) are more oriented to newline deliminated streams, but I don't see any reason we can't add arguments to make that optional for binary oriented datastreams.  

The general expectation is that those tools generally don't have much value for binary oriented datastreams and in those cases you'd always be using `go-nsq` or a similar client library to handle messages, but if there are ways to make the built in utilities more useful i'm all ears.

ploxiln:
I think it's fine if you modify nsq_tail, strip out parts you don't need and add your file-per-message feature, optionally post a link to it in the mailing list, and just use it. It's just 140 lines :) Custom nsq consumer services are pretty common.

PR #1329, opened 2021-03-17
nsqadmin: add option to skip resolving on startup

martin-sucha:
When running nsqlookupd in Kubernetes and pointing to a specific
identity of a pod from statefulset, the DNS name is not resolvable
if the pod is not currently scheduled.

The lookup check currently causes nsqadmin to exit during startup
even if only one of the listed nsqlookupd instances is not resolvable.

I want nsqadmin to always start. If no lookupd instances are currently
resolvable/reachable that fact will be visible in the user interface.


martin-sucha:
ready for review

mreiferson:
Hey @martin-sucha thanks for the PR.

This code path ultimately just validates that we have a proper hostname or IP address - I agree that the _resolution_ of the hostname is probably not something we need to fatally exit on.

So, I don't think the answer here is a new command line flag, but rather validation of the format and not the name resolution.

mreiferson (changes requested):
As per above.

PR #1330, opened 2021-03-18
*: fix invalid key in example cfg file and support log_level from cfg file

edoger:


edoger:
After verification, the configuration of the log level "log-level" in the configuration file does not take effect. This is because the third-party library (github.com/mreiferson/go-options) can only recognize the "log_level". This patch fixes the configuration file Use "log_level" to set the log level. For "log-level", the old application behavior is still maintained.

nsqadmin/options.go
```diff
@@ -59,3 +62,23 @@ func NewOptions() *Options {
 		AdminUsers:               []string{},
 	}
 }
+
+func (opts *Options) Resolve(flagSet *flag.FlagSet, cfg map[string]interface{}) error {
```
mreiferson, line 66: We already have a way to coerce config data into appropriate option types, see:

https://github.com/nsqio/nsq/blob/master/apps/nsqd/options.go#L69

So we don't actually need this new method for any of the binaries.

apps/nsqd/main.go
```diff
@@ -64,7 +63,9 @@ func (p *program) Start() error {
 	}
 	cfg.Validate()
 
-	options.Resolve(opts, flagSet, cfg)
+	if err := opts.Resolve(flagSet, cfg); err != nil {
```
mreiferson, line 66: This can be reverted as well after addressing my previous comment. For better or worse, `options.Resolve` fatally exits on errors.

mreiferson (changes requested):
Thanks for this comprehensive PR @edoger, a few changes if you don't mind!

nsqadmin/options.go
edoger: @mreiferson Do we need to migrate the Validate method to Resolve? Because nsqadmin and nsqlookupd have no corresponding validate methods.

mreiferson: No, I think each binary should have a config file validation method rather than moving that responsibility into the package / options.

edoger:
@mreiferson Done! The test failure doesn't seem to be caused by the code?

mreiferson:
Changes look great, tests are failing due to `go fmt` issues. Would you mind fixing those and squashing this down to logical commits?

edoger:
@mreiferson Done! The code style has been fixed and everything looks good. Since the logFatal function cannot be unit tested, the decrease in test coverage is an inevitable result.

PR #1333, opened 2021-03-28
*: test script failing to run -race

mreiferson:
```
./test.sh: 6: [: amd64: unexpected operator
./test.sh: 6: [: arm64: unexpected operator
```

See this in every test run, which means we're never running with `-race`

cc @ploxiln

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -3,7 +3,7 @@ set -e
 
 GOMAXPROCS=1 go test -timeout 90s ./...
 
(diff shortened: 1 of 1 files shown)


ploxiln (approved):
ah, yes, `==` is a bashism, posix sh specifies only `=` in this context, and `/bin/sh` on debian/ubuntu is dash (minimal posix shell) rather than bash

Issue #1336, opened 2021-04-11
Give the ability to configure auth endpoint


Misiu:
Hi there,
Currently when using `-auth-http-address=host:port` the endpoint is hardcoded to `/auth` 

https://github.com/nsqio/nsq/blob/c164c776ab1af72706fff6cc05cd80a36771811d/internal/auth/authorizations.go#L108

I have an identity server in my solution that is already using this endpoint. I'd like to be able to customize the URL, for example, I'd like to change it from `/auth` to `/nsq/auth`.

Also, I noticed that the docs don't describe the `common_name` parameter that can be used with auth.

mreiferson:
seems reasonable to me, would you like to submit a PR?

Misiu:
@mreiferson sadly I don't know GO at all, so it might be hard for me to set up the dev environment and do that small change in a correct way.
Maybe instead of adding another parameter, we could reuse `auth-http-address`, so if someone will enter a value containing `/` then we would do:
`endpoint := fmt.Sprintf("http://%s?%s", authd, v.Encode()) `

So we should check if the value entered in that parameter contains `/` and then do a proper replacement.

The second option would be to add `auth-http-endpoint`, but as I wrote I have no GO experience, so I can't say what option would be better.

tufanbarisyildirim:
Hi @Misiu setting the option the `-auth-http-address` option sth like `-auth-http-address=auth.example.com/nsq` worked for us, give it try maybe?. I also think, just checking the `http` prefix and not touching if it is already a url and just adding query params would be a good option. I would like to send a PR if that makes sense @mreiferson

mreiferson:
@tufanbarisyildirim sure, feel free to open a PR, thanks!

PR #1341, opened 2021-05-02
ability to configure auth endpoint path

tufanbarisyildirim:
added ability to provide full url as auth endpoint #1336


tufanbarisyildirim:
> What about https? I think it should be supported. This solution will work if we specify full url that starts with 'http'

`https` starts with `http`, so, both should work. @Misiu

Misiu:
> `https` starts with `http`, so, both should work. @Misiu

it's definitively too late for a review.
Sorry for the confusion. I can confirm that HTTPS starts with HTTP :)

tufanbarisyildirim:
> Yeah, https should work fine.
> 
> But another possible objection is that a hostname could start with "http", like `httpcluster01.company.org`, so perhaps `strings.Contains(authd, "://")` would be a more precise condition.

yeah, good point, you add and commit it quickly maybe? since this is merged already.

ploxiln:
Do you want to open a new pull request with that change 😁 
I think it's fine to take a day or two, because this "feature" is not in a release yet.

PR #1343, opened 2021-05-12
look for schema separator instead of http on checking auth endpoint

tufanbarisyildirim:
https://github.com/nsqio/nsq/pull/1341#issuecomment-839065228

### DIFF
--- a/internal/auth/authorizations.go
+++ b/internal/auth/authorizations.go
@@ -107,7 +107,7 @@ func QueryAuthd(authd string, remoteIP string, tlsEnabled bool, commonName strin
 	v.Set("common_name", commonName)
 
 	var endpoint string
-	if strings.HasPrefix(authd, "http") {
+	if strings.Contains(authd,"://"){
 		endpoint = fmt.Sprintf("%s?%s", authd, v.Encode())
 	} else {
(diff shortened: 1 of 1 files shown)


internal/auth/authorizations.go
```diff
@@ -107,7 +107,7 @@ func QueryAuthd(authd string, remoteIP string, tlsEnabled bool, commonName strin
 	v.Set("common_name", commonName)
 
 	var endpoint string
-	if strings.HasPrefix(authd, "http") {
+	if strings.Contains(authd,"://"){
```
ploxiln, line 110: minor formatting issue - needs space after comma

ploxiln, line 110:
```suggestion
	if strings.Contains(authd, "://"){
```

internal/auth/authorizations.go
```diff
@@ -107,7 +107,7 @@ func QueryAuthd(authd string, remoteIP string, tlsEnabled bool, commonName strin
 	v.Set("common_name", commonName)
 
 	var endpoint string
-	if strings.HasPrefix(authd, "http") {
+	if strings.Contains(authd, "://"){
```
ploxiln, line 110:
```suggestion
	if strings.Contains(authd, "://") {
```

PR #1344, opened 2021-05-14
Always keep last messages in memory for ephemeral topic/channel

a180285:
Always keep last messages in memory for ephemeral topic/channel


mreiferson:
@a180285 thanks for submitting the PR, however I don't think this is the right place to attempt retries before discarding a message on an ephemeral channel. If consumers can't keep up with the volume, you could adjust your mem queue size instead.

PR #1346, opened 2021-05-20
nsqd: configurable to skip ephemeral topics/channels in statsd output

jehiah:
When using nsqd with statsd/graphite ephemeral topics and channels can dramatically increase the number of datasets on the graphite side. This creates an option to filter out those from being written to graphite)

I thought this was existing behaviour but discovered this issue due to a new usage pattern that used a dynamic and frequently changing ephemeral channel name. In my case the `#` was being stripped out of key names and so I had missinterpreted missing graphs in nsqadmin for ephemeral channels as no data in graphite, when it was actually a missmatch between the query and resulting dataset key.

To provide a way to filter out ephemeral topics and channels this adds a new `--statsd-exclude-ephemeral` argument which defaults to false (existing behavior).

### DIFF
--- a/nsqd/statsd.go
+++ b/nsqd/statsd.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/nsqio/nsq/internal/statsd"
@@ -36,6 +37,7 @@ func (n *NSQD) statsdLoop() {
 		case <-ticker.C:
 			addr := n.getOpts().StatsdAddress
 			prefix := n.getOpts().StatsdPrefix
+			excludeEphemeral := n.getOpts().StatsdExcludeEphemeral
 			conn, err := net.DialTimeout("udp", addr, time.Second)
 			if err != nil {
 				n.logf(LOG_ERROR, "failed to create UDP socket to statsd(%s)", addr)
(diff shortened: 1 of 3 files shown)


nsqd/statsd.go
```diff
@@ -80,6 +86,10 @@ func (n *NSQD) statsdLoop() {
 				}
 
 				for _, channel := range topic.Channels {
+					if excludeEphemeral && strings.HasSuffix(channel.ChannelName, "#ephemeral") {
```
ploxiln, line 89: I think you can just use `channel.ephemeral` and `topic.ephemeral` boolean properties

ploxiln:
> the # was being stripped out of key names and so I had missinterpreted missing graphs in nsqadmin

Is this a bug in nsqd/nsqadmin, should we also try to fix this? (separately would be fine)

nsqd/statsd.go
jehiah: These are `TopicStats` and `ChannelStats` structs respectively so those don't have the `ephemeral` field which is on the actual `Topic` and `Channel` struct.

We do a number of other suffix matches to find this so this seemed like the best path vs expanding scope to include in `TopicStats` and `ChannelStats` as those represent response structs from stats endpoints.

ploxiln: ah got it

ploxiln (approved):

jehiah:
> > the # was being stripped out of key names and so I had missinterpreted missing graphs in nsqadmin
>
> Is this a bug in nsqd/nsqadmin, should we also try to fix this? (separately would be fine)

It looks like this is happening in `bitly/statsdaemon` (i'm not positive what other `statsd` implementations do)  so it does seem reasonable to match that behavior in `nsqadmin` by default (i'll do that separately).

https://github.com/bitly/statsdaemon/blob/fc46d9cfe29b674a0c8abc723afaa9370430cdcd/statsdaemon.go#L64-L88

Issue #1317, opened 2021-01-21
*: migrate to go:embed from go-bindata


jehiah:
With go 1.16 the Go language supports embedding files in a binary; It is therefore desirable to switch from using `go-bindata` for embedding nsqadmin resources to the stdlib approach with `go:embed` when we are ready to drop support for Go 1.15

https://tip.golang.org/pkg/embed/

mreiferson:
see #994

mreiferson:
done in #1347

PR #1347, opened 2021-05-21
nsqadmin: use go:embed; match graphite key valid charset

jehiah:
This updates `nsqadmin` to query graphite using the correct key names for ephemeral topics and channels (see #1346)


nsqadmin/static/js/lib/handlebars_helpers.js
```diff
@@ -61,17 +61,24 @@ var genColorList = function(typ, key) {
     return 'blue';
 };
 
+// sanitizeGraphiteKey removes special characters from a graphite key
+// this matches behavior of bitly/statsdaemon
+// https://github.com/bitly/statsdaemon/blob/fc46d9cfe29b674a0c8abc723afaa9370430cdcd/statsdaemon.go#L64-L88
+var sanitizeGraphiteKey = function(s) {
+    return s.replaceAll(/[/ ]/g, '-').replaceAll(/[^a-zA-Z0-9-_.]/g, '')
```
ploxiln, line 68: space should be replaced with underscore, I think

ploxiln:
interesting related PR for original statsd: https://github.com/statsd/statsd/pull/155

nsqadmin/static/js/lib/handlebars_helpers.js
ploxiln: (rephrased less confusingly: this replaces ` ` with `-`, but statsdaemon replaces ` ` with `_`)

ploxiln: I think mreiferson's update changed this to replace both `/` and ` ` with `_`, but `/` still should be replaced with `-` :)

mreiferson:
OK, this also addresses #1317

mreiferson:
heh, now the build pipeline requires the previously temporary `static/build` output files - perhaps we just check them in, so we can see exactly what the inputs are for a given commit?

mreiferson:
also, lost "debug mode" from `go-bindata`, probably something we'll need to reimplement

nsqadmin/static/js/lib/handlebars_helpers.js
```diff
@@ -61,17 +61,24 @@ var genColorList = function(typ, key) {
     return 'blue';
 };
 
+// sanitizeGraphiteKey removes special characters from a graphite key
+// this matches behavior of bitly/statsdaemon
+// https://github.com/bitly/statsdaemon/blob/fc46d9cfe29b674a0c8abc723afaa9370430cdcd/statsdaemon.go#L64-L88
+var sanitizeGraphiteKey = function(s) {
+    return s.replaceAll(/[ ]/g, '_').replaceAll('/[/]/g', '-').replaceAll(/[^a-zA-Z0-9-_.]/g, '');
```
ploxiln, line 68: the quotes around `'/[/]/g'` make it not work:
js_replace_ex

That's because the first argument to `replaceAll()` can either be a regex obj or a string (and the quotes make it a string, not a regex). But this suggests a way to simplify these calls:

```js
s.replaceAll(' ', '_').replaceAll('/', '-')
```

replace_simple_ex

ploxiln:
> now the build pipeline requires the previously temporary static/build output files - perhaps we just check them in

Yeah, I think we should. It shouldn't be *too* much, in fact it should be equivalent to `bindata.go` but a bit nicer. (Otherwise one has to install the npm dep tree and do the js build, just to build and test some changes to the go code, if using go-1.16+)

mreiferson:
OK, think that should do it, PTAL!

ploxiln (approved):

Issue #1306, opened 2020-11-26
docker: publish ARM images


duanhongyi:
Now arm is used more and more widely. At present, many projects have docker images of arm, such as [Nats](https://hub.docker.com/_/nats). I hope that maintainers can add arm related images in docker hub.

danbf:
![giphy (5)](https://user-images.githubusercontent.com/5041307/126371995-4d16f50d-79fd-4102-a70e-0bc7a2b9fa4e.gif)

this may be as simple as changing over to `buildx` in the image build chain ala'

```
docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v7
```
https://www.docker.com/blog/multi-arch-images/

i might be able to write a patch for this if someone points to the relevant section

danbf:
is it changing https://github.com/nsqio/nsq/blob/master/dist.sh#L45 to this gonna do it?

```
docker buildx build -t nsqio/nsq:v$version . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
```

danbf:
@mreiferson ^^

danbf:
actually since buildx needs some setup, i do this on my end for multi-arch's:

ensure i cleared out the buildx builder in case it got interrupted before
```
docker buildx rm <builder-name>
```
then:
```
docker buildx create --name <builder-name>
docker buildx use <builder-name>
docker buildx build --tag <remote_tag> . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
```
then clear out the buildx builder so it's ready for the next time i run it
```
docker buildx rm <builder-name>
```

danbf:
i have it running in `sso` like this https://github.com/buzzfeed/sso/pull/319 so i'll take a crack at this

mreiferson:
looking forward to it @danbf

mreiferson:
see #1355

PR #1355, opened 2021-08-03
*: build multi-arch docker images linux/amd64,linux/arm64

danbf:
## Problem

current nsq images don't support arm64 architecture.  that means we can't run them on AWS Graviton's or Apple M1's natively.

## Solution

use buildx to build a multi-arch with `linux/amd64,linux/arm64` platform support.

the biggest change is we have to go from separate build and push steps into a single build and push step.  this is because the local docker doesn't support building locally multi-arch images.

## Notes

based on https://www.docker.com/blog/multi-arch-images/ and https://www.docker.com/blog/multi-arch-build-what-about-circleci/
our upstreams, both `golang:latest` and  alpine:3.10`` support these architectures and more.

addresses: https://github.com/nsqio/nsq/issues/1306

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -42,8 +42,11 @@ for os in linux darwin freebsd windows; do
     sudo rm -r $BUILD
 done
 
-docker build -t nsqio/nsq:v$version .
+docker buildx create --name nsq
+docker buildx use nsq
+docker buildx build --tag nsqio/nsq:v$version . --platform linux/amd64,linux/arm64 --push
 if [[ ! $version == *"-"* ]]; then
     echo "Tagging nsqio/nsq:v$version as the latest release."
-    docker tag nsqio/nsq:v$version nsqio/nsq:latest
+    docker buildx build --tag nsqio/nsq:latest . --platform linux/amd64,linux/arm64 --push
 fi
+docker buildx rm nsq


danbf:
running into a bit of an issue with the arm7 build. need to dig.

```
 => [linux/arm/v7 build 5/5] RUN export GO111MODULE=on  && ./test.sh  && CGO_ENABLED  5478.4s
 => => # building apps/nsq_to_nsq
 => => # building apps/nsqadmin
 => => # building apps/nsqd
 => => # building apps/nsqlookupd
 => => # building apps/to_nsq
 => => # building bench/bench_channels
 ```

File: dist.sh [hunk 1]
```diff
@@ -2,15 +2,14 @@
...
-# 8. update nsqio/nsqio.github.io/_posts/2014-03-01-installing.md
-# 9. send release announcement emails
-# 10. update IRC channel topic
-# 11. tweet
+# 5. push to nsqio/master
+# 6. update the release metadata on github / upload the binaries there too
+# 7. update nsqio/nsqio.github.io/_posts/2014-03-01-installing.md
+# 8. send release announcement emails
+# 9. update IRC channel topic
+# 10. tweet
```
mreiferson, line 12: Can we revert these superfluous text changes in the doc?

mreiferson (changes requested):
I'll give this a test run, seems like a good time for a stable release ;)

dist.sh
danbf: 👍 they are now gone.

mreiferson (approved):

PR #1361, opened 2021-08-15
apps/nsqd: initialize nsqd in Init(); move signal handling to go-svc

mreiferson:
This continues to clean up from #1331 and also fixes #1359

### DIFF
--- a/apps/nsqd/main.go
+++ b/apps/nsqd/main.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
-	"os/signal"
-	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


mreiferson:
Depends on https://github.com/judwhite/go-svc/pull/28

apps/nsqd/main.go
```diff
@@ -72,7 +62,15 @@ func (p *program) Start() error {
 	}
 	p.nsqd = nsqd
 
-	err = p.nsqd.LoadMetadata()
+	if env.IsWindowsService() {
+		dir := filepath.Dir(os.Args[0])
+		return os.Chdir(dir)
```
ploxiln, line 67: This used to be before `nsqd.New()` (which uses `os.Getwd()` for default datapath), also before opening the config file ... this only affects windows services, but still might want to keep their behavior consistent with before, I dunno.

mreiferson: Good catch

ploxiln (approved):
This looks good to me.

We'll need some go-svc updates for it, obviously :)

apps/nsqd/main.go
```diff
@@ -27,21 +26,20 @@ type program struct {
...
+	if err := svc.Run(prg, syscall.SIGINT, syscall.SIGTERM); err != nil {
 		logFatal("%s", err)
 	}
 }
 
 func (p *program) Init(env svc.Environment) error {
 	if env.IsWindowsService() {
 		dir := filepath.Dir(os.Args[0])
-		return os.Chdir(dir)
+		err := os.Chdir(dir)
```
ploxiln, line 37: ... it turns out that go-svc (now?) has this exact code:
https://github.com/judwhite/go-svc/blob/main/svc_windows.go#L69-L76

mreiferson: Nice, I'll drop this

mreiferson:
I've rebased and added acc2061, which will allow us to make progress on a build without waiting for upstream go-svc.

PTAL

ploxiln (approved):
LGTM

PR #1362, opened 2021-08-15
Dockerfile: update base image from alpine:3.10 to latest

ploxiln:
we can just use latest stable alpine release (currently 3.14)

also switch build image from golang default (debian-based)
to alpine-based, it's smaller and faster to download

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM golang:latest AS build
+FROM golang:alpine AS build
+
+RUN apk update && apk add make gcc musl-dev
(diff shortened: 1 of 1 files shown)


mreiferson (approved):
LGTM

ploxiln:
thanks for reviewing!

I pushed up another Dockerfile tweak just now, sorry :)

fwiw I've tested docker builds with `--platform=linux/amd64` and `--platform=linux/arm64`

mreiferson:
👍

PR #1365, opened 2021-08-16
dist: don't push images

mreiferson:
The expected behavior is for `dist.sh` to produce a set of local images, to be able to separately push (which matches the behavior of what we do for `.tar.gz` files)

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -3,10 +3,10 @@
 # 1. commit to bump the version and update the changelog/readme
 # 2. tag that commit
 # 3. use dist.sh to produce tar.gz for all platforms
-# 4. upload *.tar.gz to bitly s3 bucket
-# 5. docker push nsqio/nsq
+# 4. aws s3 cp dist s3://bitly-downloads/nsq/ --recursive --include "nsq-1.2.1*" --profile bitly --acl public-read
+# 5. docker manifest push nsqio/nsq:latest
 # 6. push to nsqio/master
-# 7. update the release metadata on github / upload the binaries there too
+# 7. update the release metadata on github / upload the binaries
 # 8. update nsqio/nsqio.github.io/_posts/2014-03-01-installing.md
 # 9. send release announcement emails
 # 10. update IRC channel topic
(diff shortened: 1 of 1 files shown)


ploxiln:
I think the problem is that you can't have both architectures of the same image tag tracked by the local docker daemon, so buildx has to push them as it makes them - you can't push both architectures later.

The docker buildx thing could be a separate even shorter script though.

(... I'm currently testing on a debian-11 host system with docker 20.10.5 from the base repos, buildx 0.6.1 plugin manually installed, and qemu 5.2.0 from the base repos with qemu-user-binfmt, and the multi-platform-at-once build fails strangely in some step in the middle, but one-at-a-time works, including for arm64. I'm guessing docker-for-mac works fine doing both concurrently.)

ploxiln:
btw it looks like latest tag on docker hub only has amd64 (while v1.2.1 tag has both amd64 and arm64)

mreiferson:
Gonna try this https://github.com/docker/buildx/issues/459#issuecomment-750011371

mreiferson:
Updated, also fixed the `latest` tag in Docker registry.

dist.sh
```diff
@@ -45,9 +45,10 @@ done
...
 rnd=$(LC_ALL=C tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c10)
 docker buildx create --use --name nsq-$rnd
-docker buildx build --tag nsqio/nsq:v$version . --platform linux/amd64,linux/arm64 --push
+docker buildx build --tag nsqio/nsq:v$version --platform linux/amd64,linux/arm64 .
 if [[ ! $version == *"-"* ]]; then
-    echo "Tagging nsqio/nsq:v$version as the latest release."
-    docker buildx build --tag nsqio/nsq:latest . --platform linux/amd64,linux/arm64 --push
+    echo "Tagging nsqio/nsq:v$version as the latest release"
+    shas=$(docker manifest inspect nsqio/nsq:$version |\
+        grep digest | awk '{print $2}' | sed 's/[",]//g' | sed 's/^/nsqio\/nsq@/')
```
ploxiln, line 52: this looks like a job for `jq` - but if it works for you it's fine with me

ploxiln (approved):

dist.sh
mreiferson: yes, of course. I wasn't quite sure what our policy for shell compatibility has been, although I suppose you already need quite a few dependencies (golang, docker).

danbf:
sorry for causing this noise, but https://github.com/nsqio/nsq/pull/1365#issuecomment-899161340 is on the money here.  locally you can't store the multi-arch builds :( so you have to buildx build and push in one step.

dist.sh
```diff
@@ -45,9 +45,10 @@ done
...
+    docker manifest create nsqio/nsq:latest $shas
 fi
-docker buildx rm nsq-$rnd
```
danbf, line 53: i think without the `buildx rm` the builders will pile up.  not sure how much resources they use.  

it might be possible to do a builder cleanup prior to setting up new builders.  for now on our systems if something happens and the builder gets orphan'd, i've just been removing it manually.

ploxiln: I think one could also skip the `create` and `use` steps, and just `buildx build` for local builds, right? It would just cause build cache to accumulate until you `docker system prune` or similar?

mreiferson: I like having a namespace to inspect as an artifact of the build. I don't expect this script to be used in any automated context, so practically speaking it's fine.

danbf: for local builds i think just doing an old school `docker build` is fine.  it's really only for the multi-arch push that we need to do the create/use/build-and-push chain.

PR #1368, opened 2021-08-20
reformat for go1.17

dmitris:
PR enables run `make` (build and test) with go1.17 by reformatting for go1.17.  `make` and `.test.sh` still work with go1.16.7.


mreiferson:
thanks

PR #1372, opened 2021-09-05
nsqadmin: new flag --dev-static-dir instead of debug build tag

ploxiln:
more flexible, perhaps simpler

-------------

it's subjective, let me know what you think :)


nsqadmin/static_debug.go
```diff
@@ -1,19 +0,0 @@
...
-package nsqadmin
-
-import (
-	"io/ioutil"
-	"os"
-	"path"
-)
-
-func staticAsset(name string) ([]byte, error) {
-	path := path.Join("../../nsqadmin/static/build", name)
```
mreiferson, line 13: I think the ugliness and problems likely resulting from this hard-coded path justify this change

mreiferson (approved):
LGTM

nsqadmin/static_debug.go
ploxiln: indeed, this was the motivation :)
thanks

PR #1380, opened 2021-10-19
nsqd: use metadata struct for both marshal and unmarshal

karalabe:
Metadata marshaling currently is done via a map containing untyped interfaces, which is unfortunate given that the code also contains a `meta` type used during unmarshaling.

This commit elevates the old `meta` type a bit to be used during both operations. The changeste also exposes a GetMetadata method to return the same metadata type without pushing it to disk. This is useful for use cases where NSQD is embedded into another Go program and some introspection is needed.

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -264,15 +264,23 @@ func (n *NSQD) Main() error {
 	return err
 }
 
-type meta struct {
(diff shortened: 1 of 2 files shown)


mreiferson:
updated this myself

mreiferson (approved):

PR #1385, opened 2021-10-20
nsqd: enable support for TLS1.3

karalabe:
Currently NSQD is capped at TLS1.2. TLS1.3 has been stabilized since Go 1.14, released almost 2 years ago. Might as well bump it.

### DIFF
--- a/apps/nsqd/options.go
+++ b/apps/nsqd/options.go
@@ -52,6 +52,8 @@ func (t *tlsMinVersionOption) Set(s string) error {
 		*t = tls.VersionTLS11
 	case "tls1.2":
 		*t = tls.VersionTLS12
+	case "tls1.3":
+		*t = tls.VersionTLS13
 	default:
 		return fmt.Errorf("unknown tlsVersionOption %q", s)
 	}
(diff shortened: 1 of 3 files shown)


karalabe:
Btw, I needed to make a small tweak to the protocol tests where the TLS handshake was tested.

In TLS1.2 and before, the handshake is a multi-RTT thing, where a lot of things are negotiated. If client certificates are used, those are sent over somewhere in the middle of the handshake and further things are negotiated afterwards. This means that a bad/missing client certificate will cause the handshake to fail.

In TLS1.3, the handshake was reduced to 3 messages: Client -> Server; Server -> Client; Client -> Server ([see code](https://github.com/golang/go/blob/master/src/crypto/tls/handshake_client_tls13.go#L93)). The client certificate is transmitted in the last Client -> Server message; at which point the TLS handshake completes. If the client certificate is bad, that rejection is not returned by the handshake, rather in the first application read. This meant that the code in the tests that relied on `tlsConn.Handshake` erroring stopped returning failures in TLS1.3. The only way to get the bad client cert error client side is to do an application read and have that fail. Hence my update to the tests.

karalabe:
Alternatively, I guess it would also be fine to try and read a single byte and have that fail; not to mix in the high level NSQ mixing and risk some other error popping up. I can also do that if you want, just followed the test as close as possible in this iteration.

PR #1405, opened 2022-05-06
nsqd: BroadcastAddress hostname->IP

needkane:
IP is more applicable than hostname, and it is unnecessary for the client machine to configure the hostname of nsqd.
There is a bug when use hostname as BroadcastAddress 
`
2022/05/05 17:09:52 INF    1 [test/nsq_to_file] querying nsqlookupd http://11.168.0.1:4161/lookup?topic=test
2022/05/05 17:09:52 INF    1 [test/nsq_to_file] (hostnameB:4150) connecting to nsqd
2022/05/05 11:51:22 ERR    1 [test/nsq_to_file] (hostnameB:4150) error connecting to nsqd - dial tcp: lookup hostnameB: no such host
`

Env
nsqlookup: 11.168.0.1 , hostnameA

nsqd: 11.168.0.2 , hostnameB

client: 11.168.0.3 , but /etc/hosts/ no hostnameB info

### DIFF
--- a/nsqd/options.go
+++ b/nsqd/options.go
@@ -3,9 +3,11 @@ package nsqd
 import (
 	"crypto/md5"
 	"crypto/tls"
+	"errors"
 	"hash/crc32"
 	"io"
 	"log"
+	"net"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nsqd/options.go
```diff
@@ -104,7 +136,7 @@ func NewOptions() *Options {
 		TCPAddress:        "0.0.0.0:4150",
 		HTTPAddress:       "0.0.0.0:4151",
 		HTTPSAddress:      "0.0.0.0:4152",
-		BroadcastAddress:  hostname,
+		BroadcastAddress:  localIP,
```
jehiah, line 139: While you might find IP a more applicable default in your situation others may find hostname more appropriate; Because this change would not be backwards compatible, and `--broadcast-address` is already exposed to make this configurable, we are unlike to make this change.

PR #1419, opened 2022-08-23
contrib: update nsqadmin.cfg.example

StellarisW:
update ACL and TLS usage for nsqadmin

### DIFF
--- a/contrib/nsqadmin.cfg.example
+++ b/contrib/nsqadmin.cfg.example
@@ -1,6 +1,41 @@
 ## log verbosity level: debug, info, warn, error, or fatal
 log_level = "info"
 
+## log message prefix (default "[nsqadmin] ")
+# log_prefix = ""
+
+## HTTP header to check for authenticated admin users (default "X_Forwarded_User")
+# acl_http_header = ""
+
+## admin user (may be given multiple times; if specified, only these users will be able to perform privileged actions)
+# admin_users = [
+#     "admin"
+# ]
+
+## A CIDR from which to allow HTTP requests to the /config endpoint (default "127.0.0.1/8")
+# allow_config_from_cidr = ""
+
+## URL base path (default "/")
+# base_path = ""
+
+## timeout for HTTP connect (default 2s)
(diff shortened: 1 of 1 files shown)


contrib/nsqadmin.cfg.example
```diff
@@ -1,6 +1,42 @@
...
+## log message prefix (default "[nsqadmin] ")
+# log_prefix = ""
+
+## HTTP header to check for authenticated admin users (default "X_Forwarded_User")
+# acl_http_header = ""
+
+## admin user (may be given multiple times; if specified, only these users will be able to perform privileged actions; acl_http_header is used to determine the authenticated user)
+# admin_users = [
+#     "admin",
+#     "asdqwc124134134e"
```
ploxiln, line 13: this is an odd example username

contrib/nsqadmin.cfg.example
```diff
@@ -1,6 +1,42 @@
 ## log verbosity level: debug, info, warn, error, or fatal
 log_level = "info"
 
+## log message prefix (default "[nsqadmin] ")
+# log_prefix = ""
+
+## HTTP header to check for authenticated admin users (default "X_Forwarded_User")
+# acl_http_header = ""
+
+## admin user (may be given multiple times; if specified, only these users will be able to perform privileged actions; acl_http_header is used to determine the authenticated user)
```
ploxiln, line 10: this help text doesn't work exactly as-is for this config format ("given multiple timpes"?)

also it should be wrapped somewhere around 80 chars depending on taste ;)

contrib/nsqadmin.cfg.example
StellarisW: i consider "admin_users" as a secret string actually, I'll change it.

ploxiln: Ah ... that's a clever way to use it! (As you probably already know, it is intended to be used behind an auth proxy, which is trusted to always set or clear this header value, like https://github.com/oauth2-proxy/oauth2-proxy)

PR #1424, opened 2022-09-15
nsqd: add memory release function

guozhao-coder:
In some cases where the message volume is large, the memory occupied by the program cannot be returned to the operating system immediately. You can call this method to enforce execution. However, using this method may cause the system to pause for a period of time, so use it with caution.

### DIFF
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"runtime/debug"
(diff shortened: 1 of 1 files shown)


ploxiln:
You could try setting the environment variable `GOGC=40` (or 20) to make Go GC more aggressive, might help with your issue.

ploxiln (approved):
seems like a reasonable debug endpoint to me

PR #1425, opened 2022-09-16
fix: idFactory.NewGUID() fails to retry

xiaoqinglee:
time.Sleep(time.Millisecond) can never be reached


mreiferson:
I don't understand the problem here, the modulo condition doesn't return it's only meant to reduce frequency of logging since this loop spins...

xiaoqinglee:
I was wrong

PR #1428, opened 2022-10-19
delete `validTopicChannelNameRegex` useless escape characters

sjatsh:


ploxiln (approved):
yup, inside a character class (`[]`), `.` does not need to be escaped

ploxiln:
We could probably afford a few trivial test cases for this regex, just to assure everyone ... but anyway I whipped up a minimal sandbox demo https://go.dev/play/p/WI5A9QGHMwM?v=goprev

PR #1434, opened 2022-12-23
Ability to bind unix sockets for nsqd

telepenin:
Hello, this PR provided unix socket supports.

nsqd may started with
```
/build/nsqd --tcp-address /var/run/nsqd.sock --http-address /var/run/nsqd-http.sock --data-path /var/local/nsqd/
```
I've adopted [pynsq](https://github.com/cloudlinux/pynsq/pull/1) and [go-nsq](https://github.com/cloudlinux/go-nsq/pull/1) for connecting to unix sockets as well. I'll create PRs after resolution.

### DIFF
--- a/apps/nsqd/options.go
+++ b/apps/nsqd/options.go
@@ -129,8 +129,9 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Bool("worker-id", false, "[deprecated] use --node-id")
 
 	flagSet.String("https-address", opts.HTTPSAddress, "<addr>:<port> to listen on for HTTPS clients")
-	flagSet.String("http-address", opts.HTTPAddress, "<addr>:<port> to listen on for HTTP clients")
-	flagSet.String("tcp-address", opts.TCPAddress, "<addr>:<port> to listen on for TCP clients")
+	flagSet.String("http-address", opts.HTTPAddress, "address to listen on for HTTP clients (<addr>:<port> for TCP/IP or <path> for unix socket)")
+	flagSet.String("tcp-address", opts.TCPAddress, "address to listen on for TCP clients (<addr>:<port> for TCP/IP or <path> for unix socket)")
+
 	authHTTPAddresses := app.StringArray{}
 	flagSet.Var(&authHTTPAddresses, "auth-http-address", "<addr>:<port> or a full url to query auth server (may be given multiple times)")
 	flagSet.String("broadcast-address", opts.BroadcastAddress, "address that will be registered with lookupd (defaults to the OS hostname)")
--- a/nsqd/http.go
+++ b/nsqd/http.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"net/url"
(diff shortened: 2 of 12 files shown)


apps/nsqd/options.go
```diff
@@ -131,6 +131,8 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.String("https-address", opts.HTTPSAddress, "<addr>:<port> to listen on for HTTPS clients")
 	flagSet.String("http-address", opts.HTTPAddress, "<addr>:<port> to listen on for HTTP clients")
 	flagSet.String("tcp-address", opts.TCPAddress, "<addr>:<port> to listen on for TCP clients")
+	flagSet.Bool("use-unix-sockets", opts.UseUnixSockets, "Usinx UNIX sockets instead of IP sockets")
```
ploxiln, line 134: Instead of adding a new flag, can you detect whether the `tcp-address` or `http-address` flags value is a unix socket path instead of an ip:port (and allow them to be tcp or unix socket individually)?

(should also update the help text to say that they can be "<addr>:<port> or unix /path" or similar)

nsqd/http.go
```diff
@@ -131,6 +131,14 @@ func (s *httpServer) doInfo(w http.ResponseWriter, req *http.Request, ps httprou
 	if err != nil {
 		return nil, http_api.Err{500, err.Error()}
 	}
+	tcpAddr, ok := s.nsqd.RealTCPAddr().(*net.TCPAddr)
+	if !ok {
+		return nil, http_api.Err{500, "failed to cast to TCPAddr"}
+	}
+	httpAddr, ok := s.nsqd.RealHTTPAddr().(*net.TCPAddr)
+	if !ok {
+		return nil, http_api.Err{500, "failed to cast to TCPAddr"}
```
ploxiln, line 140: does this mean that the nsqd info endpoint fails if unix socket is used? we can use port `-1` or similar for unix sockets

apps/nsqd/options.go
telepenin: Done

nsqd/http.go
telepenin: Done

nsqd/http.go
```diff
@@ -107,8 +108,7 @@ func freeMemory(w http.ResponseWriter, req *http.Request, ps httprouter.Params)
 
 func (s *httpServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if !s.tlsEnabled && s.tlsRequired {
-		resp := fmt.Sprintf(`{"message": "TLS_REQUIRED", "https_port": %d}`,
-			s.nsqd.RealHTTPSAddr().Port)
+		resp := fmt.Sprintf(`{"message": "TLS_REQUIRED", "https_addr": "%s"}`, s.nsqd.RealHTTPSAddr())
```
ploxiln, line 111: This is a backwards-incompatible change, we want to keep the old `https_port` part of the response if nsqd is not using the new unix-socket address feature.

It seems that this may not really be relevant for unix-sockets listeners anyway - you probably wouldn't use tls on the unix socket. Only a local process could connect to a unix socket, tls is less relevant so you'd probably use plain http, or it would be an https proxy like nginx terminating tls then forwarding the request to the nsqd http unix socket. So if tls-required and https-address is a unix socket ... I'd just put 0 for `https_port` for now I guess.

telepenin: Just figured out that `internal/http_api/api_request.go` `GETV1/POSTV1` has retries by that port. I just disable ability to using https via unix socket and revert part of code.

internal/http_api/api_request.go
```diff
@@ -121,7 +120,7 @@ retry:
 
 func httpsEndpoint(endpoint string, body []byte) (string, error) {
 	var forbiddenResp struct {
-		HTTPSPort int `json:"https_port"`
+		HTTPSAddr string `json:"https_addr"`
```
ploxiln, line 123: this is the other half, of the part that I think should be reverted, for backwards compatibility

telepenin: Thank you, done

ploxiln:
> I just disable ability to using https via unix socket

I'm OK with either way, I can see arguments for removing this (not really useful) and keeping it (for consistency with the other address options).

Anyway, this looks good to me, can you rebase+squash your commits down to a single commit?

telepenin:
> Anyway, this looks good to me, can you rebase+squash your commits down to a single commit?

Done.

ploxiln:
Thanks!

PR #1437, opened 2023-02-05
fix(sec): upgrade tornado to 5.1

fynntang:
### What happened？
There are 1 security vulnerabilities found in tornado 4.3
- [CVE-2018-1000518](https://www.oscs1024.com/hd/CVE-2018-1000518)

### What did I do？
Upgrade tornado from 4.3 to 5.1 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/bench/requirements.txt
+++ b/bench/requirements.txt
@@ -1,3 +1,3 @@
-tornado==4.3
+tornado==5.1
 paramiko==1.16.0
 boto==2.38.0


mreiferson:
Thanks for the PR, but the benchmark tools are used in a very specific way that makes this low risk. Also, have you tested that it actually works with a major version upgrade of Tornado?

PR #1170, opened 2019-07-26
WIP: refactor nsqd storage engine

andyxning:
Fix #1139

PR #1454, opened 2023-06-20
Fix the issue of deferred data persistence loss

qaqhy:
Fix the issue of deferred data persistence loss(Incompatible with old data storage methods)
For disk persistence we can perhaps try to do forward compatible iterative updates using flag bits

### DIFF
--- a/nsqd/channel.go
+++ b/nsqd/channel.go
@@ -240,8 +240,14 @@ finish:
 	c.inFlightMutex.Unlock()
 
 	c.deferredMutex.Lock()
+	ts := time.Now().UnixNano()
 	for _, item := range c.deferredMessages {
 		msg := item.Value.(*Message)
+		if item.Priority > ts {
+			msg.deferred = time.Duration(item.Priority - ts)
+		} else {
+			msg.deferred = 0
+		}
 		err := writeMessageToBackend(msg, c.backend)
 		if err != nil {
 			c.nsqd.logf(LOG_ERROR, "failed to write message to backend - %s", err)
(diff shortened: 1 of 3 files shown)


mreiferson:
Thanks for submitting this @qaqhy. I've got to refresh my memory on where #1170 left off to be able to provide feedback here.

qaqhy:
> Thanks for submitting this @qaqhy. I've got to refresh my memory on where #1170 left off to be able to provide feedback here.

In this commit, I merged the latest code and fixed the data race issue

qaqhy:
@mreiferson Hello, do you think this restart retention delay time plan is feasible? Or do you have any better suggestions?

mreiferson:
Hi, thanks for submitting this.

I looked back over #1170. While it had some issues in terms of the specific implementation, I think overall it has a more generic forward compatible design, rather than _specifically_ modifying the on-disk format to support just the issue of deferred persistence.

In other words, if we're going to change the on-disk format, since this will be the first time we've basically ever done that, we should establish a new format that allows us to properly version and expand capabilities while being able to better preserve forward and backward compatibility.

qaqhy:
> Hi, thanks for submitting this.
> 
> I looked back over #1170. While it had some issues in terms of the specific implementation, I think overall it has a more generic forward compatible design, rather than _specifically_ modifying the on-disk format to support just the issue of deferred persistence.
> 
> In other words, if we're going to change the on-disk format, since this will be the first time we've basically ever done that, we should establish a new format that allows us to properly version and expand capabilities while being able to better preserve forward and backward compatibility.


Another issue is that when there are many topics, loading tasks from disk to memory can be very slow. This time, establishing a new format may require solving this problem in order to apply NSQ to large-scale projects

PR #1467, opened 2023-10-25
nsqadmin: fix counter view

### LINKED ISSUE #747: nsqadmin: counter goes backwards sometimes
When looking at our NSQ counter it once each ~10s skips backwards to a lower value. Is that an expected behavior?

dudleycarr:
Fixes #747 

Two bugs are addressed on the counter view:

1. The increment counter animation is unbounded, causing the counter to exceed the true message processing count. The counter jumps back to the true message processing count after the API poll request.
2. The message processed graph below the counter doesn't properly handle the timeframe selector. The graph briefly updates when selecting the new timeframe but reverts to the timeframe at page load after the next API poll request.

Video below shows the first bug in the nsqadmin running on `4171`. The nsqadmin on `4181` has the PR fix. 

https://github.com/nsqio/nsq/assets/218825/fc2e7077-c350-4192-ad51-32fd5e81aada


jehiah:
I haven't looked at this page recently... it's always a blast to see. Below from one of my clusters.

Screenshot 2023-10-25 at 9 43 45 AM

jehiah (approved):

PR #1470, opened 2023-10-30
nsqadmin: upgrade supported ECMAScript from ES5 to ES2020 in ESLint.

dudleycarr:
The nsqadmin ESLint configuration leaves the ECMAScript version unspecified which means it defaults to ES5. ES5 was a sensible default many years ago since nsqadmin doesn't transpile its JS nor are their polyfills for missing ECMAScript functionality. 

That said, it's 2023 and modern browsers support vast majority of features from ES2020 and earlier for several years. ES2020 contains the many of the language features that you'd use on a daily basis in modern JS development.

Concrete numbers in terms of world wide adoption can be seen here:
https://caniuse.com/?search=es2020

In addition to adding `es2020` to the environment list, the `space-before-function-paren` rule has been updated with the inclusion of async anonymous arrow functions. 

Without the rule change:

```js
const foo = async() => {
  return 1;
}
```

Generally, space is provided between the function modifier and the anonymous arrow function so that it does read as if async is the function name.

With the rule change:
```js
const foo = async () => {
  return 1;
}
```

### DIFF
--- a/nsqadmin/.eslintrc
+++ b/nsqadmin/.eslintrc
@@ -1,6 +1,7 @@
 {
   "env": {
-    "browser": true
+    "browser": true,
+    "es2020": true
   },
   // Rule docs: http://eslint.org/docs/rules/
   "rules": {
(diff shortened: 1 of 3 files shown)


mreiferson:
All of this sounds reasonable to me, but you could probably tell me anything about the current state of javascript and I'd believe you.

dudleycarr:
😂 That's why I included the CanIUse link. 

I think it's a conservative change. I presume that the majority of `nsqadmin` users have modern browsers. Exceptions exist in locked down IT departments. That said, an IT department running a 2 year old browser has some serious issues.

Of course, it doesn't change anything in the existing code base. However, new code and things that get refactored could make use of those changes.

PR #1472, opened 2023-11-03
Use Go Modules and deprecate the Dep lock file

suhailpatel:
This PR migrates NSQ to Go Modules which makes it build on later versions of the Go toolchain

We can then build NSQ using `make all`


suhailpatel:
Apologies, this was created against the wrong target 😅

PR #1473, opened 2023-11-06
nsqd: use --tls-root-ca-file in nsqauth request

intellitrend-team:
This change allows the parameter `--tls-root-ca-file` to set the root CA file for auth requests as well.

### DIFF
--- a/nsqd/nsqd.go
+++ b/nsqd/nsqd.go
@@ -57,11 +57,12 @@ type NSQD struct {
 
 	lookupPeers atomic.Value
 
-	tcpServer     *tcpServer
-	tcpListener   net.Listener
-	httpListener  net.Listener
-	httpsListener net.Listener
-	tlsConfig     *tls.Config
+	tcpServer       *tcpServer
+	tcpListener     net.Listener
+	httpListener    net.Listener
+	httpsListener   net.Listener
+	tlsConfig       *tls.Config
+	clientTLSConfig *tls.Config
 
 	poolSize int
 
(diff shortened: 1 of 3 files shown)


nsqd/nsqd.go
```diff
@@ -759,6 +766,26 @@ func buildTLSConfig(opts *Options) (*tls.Config, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if !tlsCertPool.AppendCertsFromPEM(caCertFile) {
+			return nil, errors.New("failed to append certificate to pool")
+		}
+		tlsConfig.RootCAs = tlsCertPool
+	}
+
+	return tlsConfig, nil
```
dudleycarr, line 786: Question: Shouldn't `tlsConfig` be `nil` if `opts.TLSRootCAFile` is unspecified?

intellitrend-team: Our intention was to respect the existing `TLSMinVersion` flag for clients as well, which is why we're passing a `tls.Config` with `MinVersion` set to the flag. This behavior was inspired by the very similar `buildTLSConfig` function for assembling the `tls.Config` for the server:
https://github.com/nsqio/nsq/blob/eb27dd5c0102d641c758ce59297eb2a17e955ff0/nsqd/nsqd.go#L751

The documentation for `tls.Config` in Go 1.17 [the current go version specified in go.mod](https://pkg.go.dev/crypto/tls@go1.17.13#Config) specifies TLS 1.0 as the default minimum versions (which has been deprecated by [RFC 8996](https://datatracker.ietf.org/doc/html/rfc8996), however the latest Go version (1.21.3) sets TLS 1.2 as the minimum for TLS clients per default (https://pkg.go.dev/crypto/tls@go1.21.3#Config). Because of this, we would want to respect the existing config option in Go 1.17. Alternatively, the project could be upgraded to Go 1.21 to ensure that a secure TLS version is used.

Because (according to the documentation) the zero value of `tls.Config` should work in the same way its nil value does, this should not affect any other TLS options.

mreiferson:
LGTM, thanks!

PR #1474, opened 2023-11-14
Dockerfile: handle non-existing nsswitch.conf properly

martin-sucha:
The RUN step fails with exit code 1 if using [ directly without if. Added if clause so that the build can proceed if /etc/nsswitch.conf already exists.


mreiferson:
thanks, but @dudleycarr fixed this in #1478

PR #1478, opened 2023-12-15
Dockerfile: remove nsswitch.conf check

### LINKED ISSUE #1477: Docker build broken by `nsswitch.conf` check
NSQ Dockerfile depends on the latest alpine image. When building the NSQ Docker image, it produces the following error:

SCR-20231214-oiwd

This error occurs because Alpine now includes `/etc/nsswitch.conf` in the image:

SCR-20231214-ojbq

When the file exists, the condition `[ ! -e /etc/nsswith.conf ]` will return a status code of 1, failing the `RUN` and causing the build to fail.

Aside from the fact that Alpine now includes `/etc/nsswith.conf`, Golang has addressed the lack of `/etc/nsswitch.conf` since version 1.16. See this commit for reference:

https://github.com/golang/go/commit/c80022204e8fc36ec487888d471de27a5ea47e17

dudleycarr:
Fixes #1477

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,10 +17,6 @@ EXPOSE 4150 4151 4160 4161 4170 4171
 RUN mkdir -p /data
 WORKDIR      /data
 
-# set up nsswitch.conf for Go's "netgo" implementation
-# https://github.com/golang/go/issues/35305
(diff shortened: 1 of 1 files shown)


mreiferson:
Thanks!

PR #1494, opened 2024-07-29
feat: Add channel statistics data return command

qaqhy:
API consumers can query channel statistics information
image

### DIFF
--- a/nsqd/protocol_v2.go
+++ b/nsqd/protocol_v2.go
@@ -196,6 +196,8 @@ func (p *protocolV2) Exec(client *clientV2, params [][]byte) ([]byte, error) {
 		return p.CLS(client, params)
 	case bytes.Equal(params[0], []byte("AUTH")):
 		return p.AUTH(client, params)
+	case bytes.Equal(params[0], []byte("STATS")):
+		return p.STATS(client, params)
 	}
 	return nil, protocol.NewFatalClientErr(nil, "E_INVALID", fmt.Sprintf("invalid command %s", params[0]))
 }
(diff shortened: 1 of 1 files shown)


mreiferson:
I think this seems reasonable, thanks for submitting! Why didn't you use the `client.Stats(topic)` function to pull everything relevant to the client, though?

qaqhy:
The client.Stats (topic) method cannot retrieve the information of client_countt, clients, and e2e_processing_1atency, and it is unknown whether this information is suitable to be returned together
image

qaqhy:
> I think this seems reasonable, thanks for submitting! Why didn't you use the `client.Stats(topic)` function to pull everything relevant to the client, though?

The HTTP method requires initiating a request, and in practical use, it is more convenient to directly use the Consumer object

mreiferson:
I thought these were _client_ stats, these are actually _channel_ stats. What's the thinking here? Regardless of the semantics of the `STATS` command, we should use existing methods of safely locking and pulling state rather than duplicating that code directly in this command handler (e.g. `NewChannelStats()` in this case).

(FWIW the client.Stats function does not perform an HTTP request).

qaqhy:
> I thought these were _client_ stats, these are actually _channel_ stats. What's the thinking here? Regardless of the semantics of the `STATS` command, we should use existing methods of safely locking and pulling state rather than duplicating that code directly in this command handler (e.g. `NewChannelStats()` in this case).我以为这些是客户统计，这些实际上是渠道统计。你在想什么？不管STATS命令的语义如何，我们都应该使用现有的安全锁定和拉取状态的方法，而不是直接在这个命令处理程序中复制代码（例如，本例中的NewsletelStats（））。
> 
> (FWIW the client.Stats function does not perform an HTTP request).(FWIW stats函数不执行HTTP请求）。

I read the `NewChannelStats()` method, but in the `STATS` method, it retrieves the statistics of the subscribed channel as a client. I believe that in my client connection, it shouldn't be able to access information about other clients (as this doesn't seem to be public information). If I call it like `NewChannelStats(client.Channel, nil, 0)`, the `client_count` and `clients` would be incorrect. In reality, I don't want the client to know the `client_count` and `clients` data.

mreiferson:
I still think we should improve the API for how we get channel-related stats for an individual client, rather than duplicating the code and making this method aware of the concurrency and locking necessary to do it correctly.

qaqhy:
> I still think we should improve the API for how we get channel-related stats for an individual client, rather than duplicating the code and making this method aware of the concurrency and locking necessary to do it correctly.

The main data we want to obtain from this API are depth and backend_depth. The other data are additional and can be omitted. This way, we can avoid locking and unlocking operations in the Stats method.

qaqhy:
> > I still think we should improve the API for how we get channel-related stats for an individual client, rather than duplicating the code and making this method aware of the concurrency and locking necessary to do it correctly.
> 
> The main data we want to obtain from this API are depth and backend_depth. The other data are additional and can be omitted. This way, we can avoid locking and unlocking operations in the Stats method.

We want the connected client to be able to check how many tasks are left to consume in the subscribed channel, which will help us predict the remaining consumption time for our business purposes. We prefer not to query this information via HTTP; using TCP within the business code would be more convenient.

PR #1497, opened 2024-09-02
nsqd: Fix an issue where a client sending messages to multiple topics is unable to aggregate statistics for all topics #1492

### LINKED ISSUE #1492: There is an error in the statistics information
When a client sends messages to multiple topics, the statistics information can only output the results of one topic for a single client. After reviewing the code, I found that the problem lies in the break statement on line 305 of the client_v2.go file. Can you explain why this break statement is added?
`	for topic, count := range c.pubCounts {
		if len(topicName) > 0 && topic != topicName {
			continue
		}
		pubCounts = append(pubCounts, PubCount{
			Topic: topic,
			Count: count,
		})
		break   **//  What does this break statement do?**
	}
`

zhaohaihang:
fix #1492 
* When a client sends messages to multiple topics, the statistics information can only output the results of one topic for a single client.

### DIFF
--- a/nsqd/client_v2.go
+++ b/nsqd/client_v2.go
@@ -302,7 +302,6 @@ func (c *clientV2) Stats(topicName string) ClientStats {
 			Topic: topic,
 			Count: count,
 		})
-		break
 	}
 	c.metaLock.RUnlock()
 	stats := ClientV2Stats{


nsqd/client_v2.go
```diff
@@ -302,7 +302,6 @@ func (c *clientV2) Stats(topicName string) ClientStats {
 			Topic: topic,
 			Count: count,
 		})
-		break
```
mreiferson, line 305: we should keep the break when `topicName` is not empty.

mreiferson (changes requested):
Looked through the code in more detail, there's another bug (I think) that will break with this fix.

https://github.com/nsqio/nsq/blob/master/nsqd/stats.go#L154 needs to be `t` rather than `topic`.

PR #1498, opened 2024-10-03
nsqadmin: fix graphite 'target' param serialization

johnou:
Changed the `$.param()` calls to `$.param(q, true)` to ensure that arrays are serialized using traditional style, resulting in multiple target parameters instead of target[].

From graphite docs [1]
```
The target param can also be repeated to graph multiple related metrics.

&target=company.server1.loadAvg&target=company.server1.memUsage
```

with the current array param our graphite server is responding with

```
{"errors": {"target": "This parameter is required."}}
```

removing the array from the param fixes the graphs.

[1] https://graphite.readthedocs.io/en/latest/render_api.html#target

### DIFF
--- a/nsqadmin/static/js/lib/handlebars_helpers.js
+++ b/nsqadmin/static/js/lib/handlebars_helpers.js
@@ -257,7 +257,7 @@ Handlebars.registerHelper('sparkline', function(typ, node, ns1, ns2, key) {
         return 'summarize(' + t + ',"' + interval + '","avg")';
     });
 
-    return AppState.get('GRAPHITE_URL') + '/render?' + $.param(q);
+    return AppState.get('GRAPHITE_URL') + '/render?' + $.param(q, true);
 });
 
 Handlebars.registerHelper('large_graph', function(typ, node, ns1, ns2, key) {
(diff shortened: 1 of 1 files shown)


johnou:
@mreiferson ptal, fwiw I would be more comfortable with you rebuilding the minified javascript.. :) it feels wrong submitting "obfuscated" code.

mreiferson:
Thanks for this. Yes, I'll build the JS if you want to update this PR to just have the actual fix.

johnou:
@mreiferson thanks, updated!

johnou:
@mreiferson ptal I reverted the readability / stylistic changes too.

johnou:
fwiw this regression was caused by https://github.com/nsqio/nsq/pull/1062 specifically upgrading jquery.. "As of jQuery 3.0, the `$.param()` method no longer uses jQuery.ajaxSettings.traditional as its default setting and will default to false. For best compatibility across versions, call `$.param()` with an explicit value for the second argument and do not use defaults." [1]

[1] https://api.jquery.com/jQuery.param/#jQuery-param-obj-traditional

johnou:
@mreiferson let me know if you are happy with using the traditional param, we could also change the way the params are built but it's probably not worth the effort.

mreiferson:
LGTM

PR #1502, opened 2024-11-18
*: compile js

mreiferson:
compiling #1498


mreiferson:
FWIW it seems like we have quite a few JS dependencies to update, but this is built without changing any of them (mostly because I don't really feel like dealing with the side effects of upgrading all of them right now).

Issue #1499, opened 2024-10-24
nsqd: add support for configurable `max-defer-time` for DPUB in NSQD


c3llus:
# TL;DR
Currently, NSQD does not support a dedicated max-defer-time configuration for deferred publish operations. Instead, deferred publish times are implicitly governed by the `--max-req-timeout` configuration, which (I believe) was originally intended for requeueing messages? This creates a bit of confusion since requeueing and deferring are two different operations. Ideally, there should be a dedicated setting to control the maximum defer time.

## Proposed Changes
Add a new `--max-defer-time` config for NSQD, which specifically limits deferred publish times without affecting the requeue timeout. This would make the system more flexible and reduce the confusion between the two operations.

<br>

<sub>**Next Steps**: If this sounds good, I’d love to help by submitting a PR to implement it. This is one of my first contributions to open source, so I would really appreciate any feedback and help from the maintainers to make sure everything is on track!</sub>

mreiferson:
Yea this sounds good, feel free to submit a PR, thanks!

PR #1504, opened 2024-11-18
nsqd: add support for configurable max-defer-delay for DPUB

c3llus:
## Background
https://github.com/nsqio/nsq/issues/1499

## Changes Done

1. new `max-defer-delay` command line options
2. default `1h` for `max-defer-delay`
3. refer to the new `max-defer-delay` opt instead of the old `max-req-timeout` for:
a. HTTP
b. TCP

## TODO
1. update the docs

### DIFF
--- a/apps/nsqd/options.go
+++ b/apps/nsqd/options.go
@@ -159,6 +159,7 @@ func nsqdFlagSet(opts *nsqd.Options) *flag.FlagSet {
 	flagSet.Int64("max-msg-size", opts.MaxMsgSize, "maximum size of a single message in bytes")
 	flagSet.Duration("max-req-timeout", opts.MaxReqTimeout, "maximum requeuing timeout for a message")
 	flagSet.Int64("max-body-size", opts.MaxBodySize, "maximum size of a single command body")
+	flagSet.Duration("max-defer-delay", opts.MaxDeferDelay, "maximum duration when deferring a message")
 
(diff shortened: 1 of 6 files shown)


c3llus:
Hey @mreiferson, just following up on this PR; would really appreciate your review when you have time!

Thanks!

c3llus:
hi @mreiferson, just following up on this PR; would really appreciate your review when you have time!

Thanks!

mreiferson:
The implementation is fine in general, but a few things:

1. Let's not rename "timeout" to "delay"
2. For users who are currently setting `--max-req-timeout` and expecting it to effect defers, there's probably some backward compatibility we want to add here. Maybe, if `--max-defer-timeout` is not set and `--max-req-timeout` is, we set the defer timeout to the req timeout?

c3llus:
Thanks for the feedback, @mreiferson! :)

I’ve updated the naming to stick with timeout instead of delay. Also added backward compatibility, when `--max-defer-timeout` isn’t explicitly set, it falls back to `--max-req-timeout` like before.

Let me know if there’s anything else you’d like to see adjusted!

mreiferson:
This is great, thank you! Can you squash down the commits, please?

c3llus:
pushed the squashed commits. let me know if you'd like any further tweaks, @mreiferson! :)

PR #1526, opened 2026-05-23
fix: remove erroneous break in Stats() pubCounts loop

goingforstudying-ctrl:
Closing this PR as it's no longer needed.


goingforstudying-ctrl:
Hi @mreiferson @ploxiln — this is a clear bugfix for multi-topic publish stats. Any chance someone could take a quick look? Happy to rebase if needed.

goingforstudying-ctrl:
CI is green, no conflicts. One-line bugfix removing an erroneous break in the Stats() loop. Ready for review.

ploxiln:
I see, this matters when `topic` is blank, meaning _all topics_

(would be nice to have a test for this bit of behavior :)

ploxiln (approved):

goingforstudying-ctrl:
Good point — I'll add a test case for the multi-topic (blank topic) path and push an update shortly.

goingforstudying-ctrl:
@ploxiln Added  in  — covers the blank-topic (all topics) and specific-topic filtering paths. Pushed. Let me know if you'd like anything else adjusted.

goingforstudying-ctrl:
@ploxiln Added TestClientStatsPubCounts in stats_test.go — it sets up an nsqd instance, publishes to two topics over the same producer connection, and asserts that both topics show up in the producer client stats with the correct counts. Confirmed it fails with the break in place and passes without it.

ploxiln:
I have awkwardly just realized that there are two older PRs with the same fix:
 * https://github.com/nsqio/nsq/pull/1525
 * https://github.com/nsqio/nsq/pull/1497

... what might justify this one is addressing the comment in #1497 :)

goingforstudying-ctrl:
@ploxiln thanks for catching that — I wasn't aware of #1525 and #1497. You're right that the test addition is the main differentiator here. The break causes only the first topic's pub count to be recorded when querying all topics, which is a real bug for anyone running multi-topic workloads. Happy to close this in favor of one of the older PRs if a maintainer prefers, but since the test is already written and pushed, I'll leave it open in case it's useful. Let me know if you'd like me to rebase or adjust anything.

goingforstudying-ctrl:
Thanks for the review and approval, @ploxiln! Appreciate you catching the overlap with #1525 and #1497 too — the test coverage for the blank-topic path should hopefully make this one useful regardless. Let me know if there's anything else you'd like adjusted.

